PR #15, opened 2018-05-31
statistics, data frames, incoherent averaging, closures averaging

wielgusm:
adding statistics and integration with pandas, incoherent averaging of closure quantities


achael:
merged maciek

PR #41, opened 2018-08-07
Addition of eh.image.load_image()

jrfarah:
I've added a function called `load_image()`. It can be accessed just by calling `eh.image.load_image()`. It robustly takes different kinds of image inputs and figures out which function to use to read it. `load_image` can read `.fits`, `.txt`, and `eh.image.Image` types and return an `eh.image.Image` object. You can also set the argument `display=True` to have it automatically display the image. 

This function can be useful in a lot of circumstances. Instead of having to check the filetype of what you're loading, this will do it for you. If you have a long list of images, in various formats, this will save you a couple of if statements and let you load them all in with the same line. Finally, it's very easy to add filetypes to the function, so its flexibility can increase in the future. 

Example usage:

```python
import ehtim as eh

images = [
	"avery_sgra_eofn.fits",
	"avery_sgra_eofn.txt",
	eh.image.load_fits("avery_sgra_eofn.fits")
]

for im in images:
	img = eh.image.load_image(im) # load image robustly, without worrying about type
	print img.psize
```

(p.s. -- this commit (unintentionally) gives image.py 2018 lines. Huzzah!)

### DIFF
--- a/ehtim/image.py
+++ b/ehtim/image.py
@@ -1986,7 +1986,8 @@ def load_image(image, display=False, aipscc=False):
       print("Image format is not recognized. Was expecting .fits, .txt, or Image. Got {0}. Returning False.".format(type(image)))
       return False
 
-    if display: im.display()
+    if display: 
+      im.display()
 
     return im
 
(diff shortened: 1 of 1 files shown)


jrfarah:
@achael @klbouman

PR #43, opened 2018-08-07
added SMAR info to array EHT2017.txt

wielgusm:
It's useful to have SMA reference station in EHT2017 array info


wielgusm:
So I don't add SMAR to EHT2017.txt
but I add file EHT2017_calibration.txt with SMAR for polarimetric calibration of actual EHT data with ehtim.

PR #58, opened 2018-10-14
fixed bug with mjd wrapping

wielgusm:


achael:
sorry, just saw this -- did you add it to polrep as well?

wielgusm:
I think you already accepted polrep change and we already used it for
production, because problems with wrapped date oj287 data disappeared:)

PR #80, opened 2019-02-26
fixing timestamp problem with scan long coherent averaging

wielgusm:
This fixes a timestamp problem with scan long coherent averaging. The timestamp of scan-averaged data is now corresponding to a GLOBAL beginning of the scan (whether parsed from vex file or inferred from the data itself with obsdata.add_scans) and not to the beginning of recording for an INDIVIDUAL baseline. This allows to occasionally recover few percent more closure quantities in the scan-averaged data.

### DIFF
--- a/ehtim/obsdata.py
+++ b/ehtim/obsdata.py
@@ -924,13 +924,14 @@ class Obsdata(object):
 
         return  out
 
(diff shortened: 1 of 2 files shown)


achael:
Please submit this pull request to dev

PR #89, opened 2019-04-11
Fixed some typos in docs

amandoabreu:
Noticed these when reading, unsure of how sphinx is setup and currently not on my development machine, do what you will with the PR :)


achael:
Thank you for catching all of my misspellings :p

PR #98, opened 2019-04-12
Update README.rst

davidgumzchoi:
removing unnecessary space


stevenaldinger:
While you're messing with that line anyway it might be a good opportunity to just change that to a markdown link `[here](https://achael.github.io/eht-imaging)`

Desires62:
looking at back log commits to reviews and see request can be pulled se what can fix

PR #99, opened 2019-04-13
Fix minor bug in image.py

gquarles:
Fixed typo in image.py
Exception was spelled incorrectly in 2 places.


gquarles:
Also fix typo in movie.py

achael:
Thank you for catching these bugs!

PR #104, opened 2019-05-17
Do automated testing with Travis CI

cclauss:
To enable free automated testing of all code changes, you would need to:
1. go to https://travis-ci.com
2. log in with GitHub credentials
3. select the __free plan__, and
4. flip the repository switch __on__.

When that is done, Travis CI will lint your Python code with http://flake8.pycqa.org

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@ astropy
 ephem
 future
 h5py
-html
 ipython
 matplotlib
 networkx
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,7 +2,6 @@ astropy
 ephem
```
joseangel-sc: I think these requirements should be frozen, is there a reason not to do this? Aside from ipython that I think is used only for debugging purposes I think it's the best way to go

Issue #51, opened 2018-09-11
UVFITS-files of VLA-data exported by CASA cannot be successfully read


cbrinkerink:
Hello Andrew,

I am hoping to use the EHT imaging library to image some data from the VLA, as it might allow me to automate some steps in the process (I have a LOT of data). In this data, Sgr A was observed and both the compact component and Sgr A West are visible. To test this, I exported a single scan at a single frequency (34.256 GHz, 125 MHz BW) from CASA (initially using version 4.7.0) using 'exportuvfits', and I tried loading it into ehtim with 'obsdata.load_uvfits' (using ehtimaging 1.0.0). Unfortunately, I got the following error message:

> UnicodeDecodeError: 'ascii' codec can't decode byte 0x90 in position 5: ordinal not in range(128)

I have tried changing the codec to 'utf8', which resulted in a different error message:

> UnicodeDecodeError: 'utf8' codec can't decode byte 0x90 in position 5: invalid start byte

I then tried using CASA 5.3.0 to export a UVFITS file, as some things in the export function have changed between 4.7.0 and 5.3.0. With the new UVFITS file, I got a new error message when attempting to import it:

> IndexError: index 26 is out of bounds for axis 0 with size 26

This looks like it perhaps has something to do with the antenna table, as I have 26 antennas in the array. Maybe a flagging-related issue? I also tried exporting a UVFITS file from CASA with the setting 'padwithflags=True', but this does not result in any observable difference.

For convenience, I have uploaded a zip file here with the four UVFITS files I have tried so far, using the two CASA versions I have available and various combinations of export settings. All of these files work just fine with Difmap, and the visibilities I see are sensible. Do you know what might be wrong, and how to fix it?

Thank you!

[UVFITSfiles.zip](https://github.com/achael/eht-imaging/files/2371400/UVFITSfiles.zip)

wheeyeon:
Hello, 
I am also interested in imaging VLA data with EHTIM. I was wondering if this issue has been resolved since the last post. If not, it seems that changing lines 1313 and 1314 of ehtim.io.load.py from
`t1 = np.array([tarr[i]['site'] for i in t1])`
`t2 = np.array([tarr[i]['site'] for i in t2])`
to
`t1 = np.array([tarr[np.where(tnums==i)[0][0]]['site'] for i in t1])`
`t2 = np.array([tarr[np.where(tnums==i)[0][0]]['site'] for i in t2])`
[...]

PR #128, opened 2020-11-16
Patch for importing VLA uvfits files

wheeyeon:
Patch for error in importing VLA uvfits files exported from CASA [(issue#51)](https://github.com/achael/eht-imaging/issues/51#issue-359082059). This is a minor edit to how station names are tagged.


achael:
Thanks for this and sorry for the delay in approving it!

PR #133, opened 2020-12-18
add optional multiprocessing to movie scattering method

gnwong:
This pull request allows the user to call ```ehtim.ScatteringModel().Scatter_Movie(mov, processes=nproc)``` with the new (optional) kwarg ```processes```. 
```python
nproc > 0   :  ehtim will scatter the movie frames on different threads using the multiprocessing module
nproc < 0   :  ehtim will use all available cores to scatter movie frames 
nproc == 0  :  ehtim does not attempt to use multiprocessing
```

By default ```nproc = 0```.


achael:
Thanks George!

PR #134, opened 2021-01-06
(joseph-dev) Added functions for file load-in and scan selection

jrfarah:
Two functions added:

- `load_obs()`: takes in any observation file (or in the case of MAPS, multiple files) and uses the extension and input arguments to determine how to load the file. For `uvfits`, `txt`, and `oifits` files, extension is used and passed to the corresponding function. For MAPS, it checks for two additional arguments, `obsspec` and `ifile`. The function has knowledge of all the arguments for `load_uvfits()`, `load_txt()`, `load_oifits()`, and `load_maps()`, and the arguments can be passed to `load_obs()` as if you were using the legacy function calls. Example usage: `ehtim.obsdata.load_obs("./sample_data.uvfits")`, `ehtim.obsdata.load_obs("./sample_data.txt", polrep='stokes')`. Tested on `txt` and `uvfits` files, but I couldn't get my hands on any oifits or MAPS files quickly.
- `getClosestScan()`: returns the scan of an observation closest in time to the argument. Uses `obs.split_obs()` to get a list of scans for an observation, and then checks each scan's `tstart` value to figure out which is closest in time to the argument. If you want to save time, you can preload the output of `split_obs` to `getClosestScan()` as an argument and it will skip generating it. Time is input in GMST. Example usage: `obs.getClosestScan(4)` and `obs.getClosestScan(4, splitObs=obs.split_obs())`. Tested on sample data file in `data/`.

### DIFF
--- a/ehtim/obsdata.py
+++ b/ehtim/obsdata.py
@@ -4665,3 +4665,78 @@ def load_maps(arrfile, obsspec, ifile, qfile=0, ufile=0, vfile=0,
     return ehtim.io.load.load_obs_maps(arrfile, obsspec, ifile,
                                        qfile=qfile, ufile=ufile, vfile=vfile,
                                        src=src, mjd=mjd, ampcal=ampcal, phasecal=phasecal)
+
+def load_obs(
+				fname, 					
+				polrep='stokes', 		
+            	flipbl=False, 				
+            	remove_nan=False, 
+            	force_singlepol=None, 
+            	channel=all, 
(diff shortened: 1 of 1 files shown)


wumpus:
Joseph, is this ancient PR still useful?

jrfarah:
I'd like to think so! I use both daily.

achael:
apologies for not merging this literally years ago!

Issue #131, opened 2020-12-10
Dockerfile fails to build


nitinkumartech:
**Describe the bug**
Building the docker file fails with error

**To Reproduce**
1. Go to project root
2. Run docker build -t eht .
3. Scroll down to step [5/5] RUN pip install -r requirements.txt  && conda install -y -c conda-forge pynfft  && echo "backend: Agg" >> /opt/conda/lib/python2.7/site-packages/matplotlib/mpl-data/matplotlibrc

**Expected behavior**
The docker build should finish successfully

**Screenshots**

<img width="1662" alt="Screen Shot 2020-12-10 at 3 26 22 PM" src="https://user-images.githubusercontent.com/8221538/101825794-22814980-3afc-11eb-97ef-fbe02d6a652c.png">

**Desktop (please complete the following information):**
 - OS: MacOS Big Sur
 - Docker Engine Version v19.03.13

PR #145, opened 2022-01-13
Update Dockerfile

gquarles:
When a new version of Debian is released, it will change the current stable to oldstable and this causes the docker build to fail on step 2 while running `apt-get update` and gives the following error

```
Reading package lists...
E: Repository 'http://security.debian.org/debian-security buster/updates InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
E: Repository 'http://deb.debian.org/debian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
E: Repository 'http://deb.debian.org/debian buster-updates InRelease' changed its 'Suite' value from 'stable-updates' to 'oldstable-updates'
```

This PR adds the flag `--allow-releaseinfo-change` which fixes these errors and the docker image finishes building successfully

I will also note that I found this by looking at https://github.com/achael/eht-imaging/issues/131 but this issue was fixed by https://github.com/achael/eht-imaging/commit/b61f511d2b72a1fa1b5907853f3c8cb838446907 so #131 can be closed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM continuumio/miniconda2
 
 # install gcc and common build dependencies
-RUN apt-get update \
+RUN apt-get --allow-releaseinfo-change update -y \
  && apt-get install -y \
       build-essential \
       pylint


gquarles:
@achael Is ehtim meant to be used with Python 3 or Python 2, currently there exists Python 3 only features that cause syntax errors when ran in the Python 2 interpreter

For example line 416 of model.py
```python
val = sample_1model_xy(*stretch_xy(x, y, params), model_type[10:], params_stretch, psize, pol=pol)
````

Unpacking was changed in Python 3 to allow this but throws the following error on Python 2 `SyntaxError: only named arguments may follow *expression`

If ehtim is meant only for Python 3 then I will change the docker file to reflect this

PR #148, opened 2022-01-20
starwarps lightcurve can't be subscripted when None

abao1999:
There were two instances where a getMeasurementTerms function call used argument tot_flux=lightcurve[t], despite lightcurve=None, so when running starwarps, I would get the error: "NoneType not subscriptable."

I fixed this by just adding if statements.


achael:
thanks for catching this! I have changed the base branch to 'dev', but will push 'dev' into 'main' by the end of the month.

PR #152, opened 2022-03-31
Fix dependencies

charmoniumQ:
- The `requirements.txt` is now installable by `pip install -r requirements.txt`.
  - Pip couldn't figure out how to install `pynfft` (due to external dependency) and `pyfits` (due to depending on an old setuptools).
- Removed unnecessary dependencies (`pyfits`, `html`).
- Added `environment.yml`. This gives a pinned set of packages on which eht-imaging works.
- Updated `Dockerfile` to Python 3.x.


achael:
Thanks for these fixes! Could you take a look and resolve the conflicts with the dev branch version of the Dockerfile and requirements.txt? These are the latest versions -- once these are merged I will merge the fixes in dev to main, as there are several fixes since the last time main was updated. Thanks!

charmoniumQ:
Merged

PR #155, opened 2022-11-10
(parametersurvey) added converge function in imager and fixed load_caltab bug

nitikayad96:


achael:
looks good! I think grads should be an optional argument to converge() though

achael:
changed base branch to dev

PR #163, opened 2023-03-15
remove deprecated numpy bool dtype import

dkloving:
This PR removes unneeded import of `bool` from `numpy` which is now deprecated. The change allows installation and use of `eht-imaging` in environments that require `numpy >= 1.24`.

Note: see [here](https://github.com/numpy/numpy/pull/22607) for where the breaking numpy change was made.


wumpus:
What happens with old numpy versions and this patch? Is there some numpy version way earlier than 1.24 where np.bool behaves differently from the builtin bool?

dkloving:
> What happens with old numpy versions and this patch? Is there some numpy version way earlier than 1.24 where np.bool behaves differently from the builtin bool?

Sorry for the delayed response.

I looked into this and as far as I can tell `np.bool` has always been an alias for Python's built-in `bool` datatype, so the proposed change should have no effect on users of earlier versions.

wumpus:
I did some googling and it appears that the (lack of?) wisdom of the Internet agrees with you -- it's a safe change even on older software.

wumpus (approved):
LGTM

wumpus:
Oh, just noticed -- you should do your PR against the "dev" branch from your dev branch.

PR #164, opened 2023-04-27
Logic bug fix

shitwolfymakes:
A simple PR to fix what appears to be a logic bug in `ehtim/imager.py`

Essentially, the origin of the bug looks like a quick copy-paste that was never corrected. This fixes that.

Side note, this will be the first in a number of bugfix PRs I plan to submit. If you'd prefer I bundle them together I can do that, but from experience the smaller the PR is the easier it is to review and incorporate

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,7 @@ Alternatively, first install NFFT manually following the instructions on the `re
 
 Documentation
 -------------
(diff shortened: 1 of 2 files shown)


achael:
thanks for catching this, apologies for the long delay as I haven't been able to actively monitor eht-imaging development/issues for the last few months.

PR #166, opened 2023-06-09
Prevent pandas 2 from being installed

ptiede:
I noticed that pandas 2 seems to break eht-imaging averaging and probably some other stuff. We should fix the averaging, but in the meantime, this will prevent someone from installing the wrong version of pandas.


achael:
good catch, we should start a project to remove pandas entirely from the averaging....

PR #186, opened 2024-06-18
tlist() should return list

samlaihei:
Method comments suggest that tlist() should return list. If the observation has only one scan, or if all scans have the same shape, np.array(datalist, dtype=object) can produce a bug, i.e. when you run the add_all() method, because the tlist can't be indexed by the array attribute.


wumpus:
Can you make a trivial example that shows the bad thing happening?

achael:
Thanks, Greg, and sorry for the delay on this!

I just tried manually overwriting the scan_table and using the scan_gather=True option in obs.tlist(), and the current code returned a tlist of length 1, but indeed obs.tlist()[0] was not of the right datatype. 

I agree reverting to a list for the output of obs.tlist() would fix this  issue, but I think the output of obs.tlist() has to be an array for eh.self_cal() function to work. I will check this and try to find a solution. 

In any case, thank you for pointing this out and I will try to fix it asap!

samlaihei:
I created a new python environment with python==3.9, then did "pip install ehtim". Afterwards, assuming I've got the "EHT2017.txt" array file, run
```
import ehtim as eh

empty = eh.image.make_empty(64, 0.1, 12.51372778, 12.3911222)
arr = eh.array.load_txt('EHT2017.txt')

tint_sec = 10 # integration time in seconds
tadv_sec = 48*60 # advance time between scans in seconds
tstart_hr = 4.75 # UST time of the start of the observation
tstop_hr = 5.6 # UST time of the end of the observation
bw_hz = 8e9 # bandwidth in Hz

mjd = 57848

obs = empty.observe(arr, tint_sec, tadv_sec, tstart_hr, tstop_hr, bw_hz, mjd = mjd, timetype='UTC', ttype='direct')
print(obs.tlist())
obs.add_all()
```
Error at obs.add_all() and observing obs.tlist(), the recarray has lost its column attributes. So indexing by the attribute fails. I note that this particular issue doesn't appear if I run inside a clone of the eht-imaging GitHub repository.

achael:
apologies for the extreme delay in addressing this. I've proposed a solution in #266 , which will merge into the `dev-backend' branch where we are working on active developments. Thanks for raising this, and sorry again for the excessive delay!

PR #187, opened 2024-06-25
Update requirements.txt

dominic-chang:
Numpy versions greater than 1 causes the following error on import of `ehtim`

```ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject```

[Here's the numpy 2.0 release notes](https://numpy.org/doc/stable/release/2.0.0-notes.html#required-changes-for-custom-legacy-user-dtypes) for reference.

__Update:__

This seems to be because `pandas<2` does not support `numpy@2.0.`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,6 +12,7 @@ paramsurvey[ray]
 requests
 scipy
 scikit-image
+numpy < '2'
 
 
 # pynfft requires external dependency; install with Conda.


achael:
Thanks Dominic - I really think we should remove all pandas dependence since it often seems to cause issues. But that's a longer term thing, so I'll merge this for now.

wumpus:
pandas>=2 is a pain in the butt, it changed how index columns work. Given that ehtim is mostly tested by people using it, we're going to end up with a lot of bug reports ☹️

Issue #192, opened 2024-08-05
`interp2d` is now deprecated in `scipy`


jrfarah:
The function `interp2d` is used in `ehtim` for example in `regrid_image`, but it is now deprecated in `scipy` as of version 1.14 and has been removed. See the documentation here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp2d.html

It is available but throws a deprecation warning as late as version 1.13. Currently `ehtim` requires `scipy` but does not specify a version. Alternatively, we could update the usage of `interp2d` to be compatible with the new `scipy`. This would involve using `RectBivariateSpline` or `bisplrep`/`bisplev`, per the documentation. This potentially comes with significant drawbacks, including being very slow compared to `interp2d`. See for example this StackOverflow discussion: https://stackoverflow.com/questions/75427538/regulargridinterpolator-excruciatingly-slow-compared-to-interp2d This is a known issue, as the `scipy` devs comment and say:

>  Commenting here for visibility. It is a known performance regression and we are actively working on it. The new interface was needed in term of maintainability moving forward.  

Do folks have a preference for enforcing `scipy=1.13` vs. updating `interp2d` to the new `scipy` functions?

PR #195, opened 2024-11-09
Updating `ehtim` to be compatible with `scipy>1.14`

jrfarah:
## Summary 

`scipy` has recently deprecated the `scipy.interpolate.interp2d` function as of version 1.14 (see my issue here: https://github.com/achael/eht-imaging/issues/192) [Instructions](https://docs.scipy.org/doc/scipy/tutorial/interpolate/interp_transition_guide.html#interp-transition-guide) are given for moving to the new functions, chiefly, `scipy.interpolate.RectBivariateSpline`. I implemented this in `ehtim` to return functionality primarily to the `regrid_image` method in the `Image` class, which relies on `interp2d`. 

## Changes

Previously, interpolation in `regrid_image` was done via

```python
interpfunc = scipy.interpolate.interp2d(y, x, np.reshape(imvec, (self.ydim, self.xdim)),
                                                    kind=interp) 
```

The new version of `scipy` requires we use `RectBivariateSpline`, which operates a little bit differently than `interp2d`. Also, the `interp` keyword of `regrid_image` can no longer be passed directly to `interp2d`, as `RectBivariateSpline` requires the order of the interpolation to be an integer. The new code reads:

```python
interp_order = {"linear": 1, "quadratic": 2, "cubic": 3}.get(interp, 1)
grid_z = np.reshape(imvec, (self.ydim, self.xdim))
interpfunc = scipy.interpolate.RectBivariateSpline(y, x, grid_z, kx=interp_order, ky=interp_order)
```

## Tests

I tested these changes using the following script:

```python
import sys
sys.path.append('../')

import ehtim as eh

import matplotlib.pyplot as plt
import numpy as np

## load the image
im = eh.image.load_txt('../models/avery_sgra_eofn.txt')
im.display()
input()



## resize image 
im_resize = im.regrid_image(300*eh.RADPERUAS, 128)
im_resize.display()
input()
```

This code works in the current version of `ehtim` on systems with `scipy<1.14` but breaks if `scipy>=1.14`. However, with the changes, this script works in the new version of `scipy` and is also backwards compatible, so it will not break even if you do have `scipy<1.14` installed.

### DIFF
--- a/ehtim/image.py
+++ b/ehtim/image.py
@@ -1070,8 +1070,16 @@ class Image(object):
             if np.any(np.imag(imvec) != 0):
                 return interp_imvec(np.real(imvec)) + 1j * interp_imvec(np.imag(imvec))
 
-            interpfunc = scipy.interpolate.interp2d(y, x, np.reshape(imvec, (self.ydim, self.xdim)),
-                                                    kind=interp)
+
+            # interpfunc = scipy.interpolate.interp2d(y, x, np.reshape(imvec, (self.ydim, self.xdim)),
+                                                    # kind=interp) ## DEPRECATED. commented out for legacy
+
+            ## new code to be compatible with scipy 1.14+
+            interp_order = {"linear": 1, "quadratic": 2, "cubic": 3}.get(interp, 1)
+            grid_z = np.reshape(imvec, (self.ydim, self.xdim))
+            interpfunc = scipy.interpolate.RectBivariateSpline(y, x, grid_z, kx=interp_order, ky=interp_order)
+
+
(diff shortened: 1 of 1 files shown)


jrfarah:
@achael if you or someone with write access could please link this PR to #192 that would be great! Thanks :)

achael:
@jrfarah are you able to look at the interp2d calls in rex.py? I will try to fix them myself later if not.

jrfarah:
Thanks for the merge @achael ! My bad--I missed the `interp2d` calls in `rex.py`. I'll take a look and try to update those function calls as well! (I'll use the same branch for the PR).

PR #208, opened 2025-04-09
Fix FOV flip bug in regrid_image(function) in image.py

wideforest:
Change two lines in **_regrid_image_**(function) of **_images.py_**.
Correct condition range in the **_tmpimg_** variable.
```
tmpimg[:, np.abs(xtarget) > fov_x / 2.] = 0.0
tmpimg[np.abs(ytarget) > fov_y / 2., :] = 0.0
```
Before correction
```
tmpimg[np.abs(xtarget) > fov_x / 2., :] = 0.0
tmpimg[:, np.abs(ytarget) > fov_y / 2.] = 0.0
```

### DIFF
--- a/ehtim/image.py
+++ b/ehtim/image.py
@@ -1081,8 +1081,8 @@ class Image(object):
 
 
             tmpimg = interpfunc(ytarget, xtarget)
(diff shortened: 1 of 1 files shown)


achael:
thank you for fixing this! let me know if you find any other issues in support for non-square images.

PR #211, opened 2026-03-24
Vectorize Scatter() for 2x speedup

jberg5:
And save a few copy() operations as well for a small memory win.

Previously the implementation of [equation 9](https://arxiv.org/pdf/1606.06315) looped over every `(rx, ry)` and computed the new coordinates one pixel at a time. This works, but Python loops are slow, and we can instead have Numpy operate on the entire matrix at once. (This isn't an algorithmic improvement, but means we're pushing as much work as possible down into C).

Depending on hardware, problem size, numpy backend, etc, solving eq 9 can be made around 100-250x faster this way. See this gist for a runnable repro: https://gist.github.com/jberg5/c26bd60101507318d0d8fb8d03891ea4. On a GCP c2-standard-4, that script gives the following:

| Size | Loop | Vectorized | Speedup |
|------|------|------------|---------|
| 64×64 | 20.0ms | 0.15ms | **137x** |
| 128×128 | 80.0ms | 0.54ms | **148x** |
| 256×256 | 325.8ms | 2.5ms | **130x** |
| 512×512 | 1356ms | 10.9ms | **124x** |

However, this routine only accounted for 50-60% of the total Scatter() runtime, so the overall speedup isn't going to be quite that dramatic.

As a very low-tech sanity check, I printed out some timing from [example_scattering.py](https://github.com/achael/eht-imaging/blob/main/examples/example_scattering.py)

```python
# Now approximate the ensemble-average image by averaging 100 realizations of the scattering
start = time.time()
im_scatt_List = [sm.Scatter(im,so.MakeEpsilonScreen(im.xdim,im.ydim,rngseed=j),DisplayImage=False) for j in range(100)]
print(f"Scatter list took {time.time()-start:.3f}s")
```

Running `python example_scattering.py` from main reports:
```
Scatter list took 7.608s
```

And from my branch:
```
Scatter list took 3.633s
```

To test correctness, I confirmed that the output is numerically identical (verified via ensemble average flux and direct array comparison).

### DIFF
--- a/ehtim/scattering/stochastic_optics.py
+++ b/ehtim/scattering/stochastic_optics.py
@@ -497,14 +497,14 @@ class ScatteringModel(object):
                 EA_im_U = (EA_Image.uvec).reshape(Ny,Nx)
             if len(Unscattered_Image.vvec):
                 EA_im_V = (EA_Image.vvec).reshape(Ny,Nx)
+            rx = np.arange(Nx)[np.newaxis, :]  # (1, Nx)
+            ry = np.arange(Ny)[:, np.newaxis]  # (Ny, 1)
             # Note: signs are negative to match the linearized approximation. The choice
             # shouldn't matter (-phi has the same power spectrum as phi), but getting the
             # *relative* sign for x- and y-directions correct is important.
-            rx = np.arange(Nx)[np.newaxis, :]  # (1, Nx)
-            ry = np.arange(Ny)[:, np.newaxis]  # (Ny, 1)
-            scale = rF**2.0 / self.observer_screen_distance / Unscattered_Image.psize
-            rxp = np.round(rx - scale * phi_Gradient_x).astype(int) % Nx
-            ryp = np.round(ry - scale * phi_Gradient_y).astype(int) % Ny
+            pixel_deflection = rF**2.0 / self.observer_screen_distance / Unscattered_Image.psize
+            rxp = np.round(rx - pixel_deflection * phi_Gradient_x).astype(int) % Nx
+            ryp = np.round(ry - pixel_deflection * phi_Gradient_y).astype(int) % Ny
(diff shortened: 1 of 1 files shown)


ehtim/scattering/stochastic_optics.py
```diff
@@ -492,26 +492,23 @@ def Scatter(self, Unscattered_Image, Epsilon_Screen=np.array([]), obs_frequency_
...
-                        AI_U[ry,rx] = EA_im_U[ryp,rxp]
-                    if len(Unscattered_Image.vvec):
-                        AI_V[ry,rx] = EA_im_V[ryp,rxp]
+            rx = np.arange(Nx)[np.newaxis, :]  # (1, Nx)
+            ry = np.arange(Ny)[:, np.newaxis]  # (Ny, 1)
+            # Annoyingly, the signs here must be negative to match the other approximation. I'm not sure which is correct, but it really shouldn't matter anyway because -phi has the same power spectrum as phi. However, getting the *relative* sign for the x- and y-directions correct is important.
+            pixel_deflection = rF**2.0 / self.observer_screen_distance / Unscattered_Image.psize
+            rxp = np.round(rx - pixel_deflection * phi_Gradient_x).astype(int) % Nx
+            ryp = np.round(ry - pixel_deflection * phi_Gradient_y).astype(int) % Ny
+            AI = EA_im[ryp, rxp]
```
jberg5, line 506: Removed the `np.copy()` calls above because numpy fancy indexing always returns a copy, not a view.

achael:
Thanks for doing this! I agree this is faster/more elegant and I also tested and made sure it gave the same image in example_scattering.py. 

I have rebased the PR to merge with the `dev' branch -- I will merge dev into main soon with another version release. 

Thanks again!

Issue #212, opened 2026-04-03
eht-imaging 2.0: Backend reorganization, JAX migration, and new capabilities


rohandahale:
# Roadmap for eht-imaging 2.0

## 0: Infrastructure

- [x] **0.1** Set up pytest infrastructure (`conftest.py`, shared fixtures) — #214
- [x] **0.2** Convert existing test scripts to proper pytest — #223
- [x] **0.3** Set up GitHub Actions CI with codecov — #220
- [x] **0.4** Add `pyproject.toml` — #218
- [x] **0.5** Set up Ruff for linting and formatting (PEP8 compliance) — #218

## 1: Imager Backend Extraction

**Already done:**
- [x] **1D.1** History cleanup: `ImagerRunState` dataclass replaces 27 `_*_list` attrs + 27 `*_last()` methods ([bc990c5](https://github.com/achael/eht-imaging/commit/bc990c5)), bugfix #215
- [x] **1A.1** Create `imager_backend.py` — 7 pure functions moved ([bc990c5](https://github.com/achael/eht-imaging/commit/bc990c5))

**Data terms:**
- [x] **1A.2** Extract `make_chisq_dict` → pure `compute_chisq_dict()` — #226
- [x] **1A.3** Extract `make_chisqgrad_dict` → pure `compute_chisqgrad_dict()` — #226
- [ ] **1A.4** Extract `make_reg_dict` → pure `compute_reg_dict()` — #227
- [ ] **1A.5** Extract `make_reggrad_dict` → pure `compute_reggrad_dict()` — #227

**Objective function:**
- [ ] **1B.1** Extract `objfunc` → pure `objective()` with all explicit args
- [ ] **1B.2** Extract `objgrad` → pure `objective_grad()` with all explicit args

**Setup and transforms:**
- [ ] **1C.1** Extract `init_imager` into sub-functions
- [ ] **1C.2** Unit tests for all standalone transform functions
- [x] **1C.3** Extract `set_embed` → pure `compute_embed()` — #224
- [ ] **1C.4** Extract `check_params` and `check_limits` as standalone validators

**Wrappers:**
- [ ] **1E** Rewrite all `Imager` methods as thin wrappers calling backend functions

**Integration:**
- [ ] **1F.1** End-to-end imaging test (Gaussian source → simulate → reconstruct → verify, all 3 ttypes)

## 2: Chi-squared and Regularizer Cleanup

Clean up `imager_utils.py` (~90 chisq functions).

[...]

PR #213, opened 2026-04-03
Add pyproject.toml with pytest, ruff, and coverage config

rohandahale:
## Summary

- Adds `pyproject.toml` alongside existing `setup.py` (setup.py remains the source of truth for package metadata)
- Configures **pytest** (`testpaths`, verbose output with short tracebacks)
- Configures **Ruff** for linting (PEP8, import sorting, pyupgrade) and defines sensible ignores for legacy physics code (`E501` line length, `E741` ambiguous variable names like `I`, `l`)
- Configures **coverage** settings for pytest-cov
- Declares `[project.optional-dependencies] dev` with pytest, pytest-cov, ruff

## Design decisions

- `setup.py` is kept as-is, no metadata duplication. `pyproject.toml` uses `dynamic = ["version"]` so version stays in one place.
- Ruff `line-length = 120`. This is generous for legacy code, stricter than no limit.
- `E741` ignored globally because physics code legitimately uses `I` (Stokes), `l` (baseline), `O` (operator).
- `F841` (unused variables) ignored in tests only.
- Ruff is configured but **not enforced on existing code** yet. CI (next PR) will run ruff only on changed files.

## How to test

```bash
pip install ruff pytest pytest-cov
ruff check ehtim/imaging/imager_backend.py  # should pass
python -m pytest tests/test_imager_history.py -q
```

## Note

35 test errors in `test_imager_history.py::TestAfterTwoRuns` are pre-existing (duplicate `maxit` kwarg in fixture) — will be fixed in the pytest infrastructure PR.

Implements tasks 0.4 and 0.5 from #212.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,58 @@
+[build-system]
+requires = ["setuptools>=64", "wheel"]
+build-backend = "setuptools.build_meta"
+
+# Project metadata stays in setup.py for backward compatibility.
+# This file provides the modern build system declaration and tool configuration.
+
+[project]
+name = "ehtim"
+dynamic = ["version"]
+
+[project.optional-dependencies]
+dev = [
+    "pytest>=7.0",
+    "pytest-cov",
+    "ruff",
+]
+
+# --- pytest ---
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+addopts = "-v --tb=short"
+
+# --- ruff ---
+
+[tool.ruff]
+target-version = "py310"
+line-length = 120
+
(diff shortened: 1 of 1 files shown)


achael (commented):
looks great. I think migrating the rest of setup.py to the new pyproject.toml could be done pretty straightforwardly & without losing backwards compatibility.

rohandahale:
`pip install .` on the dev-backend branch does not work, probably because of a conflict between the pyproject.toml and setup.py. Let's migrate fully to pyproject.toml instead of setup.py

PR #218, opened 2026-04-06
Feature/pyproject and ruff

achael:
Update of reverted [#213 ](https://github.com/achael/eht-imaging/pull/213)

## Summary

- Adds `pyproject.toml` fully replacing `setup.py` 
- Configures **pytest** (`testpaths`, verbose output with short tracebacks)
- Configures **Ruff** for linting (PEP8, import sorting, pyupgrade) and defines sensible ignores for legacy physics code (`E501` line length, `E741` ambiguous variable names like `I`, `l`)
- Configures **coverage** settings for pytest-cov
- Declares `[project.optional-dependencies] dev` with pytest, pytest-cov, ruff

## Design decisions

- `setup.py` is fully replaced and deleted. 
- `tool.setuptools.packages.find` automatically finds and installs all subpackages
- Ruff `line-length = 120`. This is generous for legacy code, stricter than no limit.
- `E741` ignored globally because physics code legitimately uses `I` (Stokes), `l` (baseline), `O` (operator).
- `F841` (unused variables) ignored in tests only.
- Ruff is configured but **not enforced on existing code** yet. CI (next PR) will run ruff only on changed files.

## How to test

```bash
pip install ruff pytest pytest-cov
ruff check ehtim/imaging/imager_backend.py  # should pass
python -m pytest tests/test_imager_history.py -q
```

## Note

35 test errors in `test_imager_history.py::TestAfterTwoRuns` are pre-existing (duplicate `maxit` kwarg in fixture) — will be fixed in the pytest infrastructure PR.

Implements tasks 0.4 and 0.5 from #212.

PR #220, opened 2026-04-07
Add GitHub Actions CI with ruff lint, pytest, and codecov

rohandahale:
- Adds GitHub Actions CI workflow triggered on PRs to `dev-backend`
- Ruff lint on changed Python files only (avoids legacy code noise)
- Pytest on Python 3.10 + 3.12 with coverage upload to Codecov
- Timeouts and pip caching for efficiency

This PR requires #218 (pyproject.toml) to be merged first, since `pip install -e ".[dev]"` needs pyproject.toml for the dev dependencies.

- Ruff checks only changed files, not the full codebase -- existing code will not block PRs
- Tests run only `test_imager_history.py` for now -- the only proper pytest file. Will expand as legacy test scripts are converted (task 0.2)
- Codecov uploads only from the Python 3.12 run to avoid duplicate reports

Implements task 0.3 from #212.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,48 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [dev-backend]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - run: git fetch origin ${{ github.base_ref }}
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - run: pip install ruff
+      - name: Ruff check (changed files only)
+        run: |
+          FILES=$(git diff --name-only --diff-filter=ACMR origin/${{ github.base_ref }}...HEAD -- '*.py' || true)
+          if [ -n "$FILES" ]; then
+            echo "$FILES" | xargs ruff check
+          else
+            echo "No Python files changed"
+          fi
+
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.12"]
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.12"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+      - run: pip install -e ".[dev]"
```
achael, line 41: is this for the future? It doesn't look like we have any [dev] dependencies in pyproject.toml

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - run: git fetch origin ${{ github.base_ref }}
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - run: pip install ruff
```
achael, line 19: do we want to tag a specific version of ruff?

achael (approved):

.github/workflows/ci.yml
achael: never mind, i see them now :)

rohandahale: You beat me to it haha! Adding here for the sake of completeness. We have dev dependencies in pyproject.toml:

```
[project.optional-dependencies]
dev = [
    "pytest>=7.0",
    "pytest-cov",
    "ruff",
] 
```

So ` pip install -e ".[dev]"` installs the package plus test/lint tools.

.github/workflows/ci.yml
rohandahale: Good point! I will change this to 

```
- uses: astral-sh/ruff-action@v1
  with:
    version: "0.15"
```

in the branch where I will start adding tests

PR #224, opened 2026-04-16
Extract set_embed into pure compute_embed function

rohandahale:
Extracts `Imager.set_embed()` into a pure function `compute_embed()` in `imager_backend.py`. The Imager method becomes a 3 line wrapper. Also fixed some linting issues in `imager.py` otherwise the CI does not pass.

## Changes

| File | What changed |
|------|-------------|
| imager_backend.py | Add `compute_embed(imvec, xdim, ydim, psize, clipfloor)` -- returns `(embed_mask, coord_matrix)` |
| imager.py | `set_embed()` now calls `compute_embed()` and unpacks results into `self._embed_mask`, `self._coord_matrix`; fixed linting |
| tests/test_imager_backend.py | New file with 6 tests: basic gaussian, clipping, clipfloor error, exact match vs Imager method, odd dimensions, coordinate units |

- Pure function: no self, no side effects, all inputs explicit
- Returns tuple instead of writing to self
- Wrapper unpacks into self attributes for backward compatibility
- 209 tests pass (203 existing + 6 new)

First extraction in Phase 1 of #212 (task 1C.3).

### DIFF
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -17,34 +17,33 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 # TODO FIX INDEXING MESS FOR MULTIFREQUENCY POLARIZATION
-# TODO for polarization imaging give better control than just initializing at 20% 
+# TODO for polarization imaging give better control than just initializing at 20%
 #       better initialization for all terms!! in init_imager
 
-from __future__ import division
-from __future__ import print_function
-
-from builtins import str
-from builtins import range
-from builtins import object
 
 import copy
 import time
 from dataclasses import dataclass
 from typing import Any
+
 import numpy as np
 import scipy.optimize as opt
 
-import ehtim.imaging.imager_utils as imutils
-import ehtim.imaging.pol_imager_utils as polutils
-import ehtim.imaging.multifreq_imager_utils as mfutils
-from ehtim.imaging.imager_backend import (
-    embed_imarr, pack_imarr, unpack_imarr,
-    transform_imarr, transform_imarr_inverse,
-    transform_gradients, make_initarr,
-    compute_embed,
-)
-import ehtim.image
 import ehtim.const_def as ehc
+import ehtim.image
+import ehtim.imaging.imager_utils as imutils
+import ehtim.imaging.multifreq_imager_utils as mfutils
+import ehtim.imaging.pol_imager_utils as polutils
+from ehtim.imaging.imager_backend import (
+    compute_embed,
+    embed_imarr,
+    make_initarr,
+    pack_imarr,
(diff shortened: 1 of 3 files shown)


rohandahale:
@achael This PR is ready to be reviewed. After the review is resolved, you can do squash merge. I won't extract more functions in other PR yet because I have done linting edits to `imager.py` that we would need for next PRs.

ehtim/imager.py
```diff
@@ -1058,31 +1069,12 @@ def init_imager(self):
...
-
-        coord = coord.reshape(self.prior_next.ydim * self.prior_next.xdim, 2)
-        coord = coord * self.prior_next.psize
-
-        self._coord_matrix = coord[self._embed_mask]
-
-        return
+        """Set embedding matrix."""
+        self._embed_mask, self._coord_matrix = compute_embed(
+            self.prior_next.imvec, self.prior_next.xdim,
```
achael, line 1074: eventually we may want to wrap all of these (and other things currently handled as Imager attributes) in some kind of dictionary, right?

tests/test_imager_backend.py
```diff
@@ -0,0 +1,88 @@
...
+        embed_mask, coord_matrix = compute_embed(
+            imgr.prior_next.imvec, imgr.prior_next.xdim,
+            imgr.prior_next.ydim, imgr.prior_next.psize,
+            imgr.clipfloor_next,
+        )
+
+        np.testing.assert_array_equal(embed_mask, imgr._embed_mask)
+        np.testing.assert_array_equal(coord_matrix, imgr._coord_matrix)
+
+    def test_odd_dimensions(self):
```
achael, line 70: add a test to look at rectangular images?

achael (commented):
looks great!

main comment is that it would be good for here and for lots of other areas in the code to start putting in tests that explicitly work with rectangular images. These are mostly supported but I'm suspicious that there are still bugs lurking around since we always use square images.

achael (approved):

ehtim/imager.py
rohandahale: Yes, that's exactly the plan for after phase 1. Once all the function signatures are visible, we will group them into dataclasses (like ImageState, ImagerConfig, DataProducts). I held off on doing it now because JAX's jit distinguishes static args from array args . Premature bundling might group them in ways that don't match JAX tracing boundaries. After the phase 1 extraction, we will have a clearer picture of which args go together.

tests/test_imager_backend.py
rohandahale: I have addressed this  in #225 with rectangular image tests for compute_embed. Broader rectangular image coverage across all test files is tracked as task 2.8 in #212.

PR #225, opened 2026-04-20
Add rectangular image tests + fix bugs found by them

rohandahale:
Addresses review feedback from #224 by adding rectangular image tests (xdim != ydim) across all test files. Rectangular tests exposed multiple real bugs in the FFT code that only manifest on rectangular images. All bugs are fixed here.

## Test coverage

After fixture collapse in 930de08 (per Andrew's review feedback), all chisq / gradient / regularizer tests run on a 32x48 synthetic Gaussian via `make_rect_image(32, 48)`, so rectangular image code paths are always exercised. Rect subsumes the square case, so separate square/rect classes were redundant.

| File | Coverage | Fixture |
|---|---|---|
| test_imager_backend.py | 5 parametrized shapes for `compute_embed` (square, tall, wide, odd square, odd rect) | `make_rect_image` factory |
| test_regularizers.py | 24 value + gradient checks over 12 regularizers | `reg_setup` |
| test_gradients.py | 22 chisq + regularizer gradient finite-diff checks | `grad_setup` |
| test_chisquared.py | 54 cross-ttype comparisons (6 dtypes x 3 ttype pairs x 3 tests) | `chisq_setup` |

`make_rect_image` is a factory fixture in conftest.py consolidated from 4 previously copy-pasted helpers. Test count: 213 total (+4 vs dev-backend — 5 new `compute_embed` shape tests replace 1 old `test_odd_dimensions`).

## Bug fixes exposed by rectangular tests

**1. `sgauss` / `sgauss_grad` reshape order (imager_utils.py)**
- `imvec.reshape(xdim, ydim)` -> `reshape(ydim, xdim)` (lines 2523, 2565)
- Caused rgauss regularizer to crash on rectangular images with shape mismatch
- Invisible for square images because xdim == ydim

**2. `sgauss_grad` incorrect analytic formula (imager_utils.py)**
- Original had extra `-2*(xx-x0)*dx0*im` term in dxx/dyy/dxy gradients
- Correct formula: `d(sigxx)/d(im_k) = ((x_k-x0)^2 - sigxx)/S` (cross-term through centroid sums to zero by definition)
- Verified numerically: new formula matches finite-difference to 1e-8 across 150 configurations (5 shapes x 2 centers x 5 PAs x 3 major/minor ratios). Old had per-pixel errors ~0.02.

**3. FFT padding axis swap (obs_helpers.py, obs_simulate.py)**
- `np.pad(imarr, ((padvalx1, padvalx2), (padvaly1, padvaly2)), ...)` pads ROWS (axis 0 = ydim) with padvalx but padvalx was computed from xdim
- Swapped to `((padvaly1, padvaly2), (padvalx1, padvalx2))` so each dimension is padded correctly
- Invisible on square because padvalx == padvaly; on rect it produced non-square padded arrays and triggered the explicit "FFT padding did not return a square image!" exception
- Verified with point-source test: DFT and FFT agree to machine precision (1e-16) for all shapes

**4. FFT gradient slicing axis swap (imager_utils.py, 9 functions)**
- `grad_arr[padvalx1:-padvalx2, padvaly1:-padvaly2]` sliced axis 0 (rows = ydim) using padvalx
- Swapped to `[padvaly1:-padvaly2, padvalx1:-padvalx2]`
- TODO comment `# or is x<-->y??` in original code is now resolved and removed

**5. Typo fixes (obs_helpers.py)**
- `Excption` -> `Exception` in orbit_skyfield error path
- `satellite` -> `satname` (undefined variable) in sat_skyfield_from_ephementry error path

- **Stale TODO comments** removed from 7 `chisqgrad_*_fft` functions
- **Named seed constant** `RNG_SEED = 4` added and used consistently across all test files
- **RNG API migration** from legacy `np.random.RandomState` to modern `np.random.default_rng`
- **Helper consolidation**: `_make_rect_image` moved from 4 duplicates to a single `make_rect_image` factory fixture in conftest.py
- **Phase factor verified** in obs_simulate.py with point-source test (machine precision agreement DFT vs FFT on rectangular images)

## Lint

Pre-existing ruff issues in obs_helpers.py and obs_simulate.py were auto-fixed (imports, format specifiers, isinstance checks). Formatting only, no logic changes.

## Related

- Implements task 2.8 in #212
- Addresses review feedback on #224

### DIFF
--- a/ehtim/imaging/imager_utils.py
+++ b/ehtim/imaging/imager_utils.py
@@ -784,7 +784,6 @@ def chisqgrad_vis_fft(vis_arr, A, vis, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevant cells and flatten
-    # TODO or is x<-->y??
     out = np.real(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
 
@@ -822,7 +821,6 @@ def chisqgrad_amp_fft(vis_arr, A, amp, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevent cells and flatten
-    # TODO or is x<-->y??
     out = np.real(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
 
@@ -860,7 +858,6 @@ def chisqgrad_bs_fft(vis_arr, A, bis, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevant cells and flatten
-    # TODO or is x<-->y??
     out = np.real(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
     return out
@@ -904,7 +901,6 @@ def chisqgrad_cphase_fft(vis_arr, A, clphase, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevant cells and flatten
-    # TODO or is x<-->y??
     out = np.imag(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
 
@@ -1025,7 +1021,6 @@ def chisqgrad_camp_fft(vis_arr, A, clamp, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevant cells and flatten
-    # TODO or is x<-->y??
     out = np.real(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
 
@@ -1071,7 +1066,6 @@ def chisqgrad_logcamp_fft(vis_arr, A, log_clamp, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevant cells and flatten
-    # TODO or is x<-->y??
     out = np.real(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
 
@@ -1195,7 +1189,6 @@ def chisqgrad_logamp_fft(vis_arr, A, amp, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevent cells and flatten
-    # TODO or is x<-->y??
     out = np.real(grad_arr[im_info.padvaly1:-im_info.padvaly2,
                            im_info.padvalx1:-im_info.padvalx2].flatten())
 
--- a/ehtim/imaging/imager_utils.py
+++ b/ehtim/imaging/imager_utils.py
@@ -2529,7 +2529,7 @@ def sgauss(imvec, xdim, ydim, psize, major, minor, PA):
     sigxy_prime = (lambda2 - lambda1)*np.cos(phi)*np.sin(phi)
 
     # we get the dimensions and image vector
-    im = imvec.reshape(xdim, ydim)
+    im = imvec.reshape(ydim, xdim)
     xlist, ylist = np.meshgrid(range(xdim), range(ydim))
     xlist = xlist - (xdim-1)/2.0
     ylist = ylist - (ydim-1)/2.0
@@ -2571,7 +2571,7 @@ def sgauss_grad(imvec, xdim, ydim, psize, major, minor, PA):
     sigxy_prime = (lambda2 - lambda1)*np.cos(phi)*np.sin(phi)
 
     # we get the dimensions and image vector
-    im = imvec.reshape(xdim, ydim)
+    im = imvec.reshape(ydim, xdim)
     xlist, ylist = np.meshgrid(range(xdim), range(ydim))
     xlist = xlist - (xdim-1)/2.0
     ylist = ylist - (ydim-1)/2.0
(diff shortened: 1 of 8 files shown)


ehtim/imaging/imager_utils.py
```diff
@@ -793,9 +784,8 @@ def chisqgrad_vis_fft(vis_arr, A, vis, sigma):
     grad_arr = grad_arr * (im_info.npad * im_info.npad)
 
     # extract relevant cells and flatten
-    # TODO or is x<-->y??
```
achael, line 796: glad the tests caught this! (though i don't think anyone uses the old ffts, these predate even the nfft implementation)

ehtim/imaging/imager_utils.py
```diff
@@ -2529,7 +2513,7 @@ def sgauss(imvec, xdim, ydim, psize, major, minor, PA):
     sigxy_prime = (lambda2 - lambda1)*np.cos(phi)*np.sin(phi)
 
     # we get the dimensions and image vector
-    im = imvec.reshape(xdim, ydim)
```
achael, line 2532: again good catch: sgauss is rarely used but it's good to fix

ehtim/observing/obs_helpers.py
```diff
@@ -371,7 +366,7 @@ def make_closure_amplitude(blue1, blue2, red1, red2, vtype,
        vtype is the  visibility type
     """
 
-    if not (ctype in ['camp', 'logcamp']):
+    if ctype not in ['camp', 'logcamp']:
```
achael, line 369: the changes to imager_utils.py above changed several not (x in y) to (x not in y), but this does the opposite. Apparently these are implemented identically (https://stackoverflow.com/questions/8738388/x-not-in-y-or-not-x-in-y), but idk if there is a style we want to stick to (this isn't important though)

ehtim/observing/obs_helpers.py
```diff
@@ -1448,7 +1443,9 @@ def fft_imvec(imvec, im_info):
     padvaly2 = im_info.padvaly2
 
     imarr = imvec.reshape(ydim, xdim)
-    imarr = np.pad(imarr, ((padvalx1, padvalx2), (padvaly1, padvaly2)),
+    # imarr.shape is (ydim, xdim). First axis (rows) is y, second (cols) is x.
```
achael, line 1446: good catch here, though fortunately again the fft code is  rarely used

achael: (we may even want to just cut out the fft code entirely once we have the new nfft implementation working).

tests/test_chisquared.py
```diff
@@ -52,6 +52,9 @@
 CONV_FUNC = "gaussian"
 FFT_INTERP_ORDER = 3
 
+# Random seed for image perturbation in fixtures
+RNG_SEED = 4
+
 
 @pytest.fixture(scope="module")
 def chisq_setup(sgra_im_small, eht_array):
```
achael: do we need separate chisq_setup and chisq_setup_rect functions? To me it feels like the full rectangular case automatically tests the square case

tests/test_chisquared.py
```diff
@@ -180,3 +214,56 @@ def _gradient_comparison(chisq_setup, dtype, pair):
...
     print(f"  {dtype} {ttype_a}-{ttype_b}: grad median={np.median(frac_diff):.6f} max={np.max(frac_diff):.6f}")
     return np.median(frac_diff), np.max(frac_diff)
+
+
+# ---------------------------------------------------------------------------
+# Rectangular image tests (xdim != ydim)
+# ---------------------------------------------------------------------------
+
+
+class TestChisqConsistencyRect:
```
achael: again not sure if we need duplicate rect and square tests, since if rect works i think it's guaranteed square will work. but up to you if you want to collapse these into a single class or not -- having explicit rect functions might still be useful to isolate any future bugs that come from xdim!=ydim

tests/test_imager_backend.py
```diff
@@ -10,6 +10,15 @@
 import ehtim as eh
 from ehtim.imaging.imager_backend import compute_embed
 
+# Parametrize over square, tall, and wide images
+IMAGE_SHAPES = [
+    (32, 32),  # square
```
achael, line 15: great!

achael (commented):
looks great -- i think this will be super helpful. I just wanted to check explicitly that you verified these tests on nfft as well as ffts? If the bugs were isolated to the fft side of the code that's not that big a deal since those are rarely used. If there were any bugs in the nfft code though we might consider patching into the dev/main branch sooner to fix those while we continue to make progress here.

ehtim/observing/obs_helpers.py
rohandahale: Just double checked this change goes not (x in y) --> (x not in y), same direction as the `imager_utils.py` changes. All of it is ruff's E713 auto-fix

ehtim/observing/obs_helpers.py
rohandahale: Agreed, happy to revisit when the new NFFT implementation lands. These fixes just keep rectangular images correct until then

tests/test_chisquared.py
rohandahale: Good point! The square fixture originally used `models/avery_sgra_eofn.txt` as a more realistic test image, but it's only 32x32 and `regrid_image` doesn't handle rectangular shapes yet. So the rect fixture used a synthetic Gaussian. But now I have collapsed to a single make_rect_image(32, 48) fixture in `930de08`, same for `test_gradients.py` and `test_regularizers.py`.

tests/test_chisquared.py
rohandahale: Same as above, deleted these duplicate Rect classes in `930de08`. Since the remaining TestChisqConsistency / TestChisqGradConsistency classes now run on the 32x48 fixture, rect is always exercised

ehtim/imaging/imager_utils.py
rohandahale: Thanks and agreed, the FFT path being rarely used probably explains why the axis swap stayed dormant. Same pattern was in all 9 chisqgrad_*_fft functions. This is a low impact fix but easy to include now that rect tests exist.

ehtim/imaging/imager_utils.py
rohandahale: Thanks! Worth flagging that `sgauss_grad` had a second bug beyond the reshape. The gradient formula itself was wrong even for square images (off by ~0.02/pixel). So even though `sgauss` is rarely used, the fix does actually improve square image behavior too.

rohandahale:
@achael 

> looks great -- i think this will be super helpful. I just wanted to check explicitly that you verified these tests on nfft as well as ffts? If the bugs were isolated to the fft side of the code that's not that big a deal since those are rarely used. If there were any bugs in the nfft code though we might consider patching into the dev/main branch sooner to fix those while we continue to make progress here.

Yes, test_chisquared.py rect classes parametrize over all 3 ttype pairs (direct-fast, direct-nfft, nfft-fast), so NFFT is exercised against both DFT and FFT on 32x48 images. All 54 NFFT rect tests pass with pynfft (locally not on CI). All 4 fixed bugs are FFT-only, not NFFT.  But sgauss / sgauss_grad bug was ttype independent. So no urgent dev/main patch is needed, the bugs were FFT-specific.

achael:
Excellent, i'm very glad the rect bugs were only in NFFT (and sgauss, which is also rarely used). I think we can proceed without a fix to the main branch then, will merge now.

achael (approved):
looks great, will be extremely useful for future tests!

PR #226, opened 2026-04-23
Extract make_chisq_dict + make_chisqgrad_dict into pure backend functions

rohandahale:
Extracts the two dictionary-making methods from `Imager` into pure functions in `ehtim/imaging/imager_backend.py` and addresses the review feedback. Implements tasks 1A.2 and 1A.3 from the Phase 1 roadmap in #212.

## Extraction (initial)

| File | Change |
|---|---|
| `ehtim/imaging/imager_backend.py` | `+compute_chisq_dict`, `+compute_chisqgrad_dict`; constants `DATATERMS` / `DATATERMS_POL` / `POLARIZATION_MODES` now live here |
| `ehtim/imager.py` | Method bodies replaced with thin wrappers; constants re-imported for backward compat |
| `tests/conftest.py` | New factory fixtures `observe`, `initialize_imager` |
| `tests/test_imager_backend.py` | New `TestComputeChisqDict` + `TestComputeChisqgradDict` classes |

## Bugs found and fixed in this PR

1. **`make_chisq_dict` was passing bundled polarimetric array to scalar chisq** — `imcur_nu` (shape `(4, N)`) instead of `imcur_nu[0]` (Stokes-I slice) when `pol_next in POLARIZATION_MODES`. `make_chisqgrad_dict` correctly sliced; the value version didn't. Users running `make_image_IP(data_term={'vis': 1000, 'pvis': 500})` would have hit a `ValueError`. Latent since commit 971bf0c (July 2024). Fixed in both method and backend; regression test `test_polarimetric_stokes_i_bundled`.
2. **`chisqgrad_m` typo at `pol_imager_utils.py:613`** — `np.real(   * np.dot(Amatrix.conj().T, mdiff))` had a missing left operand. The `m`-gradient path raised `TypeError` on every call. Derived correct factor `mimage * np.exp(-2j*chiimage)` from `pimage = I·m·exp(2i·chi)`, verified by FD in IMARR space to ~1e-11.
3. **Two more typo bugs** in untested regularizer code paths: `outgrad[0] = gradi` should be `gradout` (line 914 of `smgrad`), and `return gradoutout/norm` should be `gradout` (line 1097 of `vsm_grad`). Both would `NameError` on first call. Fixed.

## Review feedback addressed

- `_next` suffix dropped from backend args (`e81845c`)
- `DATATERMS`/`DATATERMS_POL`/`POLARIZATION_MODES` moved to `imager_backend.py`, re-exported from `imager.py` (`884ff72`)
- `_observe` and `_initialize_imager` moved to `conftest.py` as factory fixtures (`801f8d3`)
- Per-term chi² / grad tests across all 6 single-pol terms (`vis`, `amp`, `bs`, `cphase`, `camp`, `logcamp`) + 3 polarimetric terms (`pvis`, `m`, `vvis`); single uniform bound `(0.3, 10)` with `snrcut=3` works for all (`16adda8`, `0efb571`)
- The `compute_chisq_dict`/`compute_chisqgrad_dict` merge filed as task 2.11 in #212

## Ruff cleanup

`pol_imager_utils.py` had 301 pre-existing ruff errors that surfaced once CI ran ruff on the file (touched here for the `chisqgrad_m` fix). Cleaned in `3404b01`: 237 auto-fixed (whitespace, unsorted imports, unused stdlib imports, `from __future__` no-ops, `(x not in y)` form), 64 manual (E701 one-liners, raw-docstring conversions, `%`-formatting, the explicit-imports replacement, the two typo bugs above). Verified semantics-preserving; full suite passes.

## Test coverage

- **266 tests pass** (up from 213 on dev-backend; net +53 new)
- Three code paths covered: Stokes I, polarimetric (`DATATERMS_POL` + bundled Stokes-I), multifrequency
- Parametrized over all 3 ttypes (direct / fast / nfft) for chi² consistency
- Wrapper parity tests: `imgr.make_chisq*_dict(imcur)` == backend function
- Cross-dict parity test: `chisq` and `chisqgrad` share the same key set
- Per-term invariants:
  - **Noise-free + `debias=False`**: chi² < 1e-10, grad max|.| < 1e-10 (forward-model correctness)
  - **Noisy + `debias=True` + `snrcut=3`**: 0.3 < chi² < 10 (sigma normalization)
- Polarimetric tests use a `gauss_im_pol` fixture (Stokes I + Q + U + V)

## Out-of-scope items filed in #212

- **2.12** — `chisqdata_cphase_diag` / `chisqdata_logcamp_diag` raise `ValueError: inhomogeneous np.array` under NumPy ≥ 1.24; `chisqgrad_logcamp_diag` uses deprecated `np.float`. Diagonalized variants excluded from per-term parametrize.
- **2.14** — `transform_gradients` overwrites the log chain rule for the I-segment when both `log` and `mcv`/`vcv`/`polcv` are in transforms (`imager_backend.py:244` overwrites line 238 because `mcv_chain[0] = 1`). Pre-existing, affects all polarimetric I-segment gradients silently.

## Related

- Implements tasks 1A.2 and 1A.3 in #212
- Follows the Phase 1 extraction pattern from #224

### DIFF
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -35,6 +35,9 @@ import ehtim.imaging.imager_utils as imutils
 import ehtim.imaging.multifreq_imager_utils as mfutils
 import ehtim.imaging.pol_imager_utils as polutils
 from ehtim.imaging.imager_backend import (
+    DATATERMS,
+    DATATERMS_POL,
+    POLARIZATION_MODES,
     compute_chisq_dict,
     compute_chisqgrad_dict,
     compute_embed,
@@ -52,9 +55,6 @@ NHIST = 50   # number of steps to store for hessian approx
 MAXLS = 40   # maximum number of line search steps in BFGS-B
 STOP = 1e-6  # convergence criterion
 EPS = 1e-8
-
-DATATERMS = ['vis', 'bs', 'amp', 'cphase', 'cphase_diag', 'camp', 'logcamp', 'logcamp_diag']
-DATATERMS_POL = ['pvis', 'm', 'vvis']
 
 REGULARIZERS = ['gs', 'tv', 'tvlog','tv2', 'tv2log', 'l1', 'l1w', 'lA', 'patch',
                 'flux', 'cm', 'simple', 'compact', 'compact2', 'rgauss']
@@ -87,7 +87,6 @@ REGPARAMS_DEFAULT = {'major':50*ehc.RADPERUAS,
                      'alpha_A':1.0,
                      'epsilon_tv':0.0}
 
-POLARIZATION_MODES = ['P','QU','IP','IQU','V','IV','IQUV','IPV'] # TODO: treatment of V may be inconsistent
 MEANPOL_INIT = 0.2 # mean initial polarization if not in initial image
 SIGMAPOL_INIT = 1.e-2 # perturbations to initial polarization if not in initial image
 
@@ -1087,7 +1086,6 @@ class Imager:
             imcur, sorted(self.dat_term_next.keys()), self._data_tuples,
             self.obslist_next, self._logfreqratio_list, self.mf_next,
             self.pol_next, self._ttype, self._embed_mask,
-            DATATERMS, DATATERMS_POL, POLARIZATION_MODES,
         )
 
     def make_chisqgrad_dict(self, imcur):
@@ -1099,7 +1097,6 @@ class Imager:
             self.obslist_next, self._logfreqratio_list, self.mf_next,
             self.pol_next, self._ttype, self._embed_mask,
             self._which_solve, self._nimage,
-            DATATERMS, DATATERMS_POL, POLARIZATION_MODES,
         )
 
     def make_reg_dict(self, imcur):
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -35,6 +35,7 @@ import ehtim.imaging.imager_utils as imutils
 import ehtim.imaging.multifreq_imager_utils as mfutils
 import ehtim.imaging.pol_imager_utils as polutils
 from ehtim.imaging.imager_backend import (
+    compute_chisq_dict,
     compute_embed,
     embed_imarr,
     make_initarr,
@@ -1081,46 +1082,12 @@ class Imager:
         """Make a dictionary of current chi^2 term values
            input is image array transformed to bounded values
         """
-
-        chi2_dict = {}
-        for dname in sorted(self.dat_term_next.keys()):
-            # Loop over all observations in the list
-            for i, obs in enumerate(self.obslist_next):
-                if len(self.obslist_next)==1:
-                    dname_key = dname
-                else:
-                    dname_key = dname + (f'_{i}')
-
-                # get data products
-                (data, sigma, A) = self._data_tuples[dname_key]
-
-                # get current multifrequency image
-                if self.mf_next:
-                    logfreqratio = self._logfreqratio_list[i]
-                    imcur_nu = mfutils.image_at_freq(imcur, logfreqratio)
-                else:
-                    imcur_nu = imcur
-
-                # Polarization chi^2 terms
-                if dname in DATATERMS_POL:
-                    chi2 = polutils.polchisq(imcur_nu, A, data, sigma, dname,
-                                             ttype=self._ttype, mask=self._embed_mask)
-
-                # Single Polarization chi^2 terms
-                elif dname in DATATERMS:
-                    if self.pol_next in POLARIZATION_MODES:
-                        imcur_nu_I = imcur_nu[0]
-                    else:
-                        imcur_nu_I = imcur_nu
-                    chi2 = imutils.chisq(imcur_nu, A, data, sigma, dname,
-                                         ttype=self._ttype, mask=self._embed_mask)
-
-                else:
-                    raise Exception(f"data term {dname} not recognized!")
-
-                chi2_dict[dname_key] = chi2
-
-        return chi2_dict
+        return compute_chisq_dict(
+            imcur, sorted(self.dat_term_next.keys()), self._data_tuples,
+            self.obslist_next, self._logfreqratio_list, self.mf_next,
+            self.pol_next, self._ttype, self._embed_mask,
+            DATATERMS, DATATERMS_POL, POLARIZATION_MODES,
+        )
 
     def make_chisqgrad_dict(self, imcur):
         """Make a dictionary of current chi^2 term gradient values
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -36,6 +36,7 @@ import ehtim.imaging.multifreq_imager_utils as mfutils
 import ehtim.imaging.pol_imager_utils as polutils
 from ehtim.imaging.imager_backend import (
     compute_chisq_dict,
+    compute_chisqgrad_dict,
     compute_embed,
     embed_imarr,
     make_initarr,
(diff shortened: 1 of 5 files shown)


rohandahale:
@achael This PR is ready for review.

ehtim/imaging/imager_backend.py
```diff
@@ -394,3 +395,185 @@ def compute_embed(imvec, xdim, ydim, psize, clipfloor):
...
+    Parameters
+    ----------
+    imcur : np.ndarray
+        Current image array transformed to bounded values.
+    dat_term_keys : list of str
+        Data term names to evaluate, already sorted.
+    data_tuples : dict
+        Pre-computed data products keyed by dname or dname_i,
+        each value is a (data, sigma, A) tuple.
+    obslist_next : list
```
achael: super minor but maybe remove the "_next" subscripts from the variable names here in the backend function

ehtim/imaging/imager_backend.py
```diff
@@ -394,3 +395,185 @@ def compute_embed(imvec, xdim, ydim, psize, clipfloor):
     coord_matrix = coord[embed_mask]
 
     return embed_mask, coord_matrix
+
+
+def compute_chisq_dict(imcur, dat_term_keys, data_tuples, obslist_next,
+                       logfreqratio_list, mf_next, pol_next, ttype, embed_mask,
+                       dataterms, dataterms_pol, polarization_modes):
```
achael: do dataterms, dataterms_pol, and polarization_modes need to be arguments for future jax purposes or can they be kept as a global variable defined in the backend file? Since these aren't likely to change much I think keeping them as global definitions (defined either in the backed file or in const_def) would make more sense aesthetically, but I'm just not sure how/if jax handles global variables.

ehtim/imaging/imager_backend.py
```diff
@@ -394,3 +395,185 @@ def compute_embed(imvec, xdim, ydim, psize, clipfloor):
...
+            else:
+                imcur_nu = imcur
+
+            # Polarimetric chi^2 gradients
+            if dname in dataterms_pol:
+                if mf_next:
+                    pol_solve = which_solve[0:4]
+                else:
+                    pol_solve = which_solve
+                chi2grad = polutils.polchisqgrad(imcur_nu, A, data, sigma, dname,
```
achael, line 542: let's not do this now, but maybe we can consider if we need to have separate wrapper functions at all for chisq, polchisq and chisqgrad, polchisqgrad. That is maybe there should just be one wrapper for chisq and one for chisqgrad that merges the two existing ones. That would streamline these make_chisq_dict and make_chisqgrad_dict functions considerably at the expense of making those funcitons 2x as long

ehtim/imager.py
```diff
@@ -1081,109 +1083,24 @@ def make_chisq_dict(self, imcur):
...
-                    raise Exception(f"data term {dname} not recognized!")
-
-                chi2_dict[dname_key] = chi2
-
-        return chi2_dict
+        return compute_chisq_dict(
+            imcur, sorted(self.dat_term_next.keys()), self._data_tuples,
+            self.obslist_next, self._logfreqratio_list, self.mf_next,
+            self.pol_next, self._ttype, self._embed_mask,
+            DATATERMS, DATATERMS_POL, POLARIZATION_MODES,
```
achael: see comment re if global variables or constants are allowed in jax

tests/test_imager_backend.py
```diff
@@ -19,6 +26,29 @@
...
+# Multifrequency test frequencies
+REFFREQ_HZ = 230e9
+MF_ALT_FREQ_HZ = 345e9
+
+# Synthetic polarization fractions for polarimetric tests
+POL_FRAC_Q = 0.1
+POL_FRAC_U = 0.05
+
+
+def _observe(im, eht_array, ttype="direct", tstart=TSTART_HR, tstop=TSTOP_HR):
```
achael: should this be in conf_test.py and we just pull it in here? Not sure on best practice, but in that case we wouldn't have to redefine some of the constants above, right?

tests/test_imager_backend.py
```diff
@@ -95,3 +122,283 @@ def test_coord_matrix_units(self, gauss_im):
...
         max_coord = (im.xdim // 2) * im.psize
         assert np.all(np.abs(coord_matrix) <= max_coord + im.psize)
+
+
+# ---------------------------------------------------------------------------
+# Helpers for chisq dict tests
+# ---------------------------------------------------------------------------
+
+
+def _initialize_imager(obs, im, data_term, pol="I", ttype="direct",
```
achael: again maybe it makes sense to put this in conf_test if it is going to be shared in future imager tests? not sure

tests/test_imager_backend.py
```diff
@@ -95,3 +122,283 @@ def test_coord_matrix_units(self, gauss_im):
...
+        obs2 = _observe(gauss_im, eht_array, tstart=tmid, tstop=TSTOP_HR)
+        imgr, imcur = _initialize_imager(
+            [obs1, obs2], gauss_im, {"vis": 100},
+        )
+        result = _call_backend_chisq_dict(imgr, imcur)
+        assert set(result.keys()) == {"vis_0", "vis_1"}
+        for v in result.values():
+            assert np.isfinite(v)
+
+    def test_multifrequency(self, gauss_im, eht_array):
```
achael: nice!

tests/test_imager_backend.py
```diff
@@ -95,3 +122,283 @@ def test_coord_matrix_units(self, gauss_im):
...
+
+class TestComputeChisqDict:
+    """Tests for compute_chisq_dict (extracted from Imager.make_chisq_dict)."""
+
+    def test_stokes_i_single_term(self, gauss_im, eht_array):
+        """Stokes I, single data term — returns one finite chi^2."""
+        obs = _observe(gauss_im, eht_array)
+        imgr, imcur = _initialize_imager(obs, gauss_im, {"vis": 100})
+        result = _call_backend_chisq_dict(imgr, imcur)
+        assert set(result.keys()) == {"vis"}
```
achael: since we are testing the chisq on the ground truth image, does it make sense to not just test if its finite but that it is close to 1 with some tolerance?

achael (commented):
mostly just some general philosophical questions to consider on organization. but overall it looks very good!

ehtim/imaging/imager_backend.py
rohandahale: Done!

ehtim/imaging/imager_backend.py
rohandahale: I moved DATATERMS, DATATERMS_POL, POLARIZATION_MODES to imager_backend.py as module-level constants and imager.py re-imports them so external code doing from ehtim.imager import DATATERMS still works. JAX-wise this is fine I think. Immutable python lists captured as closures at trace time are equivalent to local constants, only mutable global state would invalidate jit caches, which these aren't.

ehtim/imaging/imager_backend.py
rohandahale: Good idea, I have marked this as task 2.11 in #212 (merge `imutils.chisq` + `polutils.polchisq` into a single dispatcher, same for `chisqgrad` + `polchisqgrad`). We will revisit when we touch those dispatchers in tasks 2.4 / 2.5.

ehtim/imager.py
rohandahale: Done!

tests/test_imager_backend.py
rohandahale: Good catch! I followed it down a rabbit hole that turned out to be informative. I replaced the finite check with parametrized tests covering all single-pol data terms (vis, amp, bs, cphase, camp, logcamp) and polarimetric terms (pvis, m, vvis):

- `test_chisq_zero_on_truth_no_debias[_pol]`: noise-free + debias=False → chi² < 1e-10 (forward-model correctness).
- `test_chisq_near_unity_on_noisy_truth[_pol]`: noisy + debias=True + snrcut=3 → 0.3 < chi² < 10 (sigma normalization).

Same for the gradient class.

Findings worth flagging:

- chi² ≈ 1 calibration requires snrcut > 0. Without it, low-SNR data points break the linearization in bisig (bispectrum) and the (1 − cos) saturation in chisq_cphase. With snrcut=0 the per-term chi² values come out at ~0.5 (bs), ~1.5 (cphase), and 1e2–1e12 (camp/logcamp). With snrcut=3 they all collapse to ~1 with reasonable spread, a single uniform bound (0.3, 10) works across all data terms, no per-term special-casing needed.
- For noise-free case the chi² is huge for camp/logcamp because debias=True by default along with add_th_noise=False: chisqdata_* debiases the measured values, but chisq_* computes raw predicted values from imcur. On noise-free truth, debiased measured ≠ raw predicted, by amounts that blow up at low-SNR quadrangles. Setting debias=False makes everything ~1e-29.
- `cphase_diag` and logcamp_diag are broken upstream. chisqdata_cphase_diag / chisqdata_logcamp_diag raise ValueError: inhomogeneous np.array under NumPy ≥ 1.24, and chisqgrad_logcamp_diag uses deprecated np.float. I excluded them from the parametrize dtests and filed as task 2.12 in #212.
- I discovered a separate `transform_gradients` bug (`imager_backend.py:238-246`): when both log and mcv are in transforms (the standard polarimetric case), line 238 sets the log chain rule for the I-segment, then line 244 overwrites it with mcv_chain * gradarr where mcv_chain[0] = 1. Affects all polarimetric I-segment gradients, not just m. This was pre-existing, not introduced here. Filed as task 2.14 in #212.

tests/test_imager_backend.py
rohandahale: Done! Both `_observe` and `_initialize_imager` are now factory fixtures in conftest.py. Tests inject them via (`self`, `gauss_im`, `observe`, `initialize_imager`) and the TINT_SEC / TADV_SEC / TSTART_HR / TSTOP_HR / BW_HZ constants live only in conftest now (the duplicates in test_imager_backend.py are gone).

tests/test_imager_backend.py
rohandahale: Same as above, moved to `conftest.py` as the `initialize_imager` factory fixture. Future imager tests can inject it via (`initialize_imager`) without reimplementing the setup.

rohandahale:
@achael Thanks for the review! I addressed all comments. The philosophical questions turned out to be quite useful and pulled three real bugs into the open along the way:

- `_next suffix` dropped from backend args (`e81845c`)
- Constants moved to imager_backend.py, re-exported from imager.py for backcompat (`884ff72`)
- `_observe` and `_initialize_imager `moved to `conftest.py` as factory fixtures (`801f8d3`)
- Per-term chi² / grad tests across all 6 single-pol terms + 3 polarimetric terms, with snrcut=3 to absorb low-SNR linearization breakdown so a single uniform bound (0.3, 10) works (`16adda8`, `0efb571`)
- The chisq + chisqgrad merge is filed as task 2.11 in #212

Bugs surfaced and handled along the way:

`chisqgrad_m` typo at `pol_imager_utils.py:613` (missing operand): fixed in `bb123b9`, factor derived from `pimage = I·m·exp(2i·chi)`, verified by FD to ~1e-11.
`chisqdata_cphase_diag` / `chisqdata_logcamp_diag` broken under NumPy ≥ 1.24 — out of scope, filed as task 2.12 in #212.
`transform_gradients` overwrites the log chain rule for the I-segment when both log and mcv/vcv/polcv are in transforms (line 244 overwrites line 238 because mcv_chain[0] = 1). This was pre-existing, affects all polarimetric I-segment gradients. Filed as task 2.14 in #212.

Ruff cleanup of `pol_imager_utils.py (3404b01)`: Touching the file for the chisqgrad_m fix made CI run ruff on it, exposing 301 pre-existing issues. Most were cosmetic (whitespace, unsorted imports, unused stdlib imports, from __future__ no-ops); 237 auto-fixed, 64 fixed manually. While doing it I also caught two more typos: `outgrad[0] = gradi `(should be `gradout`) at line 914 and return `gradoutout/norm` at line 1097, both in untested regularizer paths that would have raised `NameError` on first call. Fixed both.


Ready for re-review!

ehtim/imaging/imager_backend.py
achael: great!

tests/test_imager_backend.py
achael: fantastic, this is all excellent. 
1. Good test of snrcut! glad it is doing its job, though in practice people should probably flag data before putting in the imager
2. I've thought for a while that debias should be False by default, but the default has been True forever just because it originated that way and I didn't want to break old scripts. Maybe this reorg is a good time to reconsider this as well. I think I would argue for setting it False by default in the next PR 
3.  very good catch. I think the _diag implementations have never worked properly, or at least they have been broken for a while. we should return to this and add these as tested chisqs, gradients in test_chisquared.py as well
4. again an issue with simultaneous IP imaging, probably introduced when I hastily implemented multifrequency polarization two summers ago. This will be very nice to fix -- and simultaneous IP imaging could be a very tractable goal to show off by the collab meeting (as well as bigger datasets).

achael (approved):
Excellent, glad we are uncovering bugs that have escaped detection so far!

PR #227, opened 2026-04-28
Extract make_reg_dict + make_reggrad_dict; flip debias default

rohandahale:
Implements roadmap tasks 1A.4 + 1A.5 + 2.15, plus the regparams + docs follow-ups requested in review:

- **`f7e98c9` — Extract `make_reg_dict` into `compute_reg_dict` (task 1A.4).**
  Move method body to a pure backend function; relocate 11 `REGULARIZERS_*` constants to `imager_backend.py` (re-exported from `imager.py` for back-compat).

- **`386fe5b` — Extract `make_reggrad_dict` into `compute_reggrad_dict` (task 1A.5).**
  Same pattern; 144 LOC → 12-line wrapper. Drop now-unused `mfutils` import and 7 `REGULARIZERS_*` sub-list imports from `imager.py` — verified by grep that no external consumer was importing them.

- **`2fc67bb` — Flip `debias=True` → `False` default (task 2.15).**
  17 sites: 1 in `Imager.__init__`, 15 in `chisqdata_*` family in `imager_utils.py`, 1 in `initialize_imager` test fixture. Static audit found no test relied on the True default; full suite passes with zero modifications. New `test_debias_next_default_is_false` locks in the new default.

- **`f6568e6` — Pre-existing ruff I001 fix in `tests/test_imager_history.py`** (surfaced by CI lint after I touched the file).

- **`62c17c4` — Merge `dev-backend`** to pick up #228.

- **`1628570` — Pack regularizer args into `regparams` dict** (Andrew review).
  Bundled `flux`, `pflux`, `vflux`, `xdim`, `ydim`, `psize`, `beam_size`, `mf_flux` into `regparams` (alongside `major`/`minor`/`PA`/`alpha_A`/`epsilon_tv`); wrapper builds via new `_full_regparams()` helper. Dispatchers unchanged — `**regparams` spread handles binding. `compute_reg_dict` arg count dropped 18 → 10; `imager_backend.py` net –100 LOC.

- **`60a51b8` — Expand `TestTransformGradientsChainRule` docstrings** (#228 review follow-up).
  Added class-level docstring describing the J^T·grad strategy, the FD-as-ground-truth rationale, and the three coverage axes (real-usage / general / shape-dispatch); per-test docstrings explain what each case verifies.

- **`5c68567` — Document mcv/vcv as legacy in `transform_gradients`** (#228 review follow-up).
[...]

PR #228, opened 2026-04-28
Fix transform_gradients chain rule for polarimetric imaging

rohandahale:
Replaces the element-wise `*_chain` chain rule API for the polcv / mcv / vcv image-space transforms with proper Jacobian-vector products (`*_grad`). The old element-wise scheme had three independent issues:

1. **I-segment clobbering** (the original task 2.14 ask): `outarr[0:4] = *_chain * gradarr[0:4]` overwrote `outarr[0] = exp(imarr[0]) * gradarr[0]` (set just above by the log block) because all `*_chain[0] = 1` from `np.ones` defaults. Stokes I gradient was off by a factor `exp(imarr[0])` whenever both `log` and any of `{mcv, vcv, polcv}` were active — i.e., all polarimetric Imager modes that solve for I.

2. **`mcv_chain` cross-terms dropped**: phys[1]=rho and phys[3]=psi both depend on mprime via mfrac. The element-wise scheme captured only the diagonal `drho/dmprime`, missing `dpsi/dmprime`. Vanishes at vfrac=0 (the real ehtim regime for pol="IP"), so subtle in production but visible off the V=0 axis.

3. **`vcv_chain` cross-terms dropped, completely broken at mfrac=0** (real pol="IV" regime): the only nonzero gradient contribution comes from the off-diagonal `drho/dvprime · grad_phys[1]`, which the element-wise scheme cannot route into `outarr[3]`. pol="IV" gradient descent was silently broken.

## Changes

- **`ehtim/imaging/pol_imager_utils.py`**: removed `polcv_chain` / `mcv_chain` / `vcv_chain`; added `polcv_grad(imarr, gradarr)` / `mcv_grad(imarr, gradarr)` / `vcv_grad(imarr, gradarr)`. Each returns shape `(3, ...)` for solver slots (1, 2, 3) and writes into `outarr[1:4]`. Math derived analytically (cross-terms verified by the FD tests below). `safe_rho = np.where(rho > 0, rho, 1.0)` handles the rho=0 limit cleanly.
- **`ehtim/imaging/imager_backend.py`**: 6-line edit to `transform_gradients` — the polarimetric branches now call `*_grad` and write to `outarr[1:4]`, leaving the log block's `outarr[0]` untouched.
- **`tests/test_imager_backend.py`**: new `TestTransformGradientsChainRule` class with 13 tests covering all dispatch branches.

## Test coverage

13 new tests, all passing:

| Test | Branch / scenario |
|---|---|
| `[IP_vfrac0]` | log+mcv at vfrac=0 (real pol="IP") |
| `[IV_mfrac0]` | log+vcv at mfrac=0 (real pol="IV") |
| `[IPV]` | log+polcv (real pol="IPV") |
| `[IP_general]` | log+mcv at vfrac≠0 (cross-term coverage) |
| `[IV_general]` | log+vcv at mfrac≠0 (cross-term coverage) |
| `[log_only]` | log standalone |
| `[mcv_only]` / `[vcv_only]` / `[polcv_only]` | each pol cv standalone (no log) |
| `test_log_chain_preserved_under_mcv` | direct I-segment regression |
| `test_single_pol_log_chain_1d` | nimage==1 branch backfill |
| `test_multifreq_stokes_i_log_chain` | nimage==3 branch backfill |
| `test_mf_pol_shape_smoke` | nimage==10 branch shape contract |

FD verification uses centered differences with eps=1e-6, tolerance rtol=1e-5. Verified by stash-and-rerun that the buggy version fails 4/9 parametrized cases plus the explicit regression test (the rest cover unaffected branches).

## Test plan

- [x] 279/279 tests pass (266 baseline + 13 new)
- [x] Ruff clean on changed files
- [x] FD math verified — pre-fix code fails the new tests
- [x] No external API breaks (the removed `*_chain` functions had only `transform_gradients` as a caller)

## Notes for review

- Simultaneous IP imaging: The fix unblocks correct gradient flow for pol="IP" and pol="IV" / "IPV".
- Branch is **independent of #227** (`feature/extract-reg-dicts`); `git merge-tree` confirms clean auto-merge whichever lands first.

Closes the 2.14 box in #212.

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -238,12 +238,13 @@ def transform_gradients(gradarr, imarr, transforms, which_solve):
         if pol_which_solve[0]==1 and ('log' in transforms):  # full polarization, including stokes I imaging
             outarr[0] = np.exp(imarr[0]) * gradarr[0]
 
+        # Each *_grad returns shape (3, ...) so outarr[0] (the log term above) survives.
         if (pol_which_solve[1]==1 and pol_which_solve[3]==1 and ('polcv' in transforms)):
-            outarr[0:4] = polutils.polcv_chain(imarr[0:4]) * gradarr[0:4]
+            outarr[1:4] = polutils.polcv_grad(imarr[0:4], gradarr[0:4])
         elif (pol_which_solve[1]==1) and ('mcv' in transforms):
-            outarr[0:4] = polutils.mcv_chain(imarr[0:4]) * gradarr[0:4]
+            outarr[1:4] = polutils.mcv_grad(imarr[0:4], gradarr[0:4])
         elif (pol_which_solve[3]==1) and ('vcv' in transforms):
-            outarr[0:4] = polutils.vcv_chain(imarr[0:4]) * gradarr[0:4]
+            outarr[1:4] = polutils.vcv_grad(imarr[0:4], gradarr[0:4])
 
     return outarr
 
--- a/tests/test_imager_backend.py
+++ b/tests/test_imager_backend.py
@@ -12,6 +12,8 @@ from ehtim.imaging.imager_backend import (
     compute_chisq_dict,
     compute_chisqgrad_dict,
     compute_embed,
+    transform_gradients,
+    transform_imarr,
 )
 
 # Parametrize over square, tall, and wide images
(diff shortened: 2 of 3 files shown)


ehtim/imaging/imager_backend.py
```diff
@@ -238,12 +238,13 @@ def transform_gradients(gradarr, imarr, transforms, which_solve):
         if pol_which_solve[0]==1 and ('log' in transforms):  # full polarization, including stokes I imaging
             outarr[0] = np.exp(imarr[0]) * gradarr[0]
 
+        # Each *_grad returns shape (3, ...) so outarr[0] (the log term above) survives.
```
achael, line 241: it might be good to add a few comments here eventually about what these transforms do (mcv and vcv are legacy, ultimately i think we want polcv or an improved version going forward for multi-stokes)

tests/test_imager_backend.py
```diff
@@ -462,3 +464,149 @@ def test_chisq_and_chisqgrad_share_keys(gauss_im, observe, initialize_imager):
...
+            im = imarr.copy()
+            im[k, i] -= eps
+            f_p = scalar_obj(transform_imarr(ip, transforms, which_solve))
+            f_m = scalar_obj(transform_imarr(im, transforms, which_solve))
+            grad[k, i] = (f_p - f_m) / (2 * eps)
+    return grad
+
+
+class TestTransformGradientsChainRule:
+    """transform_gradients chain rule agrees with FD of transform_imarr."""
```
achael, line 492: i found these tests a little non intuitive, though I get the big picture. maybe a little more detail in the docstrings could help but there are probably more important things to worry about atm.

achael (approved):

achael:
great catch by the new tests; after the rushed implementation of multifrequency polarization (that still needs to be more fully worked on), I never verified if IP or IV imaging worked. This is a great fix

ehtim/imaging/imager_backend.py
rohandahale: Added in #227 (commit `5c68567`) , short pointer comment in `transform_gradients` noting mcv/vcv are legacy single axis parametrizations and polcv is preferred for full multi-Stokes, with a reference to the per transform Jacobian docstrings in polutils

tests/test_imager_backend.py
rohandahale: Expanded the docstrings in #227 (commit `60a51b8`): I added a class level docstring describing the `J^T·grad `strategy. Each parametrized case is also annotated. Happy to iterate further if the new version still reads opaquely

rohandahale:
Thanks! The FD was the discriminator, once it was in place, the chain rule bugs surfaced immediately. The architectural takeaway is that the element-wise *_chain API couldn't express off-diagonal Jacobian entries, the JVP form (*_grad) does. Glad pol="IV" is logically sound now

PR #230, opened 2026-05-04
Use inverse variance weighting

### LINKED ISSUE #229: Thermal error is incorrectly estimated when channels have very different weights
This code here is not using inverse variance weighting when averaging

https://github.com/achael/eht-imaging/blob/a5517b6d74bb93d53cd7b70cbcee241bc1339c2e/ehtim/io/load.py#L1410-L1414

When combining multiple measurements with different errors, the final variance should be computed as `sqrt(1/sum(weights))`, not `sqrt(sum(1/weights))`. Now, if the weights are identical all rv's considered, the two answers are identical. However, if they differ, there can be pretty pathological behavior. For instance, suppose one weight is near zero. With the implementation, `sqrt(sum(1/weights))` will be dominated by the weight near zero and produce a large thermal error, while `sqrt(1/sum(weights))` will effectively ignore that point. 

This comes up in eht data that has been reduced with the HOPS pipeline when there has been sub-channel notching, e.g., NOEMA in 2021-2024. Certain channels will be almost entirely removed, meaning the signal/weight in those channels is tiny. After averaging, the SNR on NOEMA baselines is artificially reduced to near zero for the NOEMA EHT data.

ptiede:
Fixes #229

### DIFF
--- a/ehtim/io/load.py
+++ b/ehtim/io/load.py
@@ -1028,7 +1028,8 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
                     allow_singlepol=True, force_singlepol=None,
                     channel=all, IF=all, remove_nan=False,
                     ignore_pzero_date=True,
-                    trial_speedups=False):
+                    trial_speedups=False,
+                    invvar_channel_avg=True):
     """Load observation data from a uvfits file.
 
        Args:
@@ -1041,9 +1042,11 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
            channel (list): list of channels to average in the import. channel=all averages all
            IF (list): list of IFs to  average in  the import. IF=all averages all
            remove_nan (bool): whether or not to remove entries with nan data
-           
            ignore_pzero_date (bool): if True, ignore the offset parameters in DATE field 
                                      TODO: what is the correct behavior per AIPS memo 117?
+           trial_speedups (bool): if True, use faster array/telescope handling paths
+           invvar_channel_avg (bool): if True, average IFs/channels with inverse-variance
+                                      weighting. If False, use the original simple averaging.
        Returns:
            obs (Obsdata): Obsdata object loaded from file
     """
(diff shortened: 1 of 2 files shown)


achael (changes requested):
I agree averaging the channels w/ inverse variance weighting should be the default, but can you add an option in load_uvfits (e.g. invvar_channel_avg=True or simple_channel_avg=False) and keep the original simple averaging logic in case it is needed for backwards compatibility in the future?

ptiede (commented):
Added a legacy switch. Do you want to add a warning if the old method is used? It really can lead to pretty pathological behavior. 

One other thing. I was looking at the time averaging and there is a similar issue where low SNR points are given equal weight to high SNR points. I'm not sure I have the cycles to fix that as well, but it is something to keep in mind for v2.

ehtim/io/load.py
```diff
@@ -1394,38 +1397,63 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
...
+        rl = (np.nansum(np.nansum(rl_2d * rlweight, axis=2), axis=1) /
+              rl_totweight)[mask]
+        lr = (np.nansum(np.nansum(lr_2d * lrweight, axis=2), axis=1) /
+              lr_totweight)[mask]
+
+        rrsig = (1. / np.sqrt(rr_totweight))[mask]
+        llsig = (1. / np.sqrt(ll_totweight))[mask]
+        rlsig = (1. / np.sqrt(rl_totweight))[mask]
+        lrsig = (1. / np.sqrt(lr_totweight))[mask]
+    else:
```
achael, line 1432: sure, adding a warning here that this is primarily for legacy compatiblity if necessary and shouldn't be used without good reason is a good idea

achael (changes requested):

achael:
I think a warning is a good idea, though if this is default most people shouldn't run into it. and thanks for the heads up on the time averaging. I'm hoping we can completely rewrite the time/scan averaging code in v2 since its the only part of the code written in pandas and has remained pretty opaque/unedited for years because of that (sorry to Maciek).

PR #231, opened 2026-05-04
Extract objfunc/objgrad + remove chisq_transform legacy

rohandahale:
Covers tasks 1B.1 / 1B.2 plus a thorough cleanup after the code review. After this lands, only `init_imager` (1C.1) and the integration test (1F.1) remain in Phase 1.

### Original extraction (commits 1–3)
- **`9c1f965`** — extract `Imager.objfunc` → pure `compute_objective` (1B.1)
- **`1777a90`** — extract `Imager.objgrad` → pure `compute_objective_grad` (1B.2)
- **`a10d040`** — fix `chisq_transform=True` multi-obs `KeyError` *(superseded by C1 below; the whole `chisq_transform` path is now removed)*

### Review response (commits 4–8, addressing comments)
- **`3f59865` (C1)** — drop the `chisq_transform` legacy path entirely. Removes the param and branches from `compute_objective` / `compute_objective_grad`, the `Imager.chisq_transform` / `chisq_offset_gradient` attrs
- **`91456fb` (C2)** — replace `obslist` argument with explicit `n_obs: int` in all six backend functions (`compute_chisq_dict`, `compute_chisqgrad_dict`, `compute_reg_dict`, `compute_reggrad_dict`, `compute_objective`, `compute_objective_grad`). Backends never used obslist for anything but `len(obslist)`. New consistency assertion in `Imager.init_imager` catches any future drift between `len(obslist_next)`, `len(freq_list)`, and `len(_logfreqratio_list)`.
- **`bdda3a6` (C3)** — reorder backend signatures (variable + initial-reference + mode flags first; data state in the middle; reg state and image-side machinery last). Rename function parameters: `xarr` → `initvec`, `xprior` → `priorvec`. Rename Imager attributes for consistency: `_xarr` → `_init_arr`, `_xprior` → `_prior_arr`, `_xinit` → `_init_vec`. Local vars in `init_imager` likewise: `initarr` → `init_phys` and `init_solver` (making the physical-vs-solver-space pipeline visible at the call site); `priorarr` → `prior_phys`. Naming convention is documented at the top of `imager_backend.py` and in `init_imager`'s docstring.
- **`0efecbe` (C4)** — full NumPy-style docstrings (Parameters / Returns) on `compute_objective` and `compute_objective_grad`. The other four backend functions already had them.
- **`df5c311` (C5)** —Rename `priorarr` parameter → `init_arr` in `unpack_imarr`. The function fills not-solved-for slots with the *initial* image, not the regularizer prior; the old name was historically misleading.

## n_obs design rationale
We went with explicit `n_obs: int` over relying on `len(logfreqratio_list)` because:
1. **JAX-friendly** — `n_obs` is a natural `static_argname` for `jax.jit`. Computing `len(traced_array)` inside a jitted function works but is less clean.
2. **Decouples** chi^2 / regularizer compute semantics from frequency-list semantics. A future Phase 6 refactor introducing per-channel frequency lists won't break the chi^2 path.
3. **Cleaner validation locus** — `Imager.init_imager` asserts the cross-list invariant upfront

## Test coverage
- `TestComputeObjective` — 10 tests: single-pol, polarimetric, IP-bundled, multifreq, rect-image parametrize.
- `TestComputeObjectiveGrad` — 11 tests: same on gradients + 2 finite-difference verifications (Stokes I + IP) against `compute_objective`.
- `test_objective_and_objective_grad_share_consistency` — parity invariant.

## JAX-readiness
The new pure functions are designed for `jax.jit` / `jax.grad` once the underlying backends are JAXified in Phase 5:
- Pure, no `self`, no `+=` on arrays, no Python branching on array values.
- Planned `static_argnames` (now flat in the namespace): `('mf', 'pol', 'transforms', 'ttype')` plus static dict-key structure on `dat_term` / `reg_term` / `regparams`.
- `compute_objective_grad` is structured so `jax.grad(compute_objective)` is a drop-in replacement during Phase 5D's three-way check (analytic vs numeric vs autodiff).

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -858,8 +858,7 @@ def compute_objective(imvec, xarr, which_solve, transforms,
                       dat_term, reg_term,
                       data_tuples, obslist, logfreqratio_list,
                       mf, pol, ttype, embed_mask,
-                      xprior, norm_reg, regparams,
-                      chisq_transform=False):
+                      xprior, norm_reg, regparams):
     """Pure objective: data fidelity + regularization, summed with hyperparameter weights."""
 
     dat_term_keys = sorted(dat_term.keys())
@@ -884,11 +883,7 @@ def compute_objective(imvec, xarr, which_solve, transforms,
         weight = dat_term[dname]
         for i in range(n_obs):
             key = dname if n_obs == 1 else f"{dname}_{i}"
-            chi2 = chi2_dict[key]
-            if chisq_transform:
-                datterm = datterm + weight * (chi2 + 1.0 / chi2 - 1.0)
-            else:
-                datterm = datterm + weight * (chi2 - 1.0)
+            datterm = datterm + weight * (chi2_dict[key] - 1.0)
 
     regterm = 0.0
     for regname in reg_term_keys:
@@ -902,8 +897,7 @@ def compute_objective_grad(imvec, xarr, which_solve, transforms,
                            data_tuples, obslist, logfreqratio_list,
                            mf, pol, ttype, embed_mask,
                            xprior, norm_reg, regparams,
-                           nimage,
-                           chisq_transform=False, chisq_offset_gradient=0.0):
+                           nimage):
     """Pure objective gradient: data-fidelity grad + regularization grad,
     chain-ruled through the bounded-value transform, packed into solver space."""
 
@@ -920,12 +914,6 @@ def compute_objective_grad(imvec, xarr, which_solve, transforms,
         obslist, logfreqratio_list, mf, pol, ttype, embed_mask,
         which_solve, nimage,
     )
-    chi2val_dict = None
-    if chisq_transform:
-        chi2val_dict = compute_chisq_dict(
-            imcur, dat_term_keys, data_tuples,
-            obslist, logfreqratio_list, mf, pol, ttype, embed_mask,
-        )
 
     reggrad_dict = compute_reggrad_dict(
         imcur, reg_term_keys, xprior, embed_mask,
@@ -939,12 +927,7 @@ def compute_objective_grad(imvec, xarr, which_solve, transforms,
         weight = dat_term[dname]
         for i in range(n_obs):
             key = dname if n_obs == 1 else f"{dname}_{i}"
-            chi2_grad = chi2grad_dict[key]
-            if chisq_transform:
-                chi2_val = chi2val_dict[key]
-                datterm = datterm + weight * chi2_grad * (1.0 - 1.0 / (chi2_val ** 2))
-            else:
-                datterm = datterm + weight * (chi2_grad + chisq_offset_gradient)
+            datterm = datterm + weight * chi2grad_dict[key]
 
     regterm = 0.0
     for regname in reg_term_keys:
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -852,3 +852,46 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
         reggrad_dict[regname] = reggrad
 
     return reggrad_dict
+
+
+def compute_objective(imvec, xarr, which_solve, transforms,
+                      dat_term, reg_term,
+                      data_tuples, obslist, logfreqratio_list,
+                      mf, pol, ttype, embed_mask,
+                      xprior, norm_reg, regparams,
+                      chisq_transform=False):
+    """Pure objective: data fidelity + regularization, summed with hyperparameter weights."""
+
+    dat_term_keys = sorted(dat_term.keys())
+    reg_term_keys = sorted(reg_term.keys())
+
+    # Unpack solver vector into image array, then apply bounded-value transform
+    imcur = unpack_imarr(imvec, xarr, which_solve)
+    imcur = transform_imarr(imcur, transforms, which_solve)
+
+    chi2_dict = compute_chisq_dict(
+        imcur, dat_term_keys, data_tuples,
+        obslist, logfreqratio_list, mf, pol, ttype, embed_mask,
+    )
+    reg_dict = compute_reg_dict(
+        imcur, reg_term_keys, xprior, embed_mask,
+        mf, obslist, logfreqratio_list, pol, norm_reg, regparams,
+    )
+
+    n_obs = len(obslist)
+    datterm = 0.0
+    for dname in dat_term_keys:
+        weight = dat_term[dname]
+        for i in range(n_obs):
+            key = dname if n_obs == 1 else f"{dname}_{i}"
+            chi2 = chi2_dict[key]
+            if chisq_transform:
+                datterm = datterm + weight * (chi2 + 1.0 / chi2 - 1.0)
+            else:
+                datterm = datterm + weight * (chi2 - 1.0)
+
+    regterm = 0.0
+    for regname in reg_term_keys:
+        regterm = regterm + reg_term[regname] * reg_dict[regname]
+
+    return datterm + regterm
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -889,7 +889,58 @@ def compute_objective(imvec, initvec,
                       dat_term, reg_term,
                       priorvec, norm_reg, regparams,
                       transforms, embed_mask, ttype):
-    """Pure objective: data fidelity + regularization, summed with hyperparameter weights."""
+    """Compute the scalar imaging objective: data fidelity + regularization.
+
+    Pure-functional version of Imager.objfunc. Unpacks the 1D solver vector
+    `imvec` into a multi-D image array (filling not-solved-for slots from
+    `initvec`), applies the bounded-value transforms, evaluates each chi^2
+    and regularizer term via the dispatcher dicts, and returns the
+    hyperparameter-weighted sum.
+
+    Parameters
+    ----------
+    imvec : np.ndarray
+        1D solver vector (length sum(which_solve) * nimage), the optimization
+        variable for L-BFGS-B.
+    initvec : np.ndarray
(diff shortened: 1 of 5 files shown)


ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
...
     return reggrad_dict
+
+
+def compute_objective(imvec, xarr, which_solve, transforms,
+                      dat_term, reg_term,
+                      data_tuples, obslist, logfreqratio_list,
+                      mf, pol, ttype, embed_mask,
+                      xprior, norm_reg, regparams,
+                      chisq_transform=False):
+    """Pure objective: data fidelity + regularization, summed with hyperparameter weights."""
```
achael: I think a more detailed docstring here would help a bit since some of these parameters are less self-explanatory when taken out of the Imager object

ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
         reggrad_dict[regname] = reggrad
 
     return reggrad_dict
+
+
+def compute_objective(imvec, xarr, which_solve, transforms,
```
achael: For clarity, i'd propose re-ordering the argument list and renaming some of the arguments. 
  
  compute_objective(imvec, initvec, mf, pol, 
  which_solve, data_tuples, logfreqratio_list, dat_term, reg_term, 
  priorvec, norm_reg, regparams, transforms, embed_mask, ttype)
  
where we replace the name 'xarr' here with initvec (it's the unwrapped initial image) and 'xprior' with priorvec

See my comment below, I think we should re-engineer compute_chisq_dict, compute_chisqgrad_dict so they are not using obslist directly; all the data is in the data_tuples at this point.

ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
...
+                      dat_term, reg_term,
+                      data_tuples, obslist, logfreqratio_list,
+                      mf, pol, ttype, embed_mask,
+                      xprior, norm_reg, regparams,
+                      chisq_transform=False):
```
achael: I appreciate catching the bug for chisq_transform, but I actually think it makes sense to delete. this was for backwards compatibility from way way way back (pre 2019) from the first version of ehtim that had a bug in chisq gradients. I don't think it's been used in years and is pretty irrelevant now

ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
...
+    dat_term_keys = sorted(dat_term.keys())
+    reg_term_keys = sorted(reg_term.keys())
+
+    # Unpack solver vector into image array, then apply bounded-value transform
+    imcur = unpack_imarr(imvec, xarr, which_solve)
+    imcur = transform_imarr(imcur, transforms, which_solve)
+
+    chi2_dict = compute_chisq_dict(
+        imcur, dat_term_keys, data_tuples,
+        obslist, logfreqratio_list, mf, pol, ttype, embed_mask,
```
achael: I was looking at compute_chisq_dict, compute_chisqgrad_dict again and they don't actually use obslist for anything more than setting the number of observations in the loop. This should be already encoded in both the length of 'logfreqratio_list' and the number of entries with label '_{i}' in the dictionary of 'data_tuples'.  I think it would make sense to edit these two so that we don't pass obslist (it could be a lot of memory, and it's unnecessary)

ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
...
+    )
+    reg_dict = compute_reg_dict(
+        imcur, reg_term_keys, xprior, embed_mask,
+        mf, obslist, logfreqratio_list, pol, norm_reg, regparams,
+    )
+
+    n_obs = len(obslist)
```
achael: again, i think len(obslist) should be the same as len(logfreqratio_list), so could use that here and not pass entire obslist (which is redundant with data_tuples and could be a lot of memory)

Alternatively, maybe the best thing to do would be to pass n_obs explicitly and put in checks in the Imager and relevant functions that len(obs_list), len(freq_list), len(logfreqratio_list) are all the same? Maybe ask for some AI advice on the cleanest thing here.

ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
...
+    datterm = 0.0
+    for dname in dat_term_keys:
+        weight = dat_term[dname]
+        for i in range(n_obs):
+            key = dname if n_obs == 1 else f"{dname}_{i}"
+            chi2 = chi2_dict[key]
+            if chisq_transform:
```
achael: i'd delete chisq_transform, see above.

ehtim/imaging/imager_backend.py
```diff
@@ -852,3 +852,104 @@ def compute_reggrad_dict(imcur, reg_term_keys, xprior, embed_mask,
...
+                datterm = datterm + weight * (chi2 - 1.0)
+
+    regterm = 0.0
+    for regname in reg_term_keys:
+        regterm = regterm + reg_term[regname] * reg_dict[regname]
+
+    return datterm + regterm
+
+
+def compute_objective_grad(imvec, xarr, which_solve, transforms,
```
achael: same comments for this function as above

tests/test_imager_backend.py
```diff
@@ -767,6 +794,310 @@ def test_reg_and_reggrad_share_keys(gauss_im, observe, initialize_imager):
...
+            reg_term={"simple": 1, "flux_mf": 1, "l2_alpha": 1},
+            mf=True, mf_order=1,
+            mf_flux=[im_lo.total_flux(), im_hi.total_flux()],
+        )
+        imvec = imgr._xinit
+        backend_grad = _call_backend_objective_grad(imgr, imvec)
+        method_grad = imgr.objgrad(imvec)
+        np.testing.assert_array_equal(backend_grad, method_grad)
+
+    def test_chisq_transform_branch(self, gauss_im, observe, initialize_imager):
```
achael: could delete if we get rid of chisq_transform

achael (commented):
Looks good overall, just some comments on cleaning up the argument list / removing redundancies. 

Unrelated to what's currently in this PR -- I should have caught it last time -- but it would be great to rename the variable 'priorarr' in the function 'unpack_imarr' to 'init_arr'. This function is taking the solved for 1D variable list 'vec' to a multiD array with all the polarizations/spectral terms, filling in any not-solved-for quantities with their initial values. Critically is replacing these with the initial values, not the 'prior' used in some of the regularizers, which could be something different. Often because the prior and initial image are the same the terminology has been confusing, but it's good to clean things up a bit now.

ehtim/imaging/imager_backend.py
rohandahale: I followed your proposed order (with xarr → initvec and xprior → priorvec). I also extended the rename to the matching Imager attrs (_xarr → _init_arr, _xprior → _prior_arr, _xinit → _init_vec) and the local vars in init_imager (initarr → init_phys + init_solver, making the physical vs solver-space distinction). The naming convention is documented at the top of imager_backend.py and in init_imager's docstring.

ehtim/imaging/imager_backend.py
rohandahale: I added more descriptions, parameters / returns blocks on both compute_objective and compute_objective_grad

ehtim/imaging/imager_backend.py
rohandahale: I confirmed that nothing outside the touched files referenced chisq_transform / chisq_offset_gradient, then deleted: param + branches in compute_objective / compute_objective_grad, the two Imager attrs, and the three obsolete tests. The earlier bugfix commit (a10d040) is now moot, but I left it in the history rather than rebasing

ehtim/imaging/imager_backend.py
rohandahale: All six backend functions now take `n_obs`: int instead of `obslist`. I went with the explicit `n_obs` option.

1. It is JAX-friendly:  natural `static_argname` for `jax.jit`.
2. It decouples chi² compute semantics from freq-list semantics (helpful for Phase 6 / per-channel lists later).
3. Imager.init_imager asserts l`en(obslist_next) == len(freq_list) == len(_logfreqratio_list)` upfront

memray confirms a small RSS reduction (~1 MB) from no longer holding `obslist` references in backend

ehtim/imaging/imager_backend.py
rohandahale: Same fix as above

ehtim/imaging/imager_backend.py
rohandahale: Same fix as above

ehtim/imaging/imager_backend.py
rohandahale: `compute_objective_grad` got the same `n_obs`, params reorder and `chisq_transform` removal

tests/test_imager_backend.py
rohandahale: The three `test_chisq_transform_`* tests (in `TestComputeObjective` and `TestComputeObjectiveGrad`) are deleted now

rohandahale:
@achael thanks for the thorough review! All comments are addressed through the new 5 commits (PR description updated for details). I also worked on the renaming `priorarr` → `init_arr` in `unpack_imarr`.

achael (approved):
looks great! The additional docstrings explaining the different arrays in the solver are super helpful.

PR #233, opened 2026-05-08
Fix _diag chisqdata bugs (NumPy >= 1.24 + np.float)

rohandahale:
Fixes the `_diag` chisqdata + chisqgrad paths that have been broken under NumPy >= 1.24. Roadmap #212 task **2.12**.

## Bugs fixed

1. **`np.float` deprecation** — removed in NumPy 1.24. 6 call sites in `imager_utils.py`:
   - `chisqgrad_cphase_diag` (586), `chisqgrad_logcamp_diag` (723)
   - `chisqgrad_cphase_diag_fft` (983), `chisqgrad_logcamp_diag_fft` (1151)
   - `chisqgrad_cphase_diag_nfft` (1598), `chisqgrad_logcamp_diag_nfft` (1976)

   All replaced with `float(...)`. The inner conversion was redundant since `len()` returns a Python int.

2. **Ragged-array construction** in 6 `chisqdata_*_diag` functions:
   - direct: 2868, 3028
   - fft: 3326, 3542
   - nfft: 3807, 4000

   Per-timestamp baseline counts vary, so the outer arrays are inhomogeneous. NumPy >= 1.24 rejects `np.array(...)` on ragged input without explicit `dtype=object`. Fix applied to all 6 locations + the matching `clphase_diag` / `clamp_diag` / `sigma_diag` return wrappers.

## Test coverage

Extends `tests/test_chisquared.py` `DATATERMS` parametrization to include `cphase_diag` and `logcamp_diag`. The existing `TestChisqConsistency` + `TestChisqGradConsistency` classes automatically cover both new dtypes across all three ttype pairs.

Adds `GRAD_MAX_TOL_DIAG = 0.5` (vs default 0.25) for `_diag` dtypes only — diagonalization orthogonalizes per-timestamp covariance, which amplifies tail outliers in `test_grad_max_frac_diff`. Median tolerance unchanged.

### DIFF
--- a/tests/test_chisquared.py
+++ b/tests/test_chisquared.py
@@ -17,14 +17,23 @@ TSTART_HR = 0
 TSTOP_HR = 24
 BW_HZ = 4e9
 
-# Data types to test
-DATATERMS = ["vis", "bs", "amp", "cphase", "camp", "logcamp"]
+# Data types to test. The `_diag` variants are the per-timestamp diagonalized
+# closures (statistically independent components after orthogonalization).
+DATATERMS = ["vis", "bs", "amp", "cphase", "cphase_diag",
+             "camp", "logcamp", "logcamp_diag"]
 
 # Transform type pairs to compare
 TTYPE_PAIRS = [("direct", "fast"), ("direct", "nfft"), ("nfft", "fast")]
 
 # NFFT max gradient tolerance is much wider at this resolution
 GRAD_MAX_TOL_NFFT = 10.0
+
+# Diagonalized closures orthogonalize per-timestamp covariance, which can
+# amplify direct-vs-fast tail outliers in test_grad_max_frac_diff. Bump the
+# max-fractional-diff tolerance for these dtypes only; median tolerance and
+# chi-squared tolerance unchanged.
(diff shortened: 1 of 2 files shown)


tests/test_chisquared.py
```diff
@@ -17,15 +17,24 @@
...
+DATATERMS = ["vis", "bs", "amp", "cphase", "cphase_diag",
+             "camp", "logcamp", "logcamp_diag"]
 
 # Transform type pairs to compare
 TTYPE_PAIRS = [("direct", "fast"), ("direct", "nfft"), ("nfft", "fast")]
 
 # NFFT max gradient tolerance is much wider at this resolution
 GRAD_MAX_TOL_NFFT = 10.0
 
+# Diagonalized closures orthogonalize per-timestamp covariance, which can
```
achael, line 31: is this only an issue in direct-fast comparison or is it also an issue in direct-nfft? If it's not in direct-nfft maybe we could just increase the tolerance in direct-nfft since this is such a large tolerance....

achael (approved):
Fantastic, it will be interesting to do some tests on imaging with diag closure quantities since I think they have been pretty unused for a long time!

PR #232, opened 2026-05-07
Extract init_imager into pure backend functions

rohandahale:
Extracts `Imager.init_imager` (~263 LOC) into pure backend functions in `imager_backend.py`. Closes roadmap #212 tasks **1C.1**, **1F.1**, and **1E** (combined with #226/#227/#231).

Adds `ImagerInitState` (`typing.NamedTuple`) + `compute_init_state` orchestrator composing `compute_embed` + new helpers `compute_logfreqratios` / `compute_which_solve` / `compute_data_tuples` + `make_initarr` + `transform_imarr_inverse` + `pack_imarr`. `Imager.init_imager` becomes a ~30-line wrapper. `Imager.set_embed` deleted (subsumed by `compute_embed` from #224). Adds 1F.1 end-to-end Gaussian → simulate → reconstruct integration tests across all three ttypes for `pol="I"` and `pol="IP"`.

## Test coverage

53 net new tests; full suite 371 passing (dev-backend baseline 324). Ruff clean. Memray peak RSS 482 MB (+0.22 % vs #231 baseline).

## JAX-readiness

`compute_init_state` is structured for `@partial(jax.jit, static_argnames=(...))` with planned statics: `('pol', 'mf', 'transforms', 'mf_order', 'mf_order_pol', 'mf_rm', 'mf_cm', 'ttype', 'dat_term_keys', 'compute_data')`. NamedTuple return is auto-registered as a JAX pytree.

## Out of scope

- `pol_next='I'` override for non-pol terms in pol mode — candidate for cleanup with task **2.11**.
- `make_initarr` random-pol non-determinism via `np.random.rand` — Phase 5 will replace with `jax.random.PRNGKey`.
- Arbitrary-order multifreq spectral terms in `compute_which_solve` — TODO comment added; Phase 6.

PR #235, opened 2026-05-11
Merge chisq + polchisq dispatchers into unified term-level family

rohandahale:
Closes task **2.11** in #212.

Merges the dual `chisq` + `polchisq` dispatchers into a single term-level family in `imager_backend.py`:

- **`compute_chisq_term`**, **`compute_chisqgrad_term`**, **`compute_chisqdata_term`** — dict-dispatched on `(dtype, ttype)`. 12 dtypes × valid ttypes, parity-tested bit-identically against the legacy paths.
- **`_pol_solve_block(which_solve, pol)`** helper encapsulates the `which_solve[0:4]` slicing pattern flagged as fragile in #227 (with a TODO for the Phase 6 `WhichSolve(stokes, spectral)` NamedTuple refactor).
- **Drops the `pol_next='I'` override** in `compute_data_tuples` flagged in #232 review; the validation moved into `compute_chisqdata_term`.
- **Legacy `chisq`/`chisqgrad`/`chisqdata` + `polchisq`/`polchisqgrad`/`polchisqdata`** become thin shims around the new term functions — back-compat preserved for `Obsdata.chisq`/`polchisq`, plotting, and calibration.
- **Cycle-break**: `embed_imarr` moved from `imager_backend.py` to `imager_utils.py` (next to `embed`) and vectorized over the `nsolve` axis. Bit-identical output to the previous per-row loop.

## JAX-readiness

`_CHISQ_DISPATCH` and `_CHISQDATA_DISPATCH` are the Phase 5 entry points: each entry can be swapped in place to register a JAX backend (e.g. `_CHISQ_DISPATCH['vis']['direct'] = (jax_chisq_vis, jax_chisqgrad_vis)`). `compute_chisq_term` is `jax.jit`-friendly under `static_argnames=('dtype', 'ttype')`.

## Design notes

- `compute_chisqgrad_term` requires explicit `pol_solve` for pol dtypes — no hidden default. The legacy shim still defaults to `POL_SOLVE_DEFAULT` and forwards it through, so external callers of `polchisqgrad(...)` without `pol_solve` are unchanged.
- Pol direct chisqdata leaves (`chisqdata_pvis`/`m`/`vvis`) now accept `**kwargs` so the unified dispatcher passes standard kwargs uniformly across all dtypes.

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -855,9 +855,9 @@ def compute_chisqdata_term(obs, prior, mask, dtype, ttype='direct', pol='I', **k
             raise Exception(f"cannot use dterm {dtype} with pol={pol}")
         pol = 'I'
 
-    # Standard direct + all pol leaves take mask positionally.
-    # Standard fast/nfft leaves omit it (they index uv coords from obs directly).
-    if is_pol or ttype == 'direct':
+    # Only `direct` leaves take mask positionally; fast/nfft leaves index uv
+    # coords from obs directly and ignore the embed mask.
+    if ttype == 'direct':
         return helper(obs, prior, mask, pol=pol, **kwargs)
     return helper(obs, prior, pol=pol, **kwargs)
 
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -894,18 +894,13 @@ def _lookup_chisq_entry(dtype, ttype):
 def compute_chisq_term(imcur, dtype, A, data, sigma, ttype='direct', mask=None):
     """Single chi^2 dispatcher unifying chisq + polchisq.
 
-    Pol dtypes consume the full multi-row imcur ((4, npix) single-frequency
-    or (10, npix) multi-frequency). Standard dtypes consume the Stokes-I
-    row: `imcur[0]` if imcur is 2D (pol-mode imager solving for several
-    Stokes), or imcur itself if 1D (non-pol mode, single Stokes-I solver).
-    Mask embedding and FFT pre-computation match the legacy chisq /
-    polchisq bodies exactly.
-
     Parameters
     ----------
     imcur : np.ndarray
-        Solver-space image. 1D (npix,) in non-pol mode, 2D (nsolve, npix)
-        in pol / mf-pol mode.
+        Solver-space image. Pol dtypes consume the full multi-row array
+        (1D for non-pol mode, 2D (nsolve, npix) for pol/mf-pol with
+        nsolve in {3, 4, 10}). Standard dtypes consume the Stokes-I row:
+        `imcur[0]` when 2D, `imcur` itself when 1D.
     dtype : str
         One of the keys of _CHISQ_DISPATCH (12 dtypes).
     A, data, sigma
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -704,6 +704,100 @@ class ImagerInitState(NamedTuple):
     reffreq: float               # may be re-bound to init.rf when mf=True
 
 
+# Dispatch table for compute_chisqdata_term.
+# Maps (dtype, ttype) to the leaf helper that produces (data, sigma, A).
+# Standard dtypes route to imutils helpers; polarimetric dtypes (pvis/m/vvis)
+# route to polutils helpers. `logamp` aliases `amp` (same data product).
+# Pol dtypes have no `fast` ttype -- polchisqdata does not support FFT.
+_CHISQDATA_DISPATCH = {
+    'vis':          {'direct': imutils.chisqdata_vis,
+                     'fast':   imutils.chisqdata_vis_fft,
+                     'nfft':   imutils.chisqdata_vis_nfft},
+    'amp':          {'direct': imutils.chisqdata_amp,
+                     'fast':   imutils.chisqdata_amp_fft,
+                     'nfft':   imutils.chisqdata_amp_nfft},
+    'logamp':       {'direct': imutils.chisqdata_amp,
+                     'fast':   imutils.chisqdata_amp_fft,
+                     'nfft':   imutils.chisqdata_amp_nfft},
+    'bs':           {'direct': imutils.chisqdata_bs,
+                     'fast':   imutils.chisqdata_bs_fft,
+                     'nfft':   imutils.chisqdata_bs_nfft},
+    'cphase':       {'direct': imutils.chisqdata_cphase,
+                     'fast':   imutils.chisqdata_cphase_fft,
+                     'nfft':   imutils.chisqdata_cphase_nfft},
+    'cphase_diag':  {'direct': imutils.chisqdata_cphase_diag,
+                     'fast':   imutils.chisqdata_cphase_diag_fft,
+                     'nfft':   imutils.chisqdata_cphase_diag_nfft},
+    'camp':         {'direct': imutils.chisqdata_camp,
+                     'fast':   imutils.chisqdata_camp_fft,
+                     'nfft':   imutils.chisqdata_camp_nfft},
+    'logcamp':      {'direct': imutils.chisqdata_logcamp,
+                     'fast':   imutils.chisqdata_logcamp_fft,
+                     'nfft':   imutils.chisqdata_logcamp_nfft},
+    'logcamp_diag': {'direct': imutils.chisqdata_logcamp_diag,
+                     'fast':   imutils.chisqdata_logcamp_diag_fft,
+                     'nfft':   imutils.chisqdata_logcamp_diag_nfft},
+    'pvis':         {'direct': polutils.chisqdata_pvis,
+                     'nfft':   polutils.chisqdata_pvis_nfft},
+    'm':            {'direct': polutils.chisqdata_m,
+                     'nfft':   polutils.chisqdata_m_nfft},
+    'vvis':         {'direct': polutils.chisqdata_vvis,
+                     'nfft':   polutils.chisqdata_vvis_nfft},
+}
+
+
+def compute_chisqdata_term(obs, prior, mask, dtype, ttype='direct', pol='I', **kwargs):
+    """Single chisqdata dispatcher unifying chisqdata + polchisqdata.
+
+    Standard dtypes route to imutils.chisqdata_*; pol dtypes route to
+    polutils.chisqdata_*. Mask handling is asymmetric across the legacy
+    leaves (standard `fast`/`nfft` do not accept a mask param); this
+    function hides that.
+
+    Parameters
+    ----------
+    obs : ehtim.obsdata.Obsdata
+    prior : ehtim.image.Image
+    mask : np.ndarray of bool
+        Embedding mask. Passed positionally to direct + pol-nfft helpers;
+        ignored for standard fast/nfft (those helpers do not take mask).
+    dtype : str
+        Must be a key of _CHISQDATA_DISPATCH (12 dtypes).
+    ttype : {'direct', 'fast', 'nfft'}
+        'fast' is not supported for polarimetric dtypes.
+    pol : str
+        Polarization mode. Standard dtypes use this to unpack the right
+        Stokes (I/Q/U/V) data; pol dtypes absorb it via **kwargs (inert).
+    **kwargs
+        Per-dtype tuning knobs forwarded to the leaf helper. See
+        imager_utils.chisqdata for the standard kwarg list and
+        pol_imager_utils.chisqdata_pvis_nfft for the FFT-related kwargs.
+
+    Returns
+    -------
+    (data, sigma, A) tuple, same shape as the legacy chisqdata dispatchers.
+    """
+    if ttype not in ('direct', 'fast', 'nfft'):
+        raise Exception("Possible ttype values are 'fast', 'direct', 'nfft'!")
+
+    try:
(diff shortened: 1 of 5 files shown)


ehtim/imaging/imager_utils.py
```diff
@@ -368,69 +222,19 @@ def regularizergrad(imvec, nprior, mask, flux, xdim, ydim, psize, stype, **kwarg
...
+    """Return (data, sigma, A) for a standard data term.
+
+    Thin shim around imager_backend.compute_chisqdata_term retained for
+    backward compatibility. New code should call compute_chisqdata_term
+    directly. Returns (False, False, False) (legacy silent fallback) if
+    `dtype` is not a standard data term.
+    """
+    from ehtim.imaging.imager_backend import compute_chisqdata_term
+    ttype = kwargs.pop('ttype', 'direct')
+    if dtype not in DATATERMS:
```
achael, line 231: I think it is fine to raise an explicit exception here, I don't think this fallback is ever used.

ehtim/imaging/imager_utils.py
```diff
@@ -42,179 +42,33 @@
...
-    if mask is None:
-        mask = []
-    chisqgrad = np.zeros(len(imvec))
+    Thin shim around imager_backend.compute_chisqgrad_term retained for
+    backward compatibility. New code should call compute_chisqgrad_term
+    directly. Returns zeros (legacy silent fallback) if `dtype` is not a
+    standard data term.
+    """
+    from ehtim.imaging.imager_backend import compute_chisqgrad_term
     if dtype not in DATATERMS:
```
achael, line 66: maybe better to put in an explicit exception, I don't think this fallback is used.

rohandahale:
@achael resolving merge conflicts now

ehtim/imaging/imager_utils.py
```diff
@@ -42,179 +42,33 @@
...
-    if mask is None:
-        mask = []
-    chisq = 1
+    Thin shim around imager_backend.compute_chisq_term retained for backward
+    compatibility. New code should call compute_chisq_term directly. Returns
+    1 (legacy silent fallback) if `dtype` is not a standard data term.
+    """
+    # Imported here to avoid a module-load cycle with pol_imager_utils.
+    from ehtim.imaging.imager_backend import compute_chisq_term
     if dtype not in DATATERMS:
```
achael, line 52: maybe better to put in an explicit exception

ehtim/imaging/imager_backend.py
```diff
@@ -536,6 +508,282 @@ class ImagerInitState(NamedTuple):
     reffreq: float               # may be re-bound to init.rf when mf=True
 
 
+# `logamp` aliases `amp` (same data product). Pol dtypes have no `fast`.
+_CHISQDATA_DISPATCH = {
```
achael: This looks great, but maybe it makes sense to reorder the imager_backend file so this is at the top of the follow, followed by classes, followed by functions?

rohandahale:
@achael The conflicts are resolved now

ehtim/imaging/imager_backend.py
```diff
@@ -536,6 +508,282 @@ class ImagerInitState(NamedTuple):
...
+    # Standard data terms consume a single Stokes letter ('I'/'Q'/'U'/'V'). When
+    # the imager is in a multi-Stokes mode (e.g. 'IP', 'IQUV'), the leaf needs
+    # pol='I' to unpack Stokes-I visibilities; pol modes without 'I' (e.g. 'P')
+    # are incompatible with standard data terms.
+    if not is_pol and pol in POLARIZATION_MODES:
+        if 'I' not in pol:
+            raise Exception(f"cannot use dterm {dtype} with pol={pol}")
+        pol = 'I'
+
+    # Standard direct + all pol leaves take mask positionally.
```
achael: polarization chisqdata functions with nfft/fast don't actually use mask -- maybe we can remove the arguments in pol_imager_utils so this is more symmetric and the only case when a mask is passed is 'direct'?

ehtim/imaging/imager_backend.py
```diff
@@ -536,6 +508,282 @@ class ImagerInitState(NamedTuple):
...
+    # Standard fast/nfft leaves omit it (they index uv coords from obs directly).
+    if is_pol or ttype == 'direct':
+        return helper(obs, prior, mask, pol=pol, **kwargs)
+    return helper(obs, prior, pol=pol, **kwargs)
+
+
+# _diag leaves index the gridded image directly per-baseline; their `fast`
```
achael: again i'd move these global dispatch variables to the top of the file or a separate file

ehtim/imaging/imager_backend.py
```diff
@@ -536,6 +508,282 @@ class ImagerInitState(NamedTuple):
...
+    if ttype not in entry:
+        raise Exception(f"ttype={ttype!r} not supported for dtype={dtype!r}")
+    chisq_fn, chisqgrad_fn = entry[ttype]
+    return chisq_fn, chisqgrad_fn, entry['is_pol']
+
+
+def compute_chisq_term(imcur, dtype, A, data, sigma, ttype='direct', mask=None):
+    """Single chi^2 dispatcher unifying chisq + polchisq.
+
+    Pol dtypes consume the full multi-row imcur ((4, npix) single-frequency
```
achael: don't think the top part of this docstring is clarifying, i would just remove it. i think the description under imcur in the parameters does the job better of saying that the shape of imcur changes depending on the mode

ehtim/imaging/imager_backend.py
```diff
@@ -536,6 +508,282 @@ class ImagerInitState(NamedTuple):
...
+
+    Returns
+    -------
+    float
+    """
+    chisq_fn, _, is_pol = _lookup_chisq_entry(dtype, ttype)
+
+    imvec = imcur if (is_pol or imcur.ndim == 1) else imcur[0]
+
+    # randomfloor=True breaks TV-regularizer gradient singularities.
```
achael: I think the newer method of breaking up the TV singularities (epsilon_tv) should also be able to help with embedded cases without requiring the randomfloor. For now, can we define a global at the top e.g. EMBED_RANDOMFLOOR=True with a note to revisit, instead of setting this as True here?

ehtim/imaging/imager_backend.py
```diff
@@ -536,6 +508,282 @@ class ImagerInitState(NamedTuple):
...
+    if (ttype != 'direct' and mask is not None
+            and len(mask) > 0 and np.any(np.invert(mask))):
+        if is_pol:
```
achael, line 928: let's put  a TODO at the top to consolidate embed_imarr and embed, I can take a look at this later

ehtim/imaging/pol_imager_utils.py
```diff
@@ -345,91 +345,33 @@ def vcv_grad(imarr, gradarr):
...
-    """
+    """Return chi^2 for a polarimetric data term.
 
-    chisq = 1
+    Thin shim around imager_backend.compute_chisq_term retained for backward
+    compatibility. New code should call compute_chisq_term directly.
+    Returns 1 (legacy silent fallback) if `dtype` is not a polarimetric term.
+    """
+    from ehtim.imaging.imager_backend import compute_chisq_term
     if dtype not in DATATERMS_POL:
```
achael, line 354: maybe better to put in explicit exception

ehtim/imaging/pol_imager_utils.py
```diff
@@ -345,91 +345,33 @@ def vcv_grad(imarr, gradarr):
...
-    """return the chi^2 gradient for the appropriate dtype
+    Thin shim around imager_backend.compute_chisqgrad_term retained for
+    backward compatibility. New code should call compute_chisqgrad_term
+    directly. Returns zeros (legacy silent fallback) if `dtype` is not a
+    polarimetric term.
     """
-
-    chisqgrad = np.zeros(imarr.shape)
+    from ehtim.imaging.imager_backend import compute_chisqgrad_term
     if dtype not in DATATERMS_POL:
```
achael, line 369: explicit exception again

ehtim/imaging/pol_imager_utils.py
```diff
@@ -525,36 +467,19 @@ def polregularizergrad(imarr, priorarr, mask, flux, pflux, vflux, xdim, ydim, ps
...
-            (data, sigma, A) = chisqdata_pvis_nfft(Obsdata, Prior, mask, **kwargs)
-        elif dtype == 'm':
-            (data, sigma, A) = chisqdata_m_nfft(Obsdata, Prior, mask, **kwargs)
-        elif dtype == 'vvis':
-            (data, sigma, A) = chisqdata_vvis_nfft(Obsdata, Prior, mask, **kwargs)
-
-    return (data, sigma, A)
+    from ehtim.imaging.imager_backend import compute_chisqdata_term
+    ttype = kwargs.pop('ttype', 'direct')
+    if dtype not in DATATERMS_POL:
```
achael, line 476: explicit exception

tests/test_imager_backend.py
achael, line 1: this diff is too big for me to comment on on the github review page -- maybe there's another way to put these comments in I don't know about? If the file is getting unwieldy we could split it up into different types of tests

rohandahale: Can you check if you can click on load diff and force load it?

ehtim/imaging/imager_utils.py
rohandahale: Done, now raises exception instead.

ehtim/imaging/imager_utils.py
rohandahale: Done!

ehtim/imaging/imager_utils.py
rohandahale: Done!

ehtim/imaging/pol_imager_utils.py
rohandahale: Done!

ehtim/imaging/pol_imager_utils.py
rohandahale: Done

ehtim/imaging/pol_imager_utils.py
rohandahale: Done

ehtim/imaging/imager_backend.py
rohandahale: Done, `_CHISQDATA_DISPATCH`, `_DIAG_DTYPES`, `_CHISQ_DISPATCH`, and `ImagerInitState` all moved to the top of the file (order is constants --> dispatch tables --> class --> functions).

ehtim/imaging/imager_backend.py
rohandahale: Done, I removed the redundant intro, the `imcur` parameter description now has the shape  info.

ehtim/imaging/imager_backend.py
rohandahale: Done,`EMBED_RANDOMFLOOR = True` lives at the top with a comment noting `epsilon_tv` may make it obsolete, and both call sites use the constant.

ehtim/imaging/imager_backend.py
rohandahale: Added: `# TODO: consolidate `embed` (1D, this function) and `embed_imarr` (1D or 2D, below) into a single implementation -- their bodies overlap.`

ehtim/imaging/imager_backend.py
rohandahale: Done, `chisqdata_pvis_nfft`, `chisqdata_m_nfft`, `chisqdata_vvis_nfft` no longer take a `mask` arg, and the dispatcher in `compute_chisqdata_term` now only passes mask to `ttype == 'direct'`. Single check covers both standard fast/nfft (which never took mask) and pol nfft.

ehtim/imaging/imager_backend.py
rohandahale: Done, `_CHISQDATA_DISPATCH`, `_DIAG_DTYPES`, `_CHISQ_DISPATCH`, and `ImagerInitState` all moved to the top of the file (order is constants --> dispatch tables --> class --> functions).

tests/test_imager_backend.py
achael: for some reason i thought that wasn't working before but it is now ...

rohandahale:
@aviadlevis `embed_imarr` function is moved back to the imager due legacy dependencies but its computation is still in `imager_backend` (`compute_embed`). This will impact your jax backend file.

I investigated embed_imarr. The actual cycle is: `imager_backend` imports `pol_imager_utils` because `transform_imarr` needs polcv/mcv/vcv. Then `pol_imager_utils` needs `embed_imarr` (defined in `imager_backend`), but importing directly would be a hard cycle, so it goes through `ehtim.imager` as a middleman, `from ehtim.imager import embed_imarr`. This currently works only because ehtim/__init__.py's plotting/calibrating chain happens to pre-load pol_imager_utils before imager.py, so the partial import resolves. 

Fix: move polcv/mcv/vcv (and the _r and _grad variants) from `pol_imager_utils` to `imager_backend`. They're general image-basis transforms, not really pol-specific in who should own them. After that, `imager_backend` has zero upward deps, `pol_imager_utils` can import from it cleanly, and the back-door through `ehtim.imager` goes away. ~50 LOC move, no functional change.

rohandahale:
@achael Should be ready to merge now. I resolved the merge conflicts.

achael (approved):
looks good, thanks for making the changes!

PR #238, opened 2026-05-12
Regularizer dispatcher refactor + parametrized FD tests

rohandahale:
> ⚠️ **Split per @achael's request** ([comment](https://github.com/achael/eht-imaging/pull/238#issuecomment-3175862175)):
> the two bug-fix commits (`stv_pol_grad` factor-of-2 + neighbor-roll, and the `vtv2` typo in `REGULARIZERS_POL`) have been removed from this branch and ship separately in **#240** against `dev` for fast release.
> **This PR will have failing CI on `ptv` FD tests until #240 merges to `dev` and `dev` is merged into `dev-backend`.** Rebasing onto the synced `dev-backend` will turn CI green.

Closes roadmap tasks **2.3** (parametrized regularizer FD parity tests) + **2.5** (regularizer dispatcher refactor, mirroring PR #235's chisq pattern).

- **`_REGULARIZER_DISPATCH` + `compute_regularizer_term` / `compute_regularizergrad_term`** in `imager_backend.py`. 35 regularizer names (Stokes-I + pol + mf) routed through one dispatch table.
- **6 legacy outer dispatchers → thin shims** delegating to the term family. Now raise on unknown stype (silent-zero gone). Same pattern as PR #235's chisq shims.
- **`compute_reg_dict` / `compute_reggrad_dict` rewired** to call the term dispatcher directly. Body shrinks from ~110 to ~50 lines.
- **`tests/test_regularizers.py`** parametrized FD parity for all 35 regnames × value + gradient (was 15 Stokes-I only). These are the tests that caught the `stv_pol_grad` bug now in #240.

Legacy outer dispatchers silently returned 0 on unknown stype. Shims now raise. All in-tree callers use validated names; external user scripts that relied on silent-zero may need updates.

## Why CI will be red until #240 syncs

The parametrized pol FD test parametrizes over `REGULARIZERS_POL`, which includes `'ptv'`. With the bug fix removed from this branch, the analytic `stv_pol_grad` is still wrong on the underlying `dev-backend`, so the FD parity test for `'ptv'` will diverge. Once #240 merges to `dev` and `dev` is merged into `dev-backend`, this branch rebases onto the synced base and CI returns to green.

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -214,6 +214,12 @@ _REGULARIZER_DISPATCH = {
         lambda v, *, mask, **kw: -imutils.sflux(v, kw['nprior'], kw['flux'], norm_reg=kw.get('norm_reg', True)),
         lambda v, *, mask, **kw: -imutils.sfluxgrad(v, kw['nprior'], kw['flux'], norm_reg=kw.get('norm_reg', True)),
         'stokes_i'),
+    # flux_mf shares the flux machinery; compute_reg_dict strips '_mf' and
+    # loops over frequencies, but direct callers should still resolve the name.
+    'flux_mf': (
+        lambda v, *, mask, **kw: -imutils.sflux(v, kw['nprior'], kw['flux'], norm_reg=kw.get('norm_reg', True)),
+        lambda v, *, mask, **kw: -imutils.sfluxgrad(v, kw['nprior'], kw['flux'], norm_reg=kw.get('norm_reg', True)),
+        'stokes_i'),
     'simple': (
         lambda v, *, mask, **kw: -imutils.ssimple(v, kw['nprior'], kw['flux'], norm_reg=kw.get('norm_reg', True)),
         lambda v, *, mask, **kw: -imutils.ssimplegrad(v, kw['nprior'], kw['flux'], norm_reg=kw.get('norm_reg', True)),
@@ -1669,6 +1675,27 @@ def compute_chisqgrad_dict(imcur, dat_term_keys,
     return chi2grad_dict
 
 
+def _spectral_slot(regname, n_slots):
+    """Return the imcur row index for a multifrequency spectral regularizer.
+
+    Slot layout: n_slots=3 (Stokes-I + alpha + beta) uses indices 1, 2.
+    n_slots=10 (full IPV + spectral expansion) uses 4-9.
+    """
+    if regname in REGULARIZERS_SPECIND:
+        return 4 if n_slots == 10 else 1
+    if regname in REGULARIZERS_CURV:
+        return 5 if n_slots == 10 else 2
+    if regname in REGULARIZERS_SPECIND_P:
+        return 6
+    if regname in REGULARIZERS_CURV_P:
+        return 7
+    if regname in REGULARIZERS_RM:
+        return 8
+    if regname in REGULARIZERS_CM:
+        return 9
+    raise Exception(f"regularizer term {regname!r} has no spectral slot mapping")
+
+
 def compute_reg_dict(imcur, reg_term_keys,
                      mf, pol,
                      logfreqratio_list, n_obs,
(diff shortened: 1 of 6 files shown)


rohandahale:
@aviadlevis This PR refactors the regularizer dispatchers in `imager_backend.py` the same way #235 did for `chisq`. New entry points: `compute_regularizer_term` / `compute_regularizergrad_term`, backed by `_REGULARIZER_DISPATCH`. The 6 legacy outer functions (`imutils.regularizer` etc.) are now thin wrappers.

rohandahale:
@achael Ready for review!

achael (commented):
Hi Rohan, 

Thanks for catching the stv_pol_grad bug -- this one is actually pretty serious. 

Tracing it back, it was introduced in the multifrequency pol update, when in make_p_image in pol_imager_utils.py, the imager parameter went from the EVPA \chi to 2*\chi. This was pushed to main in v1.2.11. (see the note in line 82 of https://github.com/achael/eht-imaging/blob/v1.2.11/ehtim/imaging/pol_imager_utils.py vs line 425 of https://github.com/achael/eht-imaging/blob/v1.2.10/ehtim/imaging/pol_imager_utils.py). But while the primitive variable changed, the gradient function did not. We should have had a test on this....
(The only good news is that this bug was only pushed to main in April, but unfortunately it has affected the dev branch since June 2024). 

Can you double check that the other pol regularizer gradients and pol chisq gradients are ok, and then can you break this PR up into 2, one that addresses the urgent bug fix and one that addresses everything else? And can you make the bug fix PR to the dev branch as well? 

Glad the tests caught this, we need to fix the dev and main releases ASAP

rohandahale:
Hi @achael,

I checked all pol regularizer gradients and pol chisq gradients for the same `χ → 2χ` pattern and I verified that everything else is fine.

- `chisqgrad_p`, `chisqgrad_m`, and their NFFT variants: math correct. Chain rule explicitly verified for each slot. The `gradi` slot uses `mimage·exp(−1j·φ)` ✓; `gradrho` multiplies by `cos(ψ)` ✓; `gradphi` is `0.5·(−2·Im[pimage*·…])/N = −Im[pimage*·…]/N` ✓ (the local `gradchi` variable name is misleading — it's the gradient w.r.t. `χ`, scaled by `dχ/dφ = 0.5` to land on the imarr primitive `φ`); `gradpsi` uses the `−mimage·tan(ψ)` chain factor ✓.
- `chisqgrad_vvis`, `chisqgrad_vvis_nfft`, `smgrad`, `shwgrad`, all V-pol regularizers: no φ/χ dependence → unaffected.
- A grep for `cos(2*` / `sin(2*` in `pol_imager_utils.py` confirmed only `stv_pol_grad` had the suspicious pattern in gradient code; the remaining hits in `plot_m` are the legitimate forward `Q = I·m·cos(2χ)`.

The two bug-fix commits (`stv_pol_grad` math + neighbor-roll, and the `vtv2` typo) now go in #240 against `dev` so they can land in main quickly. This PR has been rebased onto `dev-backend` with those two commits dropped; the dispatcher refactor + parametrized FD tests remain.

CI on this PR will fail on the `ptv` FD parity test until #240 lands in `dev` and `dev` is merged into `dev-backend`. Happy to wait on that.

rohandahale:
@achael I have rebased this PR and it is read for review.

ehtim/imaging/imager_backend.py
```diff
@@ -154,6 +154,257 @@
 }
 
 
+# Helpers for compound regularizers: tvlog / tv2log apply log to imvec before
+# calling stv / stv2 and adjust the gradient via the chain rule from log.
+def _tvlog_value(v, *, mask, **kw):
```
achael: I think it would be better and more parallel to just make new functions in imager_utils.py for the tvlog, tv2log functions: 

stvlog(imvec, xdim, ydim, psize, flux, norm_reg=norm_reg,beam_size=beam_size, epsilon=epsilon)
stv2log(imvec, xdim, ydim, psize, flux, norm_reg=norm_reg,beam_size=beam_size, epsilon=epsilon)
and their associated gradient functions. These should call the appropriate stv, stv2 functions

We may want to address the format of arguments to all the regularizers later, but for now I think it is better to keep them all with parallel structure.

ehtim/imaging/imager_backend.py
```diff
@@ -154,6 +154,257 @@
...
+    logflux = npix * np.abs(np.log(kw['flux'] / npix))
+    if np.any(np.invert(mask)):
+        v = imutils.embed(v, mask, randomfloor=True)
+    g = -imutils.stv2grad(np.log(v), kw['xdim'], kw['ydim'], kw['psize'], logflux,
+                          norm_reg=kw.get('norm_reg', True),
+                          beam_size=kw.get('beam_size'))
+    return (g / v)[mask]
+
+
+# Dispatch table: regname -> (value_fn, grad_fn, family).
```
achael, line 157: i understand the structure of this dispatch table, but it's much more confusing to read than the corresponding chisq table, given the more complicated argument and unpacking/slicing structure. Unlike the chisq case, I think it's actually somewhat more confusing than the existing function with branching paths. 

I think it's probably ok, but I would play around with Claude and see if there is a way to make this more transparent, maybe by defining some more helper functions that can be used by multiple regularizers, if necessary. For example, I might favor creating an explicit wrapper in imager_utils (pol_imager_utils, multifreq_imager_utils) for every regularizer and regularizer grad function that all have the same argument structure (imarr, mask, **kwargs) and apply any necessary pre-post-processing for imaging. Basically, move everything that is written in-line in this dictionary to one associated wrapper for each regularizer that is right next to it in the appropriate file. Then maybe in the future we can collapse the wrapper and the original function. 

That's my thought on how to make this more readable, but I'm open to some other ideas. I just don't like the current way everything is written in-line in the dict

ehtim/imaging/imager_backend.py
```diff
@@ -1369,6 +1675,27 @@ def compute_chisqgrad_dict(imcur, dat_term_keys,
     return chi2grad_dict
 
 
+def _spectral_slot(regname, n_slots):
```
achael: this is nice, maybe it should go in multifreq_imager_utils for now? I'm still not sure of the eventual global structure of all these files, but for now since this is firmly a multifreq-only function I think it makes sense there.

achael (commented):
looks good overall, but i have some suggestions on making the dispatch table a little less cumbersome

ehtim/imaging/imager_backend.py
rohandahale: I added `stvlog` / `stvloggrad` / `stv2log` / `stv2loggrad` as parallels to `stv` / `stvgrad` / `stv2` / `stv2grad` in `imager_utils.py`. Same signature shape, forward to `stv` / `stv2` with the log-transformed image and the chain rule applied via `1/imvec`.

ehtim/imaging/imager_backend.py
rohandahale: I added uniform `(imvec_or_imarr, mask, **kwargs)` wrappers in `imager_utils.py` / `pol_imager_utils.py` / `multifreq_imager_utils.py` , one `reg_X` + `reggrad_X` pair per regularizer, right next to the originals. The dispatch table in `imager_backend.py` is now a flat one line per entry mapping pointing at the wrappers. 

I removed the inline `_tvlog_value` / `_tvlog_grad` / `_tv2log_value` / `_tv2log_grad` / `_slice_pol_grad` helpers.

ehtim/imaging/imager_backend.py
rohandahale: I moved `spectral_slot` from `imager_backend.py` to `multifreq_imager_utils.py`.

rohandahale:
@achael The new commit addresses all three comments. All 753 tests passing and ruff is clean. Ready for re-review.

ehtim/imaging/imager_backend.py
achael: great, this looks much cleaner :)

ehtim/imaging/imager_utils.py
```diff
@@ -2168,6 +2050,51 @@ def stv2grad(imvec, nx, ny, psize, flux, norm_reg=NORM_REGULARIZER, beam_size=No
     return out/norm
 
 
+def stvlog(imvec, nx, ny, psize, flux, norm_reg=NORM_REGULARIZER, beam_size=None, epsilon=0.):
```
achael, line 2053: great, thanks for making this change!

ehtim/imaging/imager_utils.py
```diff
@@ -2392,6 +2319,230 @@ def sgauss_grad(imvec, xdim, ydim, psize, major, minor, PA):
...
+#
+# Each `reg_X` / `reggrad_X` adapts an `sX` / `sXgrad` regularizer above to the
+# uniform `(imvec, mask, **kwargs)` signature used by the `_REGULARIZER_DISPATCH`
+# table in `imager_backend.py`. The wrappers add the sign convention (the
+# imager minimizes `-regularizer`), unpack kwargs into the leaf's positional
+# args, and handle the embed-pre / mask-post-slice pattern for spatial regs.
+##################################################################################################
+
+
+def reg_flux(imvec, mask, **kwargs):
```
achael, line 2333: the reg_ vs s prefix here is nice, especially as this fixes the sign convention. 

Let's make a TODO to see if we can remove the original 's' functions at some point and collapse them into these. We just need to make sure they aren't used on their own elsewhere in the code.

ehtim/imaging/imager_utils.py
```diff
@@ -2392,6 +2319,230 @@ def sgauss_grad(imvec, xdim, ydim, psize, major, minor, PA):
...
+    return -spatch(imvec, kwargs['nprior'], kwargs['flux'],
+                   norm_reg=kwargs.get('norm_reg', True))
+
+
+def reggrad_patch(imvec, mask, **kwargs):
+    return -spatchgrad(imvec, kwargs['nprior'], kwargs['flux'],
+                       norm_reg=kwargs.get('norm_reg', True))
+
+
+# scm / scmgrad take the embed mask as a positional and handle masking internally,
```
achael, line 2405: re this comment, we can make this more uniform later

achael (approved):
looks great -- let's remember to see if we can eventually collapse the original 's' regularizer functions into the new 'reg_' wrappers.

rohandahale:
Thanks for the merge! Filed both follow-ups on the tracking issue:

- **#212 task 2.26** — collapse `sX` regularizer functions and their `reg_X` wrappers into single functions with the uniform interface (with the codebase audit of `sX` external uses first).
- **#212 task 2.27** — uniformize `scm` / `scmgrad` to match the embed-pre / mask-post-slice pattern of the other spatial regularizers.

Will pick them up after the current NamedTuple work (#239) lands.

PR #239, opened 2026-05-13
Promote config bundles to NamedTuples

rohandahale:
Promotes the two remaining flat-dict config helpers (`_full_data_weighting_params`, `_full_fft_params`) into typed NamedTuples, and bundles the 8 flat `*_next` / `_ttype` attrs on `Imager` into a single nested `ImagerConfig` containing an `MfConfig`. Every backend function that previously took individual `pol` / `mf` / `ttype` / `transforms` / `mf_*` args (8 functions) now takes one `config: ImagerConfig` instead. `Imager.make_image()` parameter overrides switch from attr mutation to `self._config._replace(...)`. `_append_image_history()` keeps the flat `ImagerRunState` schema (per design decision) but reads through `self._config`.

The 4 bundles (`DataWeighting`, `FourierGridParams`, `MfConfig`, `ImagerConfig`) are `typing.NamedTuple`-based — pytree-friendly for Aviad's JAX mirror, with the static-argname caveat documented on `ImagerConfig` (string leaves + `snrcut: dict` aren't traceable values).

## Roadmap tasks closed

- **2.18** — `_full_data_weighting_params()` → `DataWeighting` NamedTuple
- **2.19** — `_full_fft_params()` → `FourierGridParams` NamedTuple (renamed from `FFTParams` since `ttype='fast'` is legacy; the bundle holds grid params shared by `fast` + `nfft`)
- **2.21** — `ImagerConfig` (nested `MfConfig`) replaces 8 flat `*_next` attrs
- **2.22** — 60 access sites in `imager.py` migrated to `self._config.X` / `self._config.mf_config.X`
- **2.23** — flat `pol_next` / `transform_next` / `_ttype` / `mf_next` / `mf_order` / `mf_order_pol` / `mf_rm` / `mf_cm` attrs dropped; no shims (pre-2.0, per project decision)

## Design decisions

1. **Nested `MfConfig` inside `ImagerConfig`** — JAX-pytree cohesion + future spectral-mode extension headroom.
2. **Full backend migration** — every backend function consuming any config field takes `ImagerConfig`, not partial unpacking. 8 signatures total.
3. **`_replace()` for `make_image()` overrides** — idiomatic NamedTuple update; preserves immutability.
4. **`ImagerRunState` schema kept flat** — `_append_image_history()` reads `self._config.*` but writes the same per-field keys as before, so all `*_last()` accessors continue to work without churning consumers.
5. **NamedTuple field names match `Imager(...)` kwarg names** for grep-ability.

## Surface changes worth flagging

- `Imager(...)` kwargs (`pol=`, `transform=`, `ttype=`, `mf=`, `mf_order=`, etc.) stay back-compat; they continue to construct `ImagerConfig` internally.
- Internal `self.X_next` reads are gone — third-party code that read `imgr.pol_next`/etc. directly will `AttributeError`. Pre-2.0; expected per the JAX-first design decision.
- `polchisqdata` shim in `pol_imager_utils.py` now consumes a `pol` kwarg (previously absorbed inert via `**kwargs`). Used only to populate the synthetic `ImagerConfig.pol` field; behavior preserved for pol dtypes since `compute_chisqdata_term` ignores `pol` for them. Standard dtypes already raised — no regression.

### DIFF
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -37,6 +37,7 @@ from ehtim.imaging.imager_backend import (
     DATATERMS,
     DATATERMS_POL,
     POLARIZATION_MODES,
+    DataWeighting,
     RegParams,
     compute_chisq_dict,
     compute_chisqgrad_dict,
@@ -751,16 +752,16 @@ class Imager:
         )
 
     def _full_data_weighting_params(self):
-        """Bundle data-weighting params into a single dict for the backend."""
-        return {
-            'maxset': self.maxset_next,
-            'debias': self.debias_next,
-            'snrcut': self.snrcut_next,
-            'weighting': self.weighting_next,
-            'systematic_noise': self.systematic_noise_next,
-            'systematic_cphase_noise': self.systematic_cphase_noise_next,
-            'cp_uv_min': self.cp_uv_min,
-        }
+        """Bundle data-weighting params into a DataWeighting NamedTuple for the backend."""
+        return DataWeighting(
+            maxset=self.maxset_next,
+            debias=self.debias_next,
+            snrcut=self.snrcut_next,
+            weighting=self.weighting_next,
+            systematic_noise=self.systematic_noise_next,
+            systematic_cphase_noise=self.systematic_cphase_noise_next,
+            cp_uv_min=self.cp_uv_min,
+        )
 
     def _full_fft_params(self):
         """Bundle FFT/NFFT params into a single dict for the backend."""
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -190,7 +190,7 @@ class Imager:
         self.beam_size = self.obslist_next[0].res()
         self.regparams = {k: kwargs.get(k, REGPARAMS_DEFAULT[k]) for k in REGPARAMS_DEFAULT.keys()}
 
-        # FFT / Fourier-grid parameters (consumed by the backend via self._full_fft_params())
+        # FFT / Fourier-grid parameters (consumed by the backend via self._fft_params())
         ttype = kwargs.get('ttype', 'nfft')
         self._fft_gridder_prad = kwargs.get('fft_gridder_prad', GRIDDER_P_RAD_DEFAULT)
         self._fft_conv_func = kwargs.get('fft_conv_func', GRIDDER_CONV_FUNC_DEFAULT)
@@ -699,7 +699,7 @@ class Imager:
             self.freq_list, self.reffreq, self._config,
             self.norm_init, self.flux_next, self.clipfloor_next,
             sorted(self.dat_term_next.keys()),
-            self._full_data_weighting_params(), self._full_fft_params(),
+            self._data_weighting_params(), self._fft_params(),
             compute_data=self._change_imgr_params,
             prior_data_tuples=getattr(self, "_data_tuples", None),
         )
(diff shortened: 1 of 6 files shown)


rohandahale:
@achael This PR is also rebased and ready for review!

ehtim/imaging/imager_backend.py
```diff
@@ -1007,8 +997,91 @@ class RegParams(NamedTuple):
     epsilon_tv: float
 
 
-def compute_data_tuples(obslist, prior, embed_mask, dat_term_keys, pol,
-                        ttype, data_weighting_params, fft_params):
+class DataWeighting(NamedTuple):
```
achael: I think we should move the classes to the top -- maybe the imager_backend is big enough now that we should have one file for classes/globals and one for functions? Not sure what best practice there would be.

ehtim/imaging/imager_backend.py
```diff
@@ -1007,8 +997,91 @@ class RegParams(NamedTuple):
...
+    forwarded into compute_data_tuples, which reads its fields directly.
+
+    Field groups:
+      Closure-set selection   : maxset
+      Noise corrections       : debias, systematic_noise, systematic_cphase_noise
+      SNR filtering           : snrcut
+      Weighting scheme        : weighting
+      UV closure-phase filter : cp_uv_min
+    """
+    maxset: bool                    # use the maximal independent set of closure quantities
```
achael: general question: is there a reason to not give these classes default values for some of these params? Is it safer to make sure things are initialized every time?

ehtim/imager.py
```diff
@@ -751,33 +758,32 @@ def _full_regparams(self):
...
+            maxset=self.maxset_next,
+            debias=self.debias_next,
+            snrcut=self.snrcut_next,
+            weighting=self.weighting_next,
+            systematic_noise=self.systematic_noise_next,
+            systematic_cphase_noise=self.systematic_cphase_noise_next,
+            cp_uv_min=self.cp_uv_min,
+        )
 
     def _full_fft_params(self):
```
achael: do these need the 'full' prefix?

achael (commented):
looks great overall, just a few minor organizational suggestions.

ehtim/imaging/imager_backend.py
rohandahale: Done! All 5 NamedTuples (`RegParams`, `DataWeighting`, `FourierGridParams`, `MfConfig`, `ImagerConfig`) now live right after `ImagerInitState` near the top of `imager_backend.py`, so the file goes imports → constants → dispatch tables → classes → functions. On the file-split question (one file for classes/globals, one for functions): I woud lean against it for now, the current file is ~1700 LOC which is manageable, and a split would force every backend caller to choose between two import paths. If it grows more then we can revisit.

ehtim/imager.py
rohandahale: I dropped it now. `_full_data_weighting_params` → `_data_weighting_params`, `_full_fft_params` → `_fft_params`, `_full_regparams` → `_regparams`. The `_full_` prefix was leftover from when these returned dict bundles, now that they are single purpose NamedTuple constructors, the `full` can be dropped.

ehtim/imaging/imager_backend.py
rohandahale: The defaults already live on the Imager kwarg API (`Imager(fft_pad_factor=..., ...)`). The NamedTuples are internal only bundles, every field is set explicitly by `Imager._data_weighting_params()` / etc. Putting defaults on the NamedTuple too would hide bugs: if a future Imager change forgets to set a field, we would get the default instead of a TypeError.

rohandahale:
@achael Thanks for the review!  I have addressed all three comments.

achael (approved):
looks great, on to the next phase!

PR #240, opened 2026-05-13
Fix factor-of-2 bug in stv_pol_grad gradient

rohandahale:
Two bug fixes in `pol_imager_utils.py` extracted from PR #238 per @achael's request, targeted at `dev` so they can land in main quickly.

## 1. `stv_pol_grad` factor-of-2 + neighbor-roll bug (the urgent one)

**Root cause**: when the multifrequency pol update changed the imager primitive in `make_p_image` from EVPA `χ` to `φ = 2χ` (pushed to `main` in v1.2.11 in April 2026, on `dev` since June 2024), the matching update to `stv_pol_grad` was missed. Every analytic gradient slot in `stv_pol_grad` had `cos(2·(angle(P_l1) − angle(P)))` and `sin(2·(angle(P_l1) − angle(P)))`, but since `angle(P) = φ` is already the new primitive, the gradient should be `cos(angle(P_l1) − angle(P))` and `sin(angle(P_l1) − angle(P))` — no factor of 2.

Net impact: every imaging run using `ptv` (polarimetric total variation) regularization since the `make_p_image` change has been optimizing against a wrong gradient. Finite-difference vs analytic median fractional diff was ~0.45 before this fix, ~4e-6 after.

**Secondary bug, same function:** `im_l1r2` was a copy-paste duplicate of `im_r1l2` (both used `np.roll(impad, 1, axis=0), -1, axis=1`). Correct rolls for `im_l1r2` are the opposite — `-1, axis=0` and `+1, axis=1`. Visible at the `d3` denominator in the same function.

## 2. `vtv2` typo in `REGULARIZERS_POL`

`REGULARIZERS_POL = ['msimple', 'hw', 'ptv', 'l1v', 'l2v', 'vtv', 'v2tv2', 'vflux']` had `'v2tv2'` where the dispatcher elsewhere uses `'vtv2'`. Anyone trying `reg_term={'vtv2': 1}` would get a "not recognized" path because the name wasn't in the list. 

## Audit done

Double-checked the other pol regularizer gradients and pol chisq gradients for similar staleness against the `χ → 2χ` primitive change.

### DIFF
--- a/ehtim/imaging/pol_imager_utils.py
+++ b/ehtim/imaging/pol_imager_utils.py
@@ -1014,44 +1014,47 @@ def stv_pol_grad(imarr, flux, nx, ny, psize, pol_solve=POL_SOLVE_DEFAULT,
     im_r1 = np.roll(impad, 1, axis=0)[1:ny+1, 1:nx+1]
     im_r2 = np.roll(impad, 1, axis=1)[1:ny+1, 1:nx+1]
     im_r1l2 = np.roll(np.roll(impad, 1, axis=0), -1, axis=1)[1:ny+1, 1:nx+1]
-    im_l1r2 = np.roll(np.roll(impad, 1, axis=0), -1, axis=1)[1:ny+1, 1:nx+1]
-    
+    im_l1r2 = np.roll(np.roll(impad, -1, axis=0), 1, axis=1)[1:ny+1, 1:nx+1]
+
     # Denominators
     d1 = np.sqrt(np.abs(im_l1 - im)**2 + np.abs(im_l2 - im)**2)
     d2 = np.sqrt(np.abs(im_r1 - im)**2 + np.abs(im_r1l2 - im_r1)**2)
     d3 = np.sqrt(np.abs(im_r2 - im)**2 + np.abs(im_l1r2 - im_r2)**2)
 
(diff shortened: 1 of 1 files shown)


rohandahale:
@achael Here is the urgent bug fix for review.

ehtim/imaging/pol_imager_utils.py
```diff
@@ -1014,44 +1014,47 @@ def stv_pol_grad(imarr, flux, nx, ny, psize, pol_solve=POL_SOLVE_DEFAULT,
     im_r1 = np.roll(impad, 1, axis=0)[1:ny+1, 1:nx+1]
     im_r2 = np.roll(impad, 1, axis=1)[1:ny+1, 1:nx+1]
     im_r1l2 = np.roll(np.roll(impad, 1, axis=0), -1, axis=1)[1:ny+1, 1:nx+1]
-    im_l1r2 = np.roll(np.roll(impad, 1, axis=0), -1, axis=1)[1:ny+1, 1:nx+1]
-    
+    im_l1r2 = np.roll(np.roll(impad, -1, axis=0), 1, axis=1)[1:ny+1, 1:nx+1]
```
achael, line 1017: this bug looks like it's been in place for much longer....

achael (commented):
thanks for catching this

PR #249, opened 2026-05-15
Fix three latent bugs in Obsdata

achael:
## Summary

Re-land of the three obsdata.py one-liner bugfixes from previously-merged PR #247. That PR merged into `refactor/remove-closure-table-caching` (commit `e6a84df`), but those commits never made it into `dev-backend` — the closure-cleanup squash-merge (`fdf15ed`) was created from an earlier snapshot. This PR cherry-picks `e6a84df` onto current `dev-backend` to restore the fixes.

The three bugs:

- **`reorder_baselines_trial_speedups`** ([obsdata.py:456](ehtim/obsdata.py#L456)) — `for j in len(uniqdat):` iterates over an `int`, raising `TypeError`. Should be `range(len(uniqdat))`. The branch only runs when duplicate `(time, t1, t2)` rows exist, which the default `reorder_baselines` code path prevents — that's why it has never blown up in practice.
- **`deblur`** ([obsdata.py:2320](ehtim/obsdata.py#L2320)) — `vis2[i] = vis3[i] / ker` overwrites the second visibility slot with the (divided) third instead of dividing the third in place. The result is silent corruption of slot 3 (`uvis` on stokes, `rlvis` on circ) whenever anyone called `obs.deblur()`. Fixed to `vis3[i] = vis3[i] / ker`.
- **`plot_bl`** ([obsdata.py:3926](ehtim/obsdata.py#L3926)) — the invalid-field error path called `string.join(ehc.FIELDS)`, but `string` is not imported anywhere else in the file (it's a Python 2 idiom). Fixed to `' '.join(ehc.FIELDS)`, matching the sibling code path in `plotall`. Also drops the now-unused `import string` at the top of the file.

## Test plan

- [x] `ruff check ehtim/obsdata.py` → `All checks passed!`
- [x] `pytest tests/test_obsdata.py` → 37 passed
- [x] Andrew/Rohan: visual scan of the three lines. Behavioural test coverage for the surrounding methods is in the companion `feature/obsdata-method-tests` PR which rebases onto this one.

### DIFF
--- a/ehtim/calibrating/self_cal.py
+++ b/ehtim/calibrating/self_cal.py
@@ -17,27 +17,21 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from __future__ import division
-from __future__ import print_function
 
-from builtins import str
-from builtins import range
-from builtins import object
+import copy
+import time
+import warnings
+from multiprocessing import Pool, cpu_count
 
 import numpy as np
 import scipy.optimize as opt
-import time
-import copy
-from multiprocessing import cpu_count, Pool
 
-import ehtim.obsdata
-import ehtim.parloop as parloop
-from . import cal_helpers as calh
-from ehtim.observing.obs_simulate import add_jones_and_noise
-import ehtim.observing.obs_helpers as obsh
 import ehtim.const_def as ehc
+import ehtim.obsdata
+import ehtim.observing.obs_helpers as obsh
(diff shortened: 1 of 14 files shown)


achael:
had to redo this PR because of an issue syncing with dev-backend

rohandahale (dismissed):
Looks good! The diff matches the original `e6a84df`.

PR #246, opened 2026-05-14
Remove cached closure tables from Obsdata

achael:
Removes the cached closure-table state on `Obsdata` (`amp`, `bispec`, `cphase`, `cphase_diag`, `camp`, `logcamp`, `logcamp_diag`) and the machinery that kept it in sync. Part of [#212](https://github.com/achael/eht-imaging/issues/212).

## What changed

- **`obsdata.py`** — drop the 7 cache slots, 8 `add_*` methods, and `force_recompute` kwargs from `bispectra_tri` / `cphase_tri` / `camp_quad` / `plot_cphase` / `plot_camp` (only existed to bypass the cache).
- **`imager_utils.py` + `modeling_utils.py` `chisqdata_*`** — all 15 amp/bs/cphase/camp/logcamp variants (DFT/FFT/NFFT) had an `if Obsdata.<table> is None: compute else: use cached` branch. Collapses to the unconditional compute path.
- **`applycal` / `self_cal`** — drop `copy_closure_tables`.
- **`comp_plots.py`** — `plot_cphase_compare` / `plot_camp_compare` used to monkey-patch `obs.cphase` / `obs.camp`. Now pass closure data through the existing `Obsdata.plot_cphase` / `plot_camp` kwargs.
- **`image.py:pad_image_to_imageobs`** — inline `obs.unpack(['amp','sigma'], debias=...)` instead of the `add_amp()` side-effect pattern.
- **`io/save.py`, `io/load.py`** — `save_dtype_txt(obs, fname, data, dtype=...)` and `load_dtype_txt(filename, dtype=...) -> np.recarray`. Data flows explicitly instead of via the cache.
- **`examples/example_im_closure.py`** — drop the four pre-imaging `obs.add_*` calls (dead weight now).
- **`tests/test_obsdata.py`** (new) — idempotence on `bispectra()` / `c_phases()` / `c_amplitudes()` across all standard `vtype`s and both `ctype`s. New `obs_pol_direct` conftest fixture.

## Design decisions to flag

1. `save_dtype_txt(obs, fname, data, dtype=...)` is the only user-visible API break — `data` is now a required positional. Old form pulled `obs.<table>` implicitly.
2. `load_dtype_txt` returns the recarray instead of writing it back to `obs.<table>`.
3. The eight `add_*` methods are removed outright (callers see `AttributeError`). Worth deprecation shims pointing at `obs.bispectra()` / `obs.c_phases()` / `obs.c_amplitudes()` / `obs.unpack(['amp','sigma'])` instead?

PR #252, opened 2026-05-15
Add pandas-free ehtim.statistics.averaging + fix inverse-variance time-averaging

rohandahale:
Moves the three visibility-averaging routines (`coh_avg_vis`, `coh_moving_avg_vis`, `incoh_avg_vis`) from `ehtim.statistics.dataframes` (pandas) to a new pandas-free `ehtim.statistics.averaging` (pure NumPy).
Bundles the inverse-variance time-averaging fix,  same class of bug as PR #230.

## What changed

- **New** `ehtim/statistics/averaging.py` — pure-NumPy implementations vectorised via `np.bincount` / cumsum + `searchsorted`.
- **Bug fix:** sigma combination switched from `sqrt(mean(sigma_i**2))` (wrong) to `1/sqrt(sum(1/sigma_i**2))` (correct inverse-variance), mirroring the channel/IF averaging in `load_obs_uvfits` (PR #230).
- **`ehtim/obsdata.py`** — `avg_coherent` / `avg_incoherent` / `add_amp` rewired to the new module.
- **`ehtim/statistics/dataframes.py`** — pandas now lazy-imported (via a  `_PandasStub` that raises a clear `ImportError` on first attribute access) so users without pandas can still `import ehtim`. Closure-quantity helpers (`make_*_df`, `average_*`) remain; they're called only by `Obsdata.add_*` methods that PR #246 removes on
  `dev-backend`.  Once #246 syncs through to `dev`, `dataframes.py` can be deleted in a follow-up PR.

## Behavior change

Inverse-variance sigmas are systematically smaller and more physical than the old formula.  Pipelines with tuned SNR cuts may need to retune; same class of behavior change as PR #230.

## API breaks

- `incoh_avg_vis(..., return_type='df')` and `coh_avg_vis(..., return_type='df')` (and the moving variant) now raise `ValueError`; the new module only emits recarrays.  `Obsdata.add_amp` raises `NotImplementedError` if a
  user requests `return_type='df'` for the averaging path.  Pre-2.0 internal break; no external pipelines should be affected.

Part of #212 — closes tasks 2.7 and 2.17.

### DIFF
--- a/ehtim/statistics/averaging.py
+++ b/ehtim/statistics/averaging.py
@@ -1,0 +1,409 @@
+"""Visibility time-averaging.
+
+Pure-NumPy implementations of the three averaging routines historically
+provided by ``ehtim.statistics.dataframes``:
+
+- :func:`coh_avg_vis` — coherent (complex) averaging into fixed time bins
+  or per scan.
+- :func:`coh_moving_avg_vis` — coherent moving-window averaging.
+- :func:`incoh_avg_vis` — incoherent (amplitude) averaging with Rice
+  debiasing.
+
+All three propagate visibility errors via the inverse-variance combination
+
+.. math:: \\sigma_{\\rm avg} = 1 / \\sqrt{\\sum_i 1/\\sigma_i^2}.
+
+This matches the channel/IF averaging in :func:`ehtim.io.load.load_obs_uvfits`
+(see PR #230) and replaces the historical ``sqrt(mean(sigma_i**2))`` formula in
+``dataframes.py`` which underestimated errors and inflated SNRs.
+"""
+
+import numpy as np
+
+import ehtim.const_def as ehc
+from ehtim.statistics.stats import bootstrap, mean_incoh_avg
+
+__all__ = ["coh_avg_vis", "coh_moving_avg_vis", "incoh_avg_vis"]
+
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+
+def _polrep_fields(polrep):
+    """Return (vis_fields, sig_fields, out_dtype) for a given polrep."""
+    if polrep == "stokes":
+        return (
+            ("vis", "qvis", "uvis", "vvis"),
+            ("sigma", "qsigma", "usigma", "vsigma"),
+            ehc.DTPOL_STOKES,
+        )
+    if polrep == "circ":
+        return (
+            ("rrvis", "llvis", "rlvis", "lrvis"),
+            ("rrsigma", "llsigma", "rlsigma", "lrsigma"),
+            ehc.DTPOL_CIRC,
+        )
+    raise ValueError(f"unsupported polrep {polrep!r}")
+
+
+def _assign_scan_bin(times_hr, scans):
+    """Assign each row in `times_hr` to a scan index (or -1 if outside)."""
+    out = np.full(len(times_hr), -1, dtype=np.int64)
+    for idx, scan in enumerate(scans):
+        tstart, tstop = scan[0], scan[1]
+        out[(times_hr >= tstart) & (times_hr <= tstop)] = idx
+    return out
+
+
+def _group_ids(*keys):
+    """Map a row's tuple of key values to a contiguous group index.
+
+    Returns ``(gids, n_groups)`` where ``gids[i]`` is the group index of row i.
+    """
+    n = len(keys[0])
+    if n == 0:
+        return np.empty(0, dtype=np.int64), 0
+    dtype = [(f"k{i}", k.dtype) for i, k in enumerate(keys)]
+    rec = np.empty(n, dtype=dtype)
+    for i, k in enumerate(keys):
+        rec[f"k{i}"] = k
+    _, gids = np.unique(rec, return_inverse=True)
+    gids = gids.astype(np.int64, copy=False)
+    return gids, int(gids.max()) + 1
+
+
+def _first_index_per_group(gids, n_groups):
+    """Return an array of length n_groups giving one representative row index
+    (the first occurrence) for each group."""
+    out = np.full(n_groups, -1, dtype=np.int64)
+    for i, g in enumerate(gids):
+        if out[g] == -1:
+            out[g] = i
+    return out
+
+
+def _inverse_variance_sigma(sig_per_row, gids, n_groups):
+    """Reduce a per-row sigma array to per-group inverse-variance sigma.
+
+    For each group g, the result is ``1 / sqrt(sum_i 1/sigma_i**2)`` over rows
+    i in group g with finite, positive sigma.  Groups with no valid rows get
+    ``NaN``.
+    """
+    finite = np.isfinite(sig_per_row) & (sig_per_row > 0)
+    inv_var = np.zeros_like(sig_per_row, dtype=np.float64)
+    inv_var[finite] = 1.0 / sig_per_row[finite] ** 2
+    sums = np.bincount(gids, weights=inv_var, minlength=n_groups)
+    out = np.full(n_groups, np.nan)
+    out[sums > 0] = 1.0 / np.sqrt(sums[sums > 0])
+    return out
+
+
+def _mean_complex(vis_per_row, gids, n_groups):
+    """Reduce a per-row complex visibility array to per-group complex mean.
+
+    Skips NaN entries.  Groups with no finite rows get ``NaN + NaN*1j``.
+    """
+    finite = np.isfinite(vis_per_row.real) & np.isfinite(vis_per_row.imag)
+    re = np.where(finite, vis_per_row.real, 0.0)
+    im = np.where(finite, vis_per_row.imag, 0.0)
+    sums_re = np.bincount(gids, weights=re, minlength=n_groups)
+    sums_im = np.bincount(gids, weights=im, minlength=n_groups)
+    counts = np.bincount(gids, weights=finite.astype(np.float64), minlength=n_groups)
+    out = np.full(n_groups, np.nan + 1j * np.nan, dtype=np.complex128)
+    valid = counts > 0
+    out.real[valid] = sums_re[valid] / counts[valid]
+    out.imag[valid] = sums_im[valid] / counts[valid]
+    return out
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def coh_avg_vis(obs, dt=0, scan_avg=False, return_type="rec",
+                err_type="predicted", num_samples=int(1e3)):
+    """Coherently average visibilities into fixed time bins or per scan.
+
+    Parameters
+    ----------
+    obs : ehtim.obsdata.Obsdata
+    dt : float
+        Bin width in seconds.  Ignored when ``scan_avg=True``.
+    scan_avg : bool
+        If True, average each scan into a single bin.  Requires ``obs.scans``.
+    return_type : {'rec'}
+        Output format.  Only 'rec' is supported here; the legacy 'df' route
+        is gone.
+    err_type : {'predicted', 'measured'}
+        'predicted' propagates the per-row sigmas via inverse variance.
+        'measured' bootstraps the dispersion of the visibility samples.
+    num_samples : int
+        Bootstrap resample size when ``err_type='measured'``.
+
+    Returns
+    -------
+    np.recarray of dtype DTPOL_STOKES or DTPOL_CIRC matching ``obs.polrep``.
+    """
+    if return_type != "rec":
+        raise ValueError("only return_type='rec' is supported")
+    if err_type not in ("predicted", "measured"):
+        raise ValueError(f"err_type must be 'predicted' or 'measured', got {err_type!r}")
+    if dt <= 0 and not scan_avg:
+        return obs.data
+
+    vis_fields, sig_fields, out_dtype = _polrep_fields(obs.polrep)
+    data = obs.data
+
+    # Bin assignment.
+    if scan_avg:
+        if obs.scans is None or len(obs.scans) == 0:
+            raise ValueError("scan_avg=True but obs has no scan table; call add_scans() first")
+        bin_id = _assign_scan_bin(data["time"], obs.scans)
+        keep = bin_id >= 0
+        data = data[keep]
+        bin_id = bin_id[keep]
+    else:
+        # obs.data['time'] is in hours since mjd start; dt is seconds.
+        bin_id = np.floor(data["time"] * 3600.0 / dt).astype(np.int64)
+
+    if len(data) == 0:
+        return np.empty(0, dtype=out_dtype)
+
+    keys = (data["t1"], data["t2"], data["tau1"], data["tau2"], bin_id)
+    gids, n_groups = _group_ids(*keys)
+    first = _first_index_per_group(gids, n_groups)
+
+    out = np.empty(n_groups, dtype=out_dtype)
+    out["t1"] = data["t1"][first]
+    out["t2"] = data["t2"][first]
+    out["tau1"] = data["tau1"][first]
+    out["tau2"] = data["tau2"][first]
+    out["tint"] = np.bincount(gids, weights=data["tint"], minlength=n_groups)
+    counts = np.bincount(gids, minlength=n_groups).astype(np.float64)
+    out["u"] = np.bincount(gids, weights=data["u"], minlength=n_groups) / counts
+    out["v"] = np.bincount(gids, weights=data["v"], minlength=n_groups) / counts
+
+    # Time of the bin: midpoint for fixed-dt, scan min for scan_avg.
+    if scan_avg:
+        time_min = np.full(n_groups, np.inf)
+        np.minimum.at(time_min, gids, data["time"])
+        out["time"] = time_min
+    else:
+        bin_id_per_group = bin_id[first]
+        out["time"] = (bin_id_per_group * dt + dt / 2.0) / 3600.0
+
+    # Visibilities (complex mean) and sigmas (inverse-variance).
+    for vf in vis_fields:
+        out[vf] = _mean_complex(data[vf], gids, n_groups)
+
+    if err_type == "predicted":
+        for sf in sig_fields:
+            out[sf] = _inverse_variance_sigma(data[sf], gids, n_groups)
+    else:  # 'measured': bootstrap the visibility-amplitude dispersion per group.
+        for vf, sf in zip(vis_fields, sig_fields):
+            sig_out = np.full(n_groups, np.nan)
+            for g in range(n_groups):
+                rows = data[vf][gids == g]
+                rows = rows[np.isfinite(rows)]
+                if len(rows) >= 2:
+                    lo, hi = bootstrap(np.abs(rows), np.mean,
+                                       num_samples=num_samples, wrapping_variable=False)[1]
+                    sig_out[g] = 0.5 * (hi - lo)
+            out[sf] = sig_out
+
+    return out
+
+
+def coh_moving_avg_vis(obs, dt=50, return_type="rec"):
+    """Coherent moving-window average over time, per baseline.
+
+    For each row at time ``t``, average over rows in the same baseline whose
+    times fall in ``[t - dt/2, t + dt/2]``.  Output preserves the input rows
+    one-for-one; only the visibility values are replaced.
+
+    Parameters
+    ----------
+    obs : ehtim.obsdata.Obsdata
+    dt : float
+        Window full-width in seconds.
+    return_type : {'rec'}
+        Output format.  Only 'rec' is supported.
+
+    Returns
+    -------
+    np.recarray of dtype DTPOL_STOKES or DTPOL_CIRC matching ``obs.polrep``.
+    """
+    if return_type != "rec":
+        raise ValueError("only return_type='rec' is supported")
+    if dt <= 0:
+        raise ValueError("dt must be positive")
+
+    vis_fields, sig_fields, out_dtype = _polrep_fields(obs.polrep)
+    data = obs.data
+    n = len(data)
+    if n == 0:
+        return np.empty(0, dtype=out_dtype)
+
+    half = dt / 2.0 / 3600.0  # half-width in hours
+    out = data.copy()
+
+    # Group rows by baseline (t1, t2).
+    bl_keys = (data["t1"], data["t2"])
+    bl_ids, _ = _group_ids(*bl_keys)
+
+    for bl in np.unique(bl_ids):
+        rows = np.where(bl_ids == bl)[0]
+        times = data["time"][rows]
+        order = np.argsort(times, kind="stable")
+        sorted_rows = rows[order]
+        sorted_times = times[order]
+
+        # For each row, the inclusive window [t - half, t + half]
+        # corresponds to indices [left, right) in the sorted-time array.
+        left = np.searchsorted(sorted_times, sorted_times - half, side="left")
+        right = np.searchsorted(sorted_times, sorted_times + half, side="right")
+
+        for vf in vis_fields:
+            vals = data[vf][sorted_rows]
+            re_cs = np.concatenate(([0.0], np.cumsum(np.where(np.isfinite(vals), vals.real, 0.0))))
+            im_cs = np.concatenate(([0.0], np.cumsum(np.where(np.isfinite(vals), vals.imag, 0.0))))
+            cnt_cs = np.concatenate(([0.0], np.cumsum(np.isfinite(vals).astype(np.float64))))
+            cnt = cnt_cs[right] - cnt_cs[left]
+            re_mean = np.where(cnt > 0, (re_cs[right] - re_cs[left]) / np.maximum(cnt, 1.0), np.nan)
+            im_mean = np.where(cnt > 0, (im_cs[right] - im_cs[left]) / np.maximum(cnt, 1.0), np.nan)
+            out[vf][sorted_rows] = re_mean + 1j * im_mean
+
+        for sf in sig_fields:
+            vals = data[sf][sorted_rows]
+            finite = np.isfinite(vals) & (vals > 0)
+            inv_var = np.where(finite, 1.0 / np.maximum(vals, 1e-300) ** 2, 0.0)
+            inv_var_cs = np.concatenate(([0.0], np.cumsum(inv_var)))
+            sums = inv_var_cs[right] - inv_var_cs[left]
+            out[sf][sorted_rows] = np.where(sums > 0, 1.0 / np.sqrt(np.maximum(sums, 1e-300)), np.nan)
+
+    return out
+
+
+def incoh_avg_vis(obs, dt=0, debias=True, scan_avg=False, return_type="rec",
+                  rec_type="vis", err_type="predicted", num_samples=int(1e3)):
+    """Incoherently (amplitude) average visibilities, with optional Rice debias.
+
+    Parameters
+    ----------
+    obs : ehtim.obsdata.Obsdata
+        Must be polrep='stokes'.
+    dt : float
+        Bin width in seconds.  Ignored when ``scan_avg=True``.
+    debias : bool
+        Apply Rice debiasing to per-bin mean amplitude.
+    scan_avg : bool
+        If True, average each scan into a single bin.
+    return_type : {'rec'}
+        Output format.  Only 'rec' is supported.
+    rec_type : {'vis', 'amp'}
+        Output recarray dtype.  ``'vis'`` returns DTPOL_STOKES (the I,Q,U,V
+        amplitudes packed into the real parts of the vis fields);
+        ``'amp'`` returns DTAMP.
+    err_type : {'predicted', 'measured'}
+        Sigma propagation method.
+    num_samples : int
+        Bootstrap resample size when ``err_type='measured'``.
+
+    Returns
+    -------
+    np.recarray
+    """
+    if return_type != "rec":
+        raise ValueError("only return_type='rec' is supported")
+    if err_type not in ("predicted", "measured"):
+        raise ValueError(f"err_type must be 'predicted' or 'measured', got {err_type!r}")
+    if rec_type not in ("vis", "amp"):
+        raise ValueError(f"rec_type must be 'vis' or 'amp', got {rec_type!r}")
+    if obs.polrep != "stokes":
+        raise ValueError("incoh_avg_vis requires polrep='stokes'")
+    if dt <= 0 and not scan_avg:
+        return obs.data
+
+    data = obs.data
+
+    if scan_avg:
+        if obs.scans is None or len(obs.scans) == 0:
+            raise ValueError("scan_avg=True but obs has no scan table; call add_scans() first")
+        bin_id = _assign_scan_bin(data["time"], obs.scans)
+        keep = bin_id >= 0
+        data = data[keep]
+        bin_id = bin_id[keep]
+    else:
+        bin_id = np.floor(data["time"] * 3600.0 / dt).astype(np.int64)
+
+    if rec_type == "vis":
+        out_dtype = ehc.DTPOL_STOKES
+        amp_fields = ("vis", "qvis", "uvis", "vvis")
+        sig_fields = ("sigma", "qsigma", "usigma", "vsigma")
+    else:
+        out_dtype = ehc.DTAMP
+        amp_fields = ("vis",)
+        sig_fields = ("sigma",)
+
+    if len(data) == 0:
+        return np.empty(0, dtype=out_dtype)
+
+    keys = (data["t1"], data["t2"], data["tau1"], data["tau2"], bin_id)
+    gids, n_groups = _group_ids(*keys)
+    first = _first_index_per_group(gids, n_groups)
+
+    out = np.empty(n_groups, dtype=out_dtype)
+    out["t1"] = data["t1"][first]
+    out["t2"] = data["t2"][first]
+    out["tint"] = np.bincount(gids, weights=data["tint"], minlength=n_groups)
+    counts = np.bincount(gids, minlength=n_groups).astype(np.float64)
+    out["u"] = np.bincount(gids, weights=data["u"], minlength=n_groups) / counts
+    out["v"] = np.bincount(gids, weights=data["v"], minlength=n_groups) / counts
+
+    if scan_avg:
+        time_min = np.full(n_groups, np.inf)
+        np.minimum.at(time_min, gids, data["time"])
+        out["time"] = time_min
+    else:
+        bin_id_per_group = bin_id[first]
+        out["time"] = (bin_id_per_group * dt + dt / 2.0) / 3600.0
+
+    if "tau1" in out.dtype.names:
+        out["tau1"] = data["tau1"][first]
+        out["tau2"] = data["tau2"][first]
+
+    # Per-group amplitude + sigma via mean_incoh_avg / bootstrap.
+    for vf, sf in zip(amp_fields, sig_fields):
+        amp_out = np.full(n_groups, np.nan)
+        sig_out = np.full(n_groups, np.nan)
+        for g in range(n_groups):
+            mask = gids == g
+            pairs = list(zip(np.abs(data[vf][mask]), data[sf][mask]))
+            if not pairs:
+                continue
+            if err_type == "predicted":
+                a, s = mean_incoh_avg(pairs, debias=debias)
+                amp_out[g] = a
+                sig_out[g] = s
+            else:
+                amps = np.abs(np.asarray([y[0] for y in pairs]))
+                amps = amps[np.isfinite(amps)]
+                if len(amps) >= 2:
+                    centre, (lo, hi) = bootstrap(amps, np.mean,
+                                                 num_samples=num_samples,
+                                                 wrapping_variable=False)
+                    amp_out[g] = centre
+                    sig_out[g] = 0.5 * (hi - lo)
+                elif len(amps) == 1:
+                    amp_out[g] = amps[0]
+                    sig_out[g] = data[sf][mask][0]
+        if rec_type == "vis":
+            out[vf] = amp_out  # cast to complex implicitly via dtype
+        else:
+            out["amp"] = amp_out
+        out[sf] = sig_out
+
+    return out
--- a/ehtim/statistics/averaging.py
+++ b/ehtim/statistics/averaging.py
@@ -67,9 +67,9 @@ import ehtim.const_def as ehc
 
 # TODO: the helpers below (and the per-group amplitude reductions in this
 # module) overlap conceptually with stats.deb_amp / stats.inc_sig /
-# stats.mean_incoh_avg / stats.coh_sig. Worth consolidating into a single
-# set of vectorised reducers once the legacy stats.py call sites elsewhere
-# in the codebase are audited.
+# stats.mean_incoh_avg / stats.coh_sig / stats.invvar_coh_sig. Worth
+# consolidating into a single set of vectorised reducers once the legacy
+# stats.py call sites elsewhere in the codebase are audited.
 from ehtim.statistics.stats import bootstrap, mean_incoh_avg
 
 __all__ = ["coh_avg_vis", "coh_moving_avg_vis", "incoh_avg_vis"]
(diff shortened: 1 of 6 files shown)


ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+
+__all__ = ["coh_avg_vis", "coh_moving_avg_vis", "incoh_avg_vis"]
+
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+
+def _polrep_fields(polrep):
```
achael, line 83: rather than create a new helper function, it might be better to use the existing POLDICT_STOKES and POLDICT_CIRC which give the specific field names in terms of generic labels? It might make it easier when we add the mixed pol fields.

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        )
+    raise ValueError(f"unsupported polrep {polrep!r}")
+
+
+def _assign_scan_bin(times_hr, scans):
+    """Assign each row in `times_hr` to a scan index (or -1 if outside)."""
+    out = np.full(len(times_hr), -1, dtype=np.int64)
+    for idx, scan in enumerate(scans):
+        tstart, tstop = scan[0], scan[1]
+        out[(times_hr >= tstart) & (times_hr <= tstop)] = idx
```
achael: the dataframes.py function "get_bins_labels" had some logic to deal with some edge cases where the scan intervals overlapped or overlapped midnight. Is that accounted for anywhere in the new code?

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+            ehc.DTPOL_CIRC,
+        )
+    raise ValueError(f"unsupported polrep {polrep!r}")
+
+
+def _assign_scan_bin(times_hr, scans):
+    """Assign each row in `times_hr` to a scan index (or -1 if outside)."""
+    out = np.full(len(times_hr), -1, dtype=np.int64)
+    for idx, scan in enumerate(scans):
+        tstart, tstop = scan[0], scan[1]
```
achael, line 135: I think the pandas.cut function used in the original code defines the bin edges as inclusive on the right only, so the bins are "(scan_start, scan_end]" This prevents any data points that are exactly on the bin edge from contributing to two bins. 
https://pandas.pydata.org/docs/reference/api/pandas.cut.html

Can you double check, and if this is right I think the fix is to change "times_hr>=tstart" to "times_hr>tstart" in line 51

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        out[(times_hr >= tstart) & (times_hr <= tstop)] = idx
+    return out
+
+
+def _group_ids(*keys):
+    """Map a row's tuple of key values to a contiguous group index.
```
achael, line 141: can you clarify the docstring a little? maybe with an example? I get that this is converting the label tuple (t1,t2) into an integer index but it took me a while to understand what was happening.

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    for g in range(n_groups):
+        mask = gids == g
+        pairs = list(zip(amps_per_row[mask], sigs_per_row[mask]))
+        if not pairs:
+            continue
+        if err_type == "predicted":
+            a, s = mean_incoh_avg(pairs, debias=debias)
+            amp_out[g] = a
+            sig_out[g] = s
+        else:
```
achael, line 366: put in comment that err_type=='predicted' is using reported sigmas and err_type=='measured' is re-estimating them from time variance (and check if that's right)

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+            amps = np.abs(np.asarray([y[0] for y in pairs]))
+            amps = amps[np.isfinite(amps)]
+            if len(amps) >= 2:
+                centre, (lo, hi) = bootstrap(amps, np.mean,
+                                             num_samples=num_samples,
+                                             wrapping_variable=False)
+                amp_out[g] = centre
+                sig_out[g] = 0.5 * (hi - lo)
```
achael, line 374: this feels like an odd way to me to estimate the sig_out, but i see it's what is in the original code. let's put a comment on it, maybe a TODO at the top of the file. I think it can't be stddev because this is incoherent averaging, but it's worth flagging.

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+            elif len(amps) == 1:
+                amp_out[g] = amps[0]
+                sig_out[g] = sigs_per_row[mask][0]
+    return amp_out, sig_out
+
+
+def _mean_complex(vis_per_row, gids, n_groups):
```
achael: see main comment: I think we want an option for inverse-variance weighting the coherent means as well as the sigmas

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    out.imag[valid] = sums_im[valid] / counts[valid]
+    return out
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def coh_avg_vis(obs, dt=0, scan_avg=False, return_type="rec",
```
achael: we want a new kwarg on averaging type; perhaps invvar_avg==True by default. When this kwarg is True then we do inverse variance weighting, otherwise we do the original weighting exactly as in old dataframes.py.

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    """Coherently average visibilities into fixed time bins or per scan.
+
+    Parameters
+    ----------
+    obs : ehtim.obsdata.Obsdata
+    dt : float
+        Bin width in seconds.  Ignored when ``scan_avg=True``.
+    scan_avg : bool
+        If True, average each scan into a single bin.  Requires ``obs.scans``.
+    return_type : {'rec'}
```
achael: do we need to keep return_type if only "rec" is supported?

ehtim/statistics/averaging.py
achael: it might still be useful to define a helper function here, but I would tie the fields to what is in const_def so we have to change things in fewer places, which will be especially helpful when implementing mixedpol

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        data = data[keep]
+        bin_id = bin_id[keep]
+    else:
+        # obs.data['time'] is in hours since mjd start; dt is seconds.
+        bin_id = np.floor(data["time"] * 3600.0 / dt).astype(np.int64)
+
+    if len(data) == 0:
+        return np.empty(0, dtype=out_dtype)
+
+    keys = (data["t1"], data["t2"], data["tau1"], data["tau2"], bin_id)
```
achael: it looks like for some reason the original code is grouping by 'tau1' and 'tau2', but i don't think it should. These are basically never used fields that were supposed to carry around atmospheric opacity information for calibration purposes but which are almost always zero. We might want to remove them from the internal data structure at some point. 

I'm not sure why these would be included in the sort. If there actually was data in these fields that would give what I think is the wrong behavior --> that is, if the opacity at a site was changing within a bit, it would break up the bin into smaller ones. Since 'tau1' and 'tau2' are basically always just zero this probably doesn't occur in practice. 

Can you double check that my understanding of what is going on with tau1 and tau2 is correct? If so, I think we should only sort on 't1','t2', unless there is some issue in matching the original code that i'm not seeing.

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    out["t1"] = data["t1"][first]
+    out["t2"] = data["t2"][first]
+    out["tau1"] = data["tau1"][first]
+    out["tau2"] = data["tau2"][first]
+    out["tint"] = np.bincount(gids, weights=data["tint"], minlength=n_groups)
+    counts = np.bincount(gids, minlength=n_groups).astype(np.float64)
+    out["u"] = np.bincount(gids, weights=data["u"], minlength=n_groups) / counts
+    out["v"] = np.bincount(gids, weights=data["v"], minlength=n_groups) / counts
+
+    # Time of the bin: midpoint for fixed-dt, scan min for scan_avg.
```
achael: i vaguely remember this convention being a point of controversy. Can we add a comment on the top of the file or a TODO item to make this obvious for future readers?

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    first = _first_index_per_group(gids, n_groups)
+
+    out = np.empty(n_groups, dtype=out_dtype)
+    out["t1"] = data["t1"][first]
+    out["t2"] = data["t2"][first]
+    out["tau1"] = data["tau1"][first]
+    out["tau2"] = data["tau2"][first]
+    out["tint"] = np.bincount(gids, weights=data["tint"], minlength=n_groups)
+    counts = np.bincount(gids, minlength=n_groups).astype(np.float64)
+    out["u"] = np.bincount(gids, weights=data["u"], minlength=n_groups) / counts
```
achael, line 514: add comment that (u,v) values are averaged

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        time_min = np.full(n_groups, np.inf)
+        np.minimum.at(time_min, gids, data["time"])
+        out["time"] = time_min
+    else:
+        bin_id_per_group = bin_id[first]
+        out["time"] = (bin_id_per_group * dt + dt / 2.0) / 3600.0
+
+    # Visibilities (complex mean) and sigmas (inverse-variance).
+    for vf in vis_fields:
+        out[vf] = _mean_complex(data[vf], gids, n_groups)
```
achael: want option for full inverse variance weighting vs current direct unweighted implementation

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    else:  # 'measured': bootstrap the visibility-amplitude dispersion per group.
+        for vf, sf in zip(vis_fields, sig_fields):
+            sig_out = np.full(n_groups, np.nan)
+            for g in range(n_groups):
+                rows = data[vf][gids == g]
+                rows = rows[np.isfinite(rows)]
+                if len(rows) >= 2:
+                    lo, hi = bootstrap(np.abs(rows), np.mean,
+                                       num_samples=num_samples, wrapping_variable=False)[1]
+                    sig_out[g] = 0.5 * (hi - lo)
```
achael, line 552: add comment/todo item on this convention for bootstrapped sigma

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        order = np.argsort(times, kind="stable")
+        sorted_rows = rows[order]
+        sorted_times = times[order]
+
+        # For each row, the inclusive window [t - half, t + half]
+        # corresponds to indices [left, right) in the sorted-time array.
+        left = np.searchsorted(sorted_times, sorted_times - half, side="left")
+        right = np.searchsorted(sorted_times, sorted_times + half, side="right")
+
+        for vf in vis_fields:
```
achael: i think we want an option for inverse variance and direct weighting here as well

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    For each row at time ``t``, average over rows in the same baseline whose
+    times fall in ``[t - dt/2, t + dt/2]``.  Output preserves the input rows
+    one-for-one; only the visibility values are replaced.
+
+    Parameters
+    ----------
+    obs : ehtim.obsdata.Obsdata
+    dt : float
+        Window full-width in seconds.
+    return_type : {'rec'}
```
achael: don't need return_type?

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+    ----------
+    obs : ehtim.obsdata.Obsdata
+        Must be polrep='stokes'.
+    dt : float
+        Bin width in seconds.  Ignored when ``scan_avg=True``.
+    debias : bool
+        Apply Rice debiasing to per-bin mean amplitude.
+    scan_avg : bool
+        If True, average each scan into a single bin.
+    return_type : {'rec'}
```
achael: don't need return_type arguement?

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        bin_id = _assign_scan_bin(data["time"], obs.scans)
+        keep = bin_id >= 0
+        data = data[keep]
+        bin_id = bin_id[keep]
+    else:
+        bin_id = np.floor(data["time"] * 3600.0 / dt).astype(np.int64)
+
+    if rec_type == "vis":
+        out_dtype = ehc.DTPOL_STOKES
+        amp_fields = ("vis", "qvis", "uvis", "vvis")
```
achael, line 734: shouldn't this vary based on polrep? is this hard coded in the original code?

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+        sig_fields = ("sigma", "qsigma", "usigma", "vsigma")
+    else:
+        out_dtype = ehc.DTAMP
+        amp_fields = ("vis",)
+        sig_fields = ("sigma",)
+
+    if len(data) == 0:
+        return np.empty(0, dtype=out_dtype)
+
+    keys = (data["t1"], data["t2"], data["tau1"], data["tau2"], bin_id)
```
achael: see above comment on tau1, tau2

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,414 @@
...
+
+    out = np.empty(n_groups, dtype=out_dtype)
+    out["t1"] = data["t1"][first]
+    out["t2"] = data["t2"][first]
+    out["tint"] = np.bincount(gids, weights=data["tint"], minlength=n_groups)
+    counts = np.bincount(gids, minlength=n_groups).astype(np.float64)
+    out["u"] = np.bincount(gids, weights=data["u"], minlength=n_groups) / counts
+    out["v"] = np.bincount(gids, weights=data["v"], minlength=n_groups) / counts
+
+    if scan_avg:
```
achael, line 760: add comment here too on different conventions on midpoint / stat time for dt vs scan table averaging

tests/test_averaging.py
```diff
@@ -0,0 +1,382 @@
...
+    """Averaged visibility values match the legacy pandas implementation."""
+    new = coh_avg_vis(obs_direct, dt=60.0)
+    old = _legacy_ehdf.coh_avg_vis(obs_direct, dt=60.0, return_type='rec')
+    assert len(new) == len(old)
+    # Sort both by (t1, t2, time) for an order-independent comparison.
+    new_key = np.argsort(new, order=("t1", "t2", "time"))
+    old_key = np.argsort(old, order=("t1", "t2", "time"))
+    new_s, old_s = new[new_key], old[old_key]
+    for vf in ("vis", "qvis", "uvis", "vvis"):
+        np.testing.assert_allclose(new_s[vf], old_s[vf], rtol=1e-10, atol=1e-12,
```
achael, line 315: is the reason that this test passes even though the defn of the averaged sigma has changed  that all the sigmas are equal in the test data set?

achael: can we try testing it on a more complicated data set, like the Gaussian EHT2017 one from conftest.py?

tests/test_averaging.py
```diff
@@ -0,0 +1,382 @@
...
+    obs.data["vsigma"] = np.array([0.1, 0.2])
+
+    new = coh_avg_vis(obs, dt=60.0)
+    old = _legacy_ehdf.coh_avg_vis(obs, dt=60.0, return_type='rec')
+
+    expected_new = 1.0 / np.sqrt(1.0 / 0.1**2 + 1.0 / 0.2**2)
+    expected_old = np.sqrt((0.1**2 + 0.2**2) / 4)
+
+    assert new["sigma"][0] == pytest.approx(expected_new, rel=1e-12)
+    assert old["sigma"][0] == pytest.approx(expected_old, rel=1e-12)
```
achael: we want to implement new kwarg arg for type of weighting, then test that codes are identical on old weighting and differ with new one (which can be default going forward)

achael (commented):
This is going to be an enormous improvement, but I think we want to make some more adjustments. 

1. I think we should merge this branch into dev-backend instead of dev. Then we don't have to worry about the legacy closure averaging functions in dataframes being imported for methods like "avg_cphases" that we have dropped. We can remove the dataframes import entirely everywhere except the test_averaging module. 

2. In the new inverse-variance weighted coherent averaging, I think we want to inverse-variance weight the visibilities themselves as well as the sigmas. That is, <V> = (\Sum (V_i / \sigma_i^2)) / (\Sum (\sigma_i^2)). This is what was implemented in channel averaging in #230 

3. We probably want the same inverse variance weighting by default in the incoherent averaging functions, but I am less sure

4. We should have a keyword argument on the averaging strategy in all functions that will allow us to use the old direct averaging behavior. Something like "invvar_avg=True", or "invvar_weighting=True" 

5. I might be wrong about #2 -- maybe even when directly averaging the visibilites we want to average the sigmas by as sqrt{1/(\Sum (\sigma_i^2))}. I'd check this online or with Claude -- but either way, we want the option to reproduce the old strategy exactly, even if it is no longer default.

ehtim/statistics/averaging.py
rohandahale: Done,  `_polrep_fields` now reads `ehc.POLDICT_STOKES `/ `ehc.POLDICT_CIRC` and pulls field names as `d['vis1'..4] / d['sigma1'..4']`. Mixed pol will plug in via a new POLDICT entry rather than editing this module.

rohandahale: The helper now exists only to map (vis/sig fields, dtype) from a polrep key and the field names themselves live in `const_def`.

ehtim/statistics/averaging.py
rohandahale: Not yet, the new implementation doesn't have the legacy `dataframes.get_bins_labels` overlap/midnight handling. I left that out because `obs.scans` is expected pre-sorted and non-overlapping in current code paths. I am happy to port the overlap logic if you have got a real case where it matters.

ehtim/statistics/averaging.py
rohandahale: I fixed it: `_assign_scan_bin` now uses `(tstart, tstop]` (open-left, closed-right) to match pandas.cut. The docstring is updated as well.

ehtim/statistics/averaging.py
rohandahale: I expanded  this with an explicit example:
keys = (np.array(["a","a","b"]), np.array([0,1,0])) → unique tuples ("a",0), ("a",1), ("b",0) → (array([0, 1, 2]), 3) (label order undefined).

ehtim/statistics/averaging.py
rohandahale: Docstring on `_amplitude_per_group` now spells out predicted = per-row sigmas via inc_sig (analytic Rician-SNR estimator), measured = bootstrap dispersion (half the 68% CI width, legacy convention).

ehtim/statistics/averaging.py
rohandahale: The module docstring now has a paragraph:

>  inc_sig is eq. 9.86 Thompson et al., the noise level producing the equivalent SNR penalty after Rice-debiased averaging — not stddev nor inverse variance. 

Also fixed the `coh_sig` fallback formula in `stats.py` to inverse variance (was the same buggy `sqrt(sum(σ²)/N²))`

ehtim/statistics/averaging.py
rohandahale: Done, new `_invvar_mean_complex` helper; `coh_avg_vis` calls it when `invvar_avg=True` (default), falls back to `_mean_complex` when False.

ehtim/statistics/averaging.py
rohandahale: Added on all three public functions with default True. `invvar_avg=False` reproduces the legacy direct mean exactly and that's what the parity tests now use.

ehtim/statistics/averaging.py
rohandahale: Dropped from all three + the three `obsdata.py` callsites updated.

ehtim/statistics/averaging.py
rohandahale: You are right. I confirmed `tau1 / tau2` are basically always zero in practice and grouping on them would fragment bins on opacity drift. Removed from the grouping keys in both `coh_avg_vis` and `incoh_avg_vis`

ehtim/statistics/averaging.py
rohandahale: I added a module docstring now which spells out the per-bin time choice: midpoint of the dt window for fixed-dt, earliest sample in the scan for scan_avg.

ehtim/statistics/averaging.py
rohandahale: Inline comment added at both `coh_avg_vis` and `incoh_avg_vis`. The module docstring also lists it as a convention.

ehtim/statistics/averaging.py
rohandahale: Same `invvar_avg` kwarg covers this

ehtim/statistics/averaging.py
rohandahale: Inline comment added at the bootstrap branch in `coh_avg_vis` and the module docstring lists it as a convention.

ehtim/statistics/averaging.py
rohandahale: Added the `invvar_avg` kwarg to `coh_moving_avg_vis` as well

ehtim/statistics/averaging.py
rohandahale: Dropped

ehtim/statistics/averaging.py
rohandahale: Dropped

ehtim/statistics/averaging.py
rohandahale: The function already requires polrep='stokes' (raises otherwise), so the hardcoded Stokes I/Q/U/V field list is intentional. Added an inline comment to make that explicit. Generalizing to other polreps would mean lifting the Rice-debias path through `mean_incoh_avg` which is currently Stokes only.

ehtim/statistics/averaging.py
rohandahale: Same as above, dropped from grouping keys

ehtim/statistics/averaging.py
rohandahale: Inline comment added, matches the `dt` vs `scan_avg` convention used by `coh_avg_vis`.

tests/test_averaging.py
rohandahale: You were right. Switched the existing parity test to `invvar_avg=False` (matches legacy regardless of sigma distribution).

I added `test_parity_coh_avg_vis_varied_sigma`: 4 samples in one bin with sigmas [0.1, 0.4, 0.1, 0.4] and increasing vis values. `invvar_avg=False` matches legacy; `invvar_avg=True` returns the analytic invvar mean (1.1) and differs from the direct mean (1.15) by ~0.05. I built it as a controlled fixture rather than `obs_direct` because obs_direct has tadv=600s → one row per baseline per bin, so nothing to average.

rohandahale:
@achael thanks for the thorough review! I worked through all of it. Rebased onto `dev-backend` and switched the PR target to `dev-backend` instead of `dev`.

New commits:

1. Latent bug fixes in `stats.py`: `coh_sig` had the same bad sigma formula as the original `coh_avg_vis`, plus two undefined-name bugs in `dicts_TV_report` / `compare_TV`.
2. Ruff cleanup in `stats.py` + `dataframes.py`
3. `invvar_avg` kwarg, POLDICT, bin edges, drop `tau1` / `tau2` from grouping, drop `return_type`, module docstring with all the conventions spelled out.
4. Gate the predicted-sigma branch behind `invvar_avg` too, so   `invvar_avg=False` reproduces the legacy `sqrt(sum(sig_i^2))/N` formula exactly (not just the vis mean). Parity test rewritten.

`invvar_avg=True` is now the default on all three functions. Made it default on `incoh_avg_vis` too. `invvar_avg=False` reproduces the legacy direct mean exactly, which is what the parity tests now use.

The new `test_parity_coh_avg_vis_varied_sigma` directly answers your question: 4 samples in one bin with sigmas `[0.1, 0.4, 0.1, 0.4]` and increasing vis values, `invvar_avg=False` matches legacy bit-for-bit, `invvar_avg=True` gives the analytic weighted mean (1.1) and differs from the direct mean (1.15) by ~0.05.

Let me know if you want a different default or have more comments.

tests/test_averaging.py
rohandahale: Done. The predicted-sigma branch in `coh_avg_vis` and `coh_moving_avg_vis` now goes through `_inverse_variance_sigma` when `invvar_avg=True` (default) and through a new `_legacy_sigma` helper (`sqrt(sum(sig_i^2)) / N`) when `invvar_avg=False`. So `invvar_avg=False` reproduces the legacy formula
bit-for-bit and `invvar_avg=True` keeps the inverse-variance one.

This test is now rewritten as `test_parity_coh_avg_vis_sigma_invvar_vs_legacy` in the parity-vs-divergence pattern: with sigmas `[0.1, 0.2]`, `invvar_avg=False` matches `_legacy_ehdf.coh_avg_vis` exactly on all four pol sigmas; `invvar_avg=True` returns the analytic `1/sqrt(125) ≈ 0.0894`

ehtim/statistics/averaging.py
achael: it looks like the get_bins_labels function in the original dataframes.py actually expanded the intervals in the obs.scans table by a small delta (dt=0.00001 hr) by default, maybe to account for any issues on the edges. I don't think we should replicate this behavior, but maybe it is worth adding a comment in case this ever pops up as an issue. I think the correct behavior is always to sort this out in the scans table, but i'm vary wary about scan averaging as it has been the source of past issues. 

Alternatively, we could add a dt offset as in get_bins_labels but make it zero by default.

ehtim/statistics/averaging.py
achael: great -- let's rename these to "_inverse_variance_mean_complex" and "_legacy_mean_complex" for parallelism

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,565 @@
...
+    sq = np.where(finite, sig_per_row ** 2, 0.0)
+    sums = np.bincount(gids, weights=sq, minlength=n_groups)
+    counts = np.bincount(gids, weights=finite.astype(np.float64), minlength=n_groups)
+    out = np.full(n_groups, np.nan)
+    valid = counts > 0
+    out[valid] = np.sqrt(sums[valid]) / counts[valid]
+    return out
+
+
+def _amplitude_per_group(amps_per_row, sigs_per_row, gids, n_groups,
```
achael: rename to "_mean_amplitude_group"? And move to below the coherent averaging helper functions

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,565 @@
...
+  - ``tau1`` / ``tau2`` (per-site opacities) are copied through from the
+    first row in each bin. They are not used as grouping keys: they are
+    basically always zero in practice, and using them would needlessly
+    fragment bins if a site's opacity drifted within a window.
+"""
+
+import numpy as np
+
+import ehtim.const_def as ehc
+from ehtim.statistics.stats import bootstrap, mean_incoh_avg
```
achael, line 73: eventually we probably want to merge some of the helper functions here and the older ones we are pulling from stats.py. I don't think it's worth it for now but let's put a TODO next to the import

ehtim/statistics/averaging.py
achael: also add docstrings entries for the other arguments, since this is a little intricate

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,565 @@
...
+    """
+    amp_out = np.full(n_groups, np.nan)
+    sig_out = np.full(n_groups, np.nan)
+    for g in range(n_groups):
+        mask = gids == g
+        pairs = list(zip(amps_per_row[mask], sigs_per_row[mask]))
+        if not pairs:
+            continue
+        if err_type == "predicted":
+            a, s = mean_incoh_avg(pairs, debias=debias)
```
achael, line 363: if i'm reading this correctly, the mean_coh_avg with invvar_avg=False will still apply inverse variance weighting to estimate the mean sigma but not the amplitude. I think we want invvar_avg==True to apply the inverse variance weighting to both, and invvar_avg==False to use exactly the old method without inverse variance weighting. Does that make sense? 

might be worth emailing Maciek and asking him to look at the changes once we merge

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,565 @@
...
+        d = ehc.POLDICT_CIRC
+        out_dtype = ehc.DTPOL_CIRC
+    else:
+        raise ValueError(f"unsupported polrep {polrep!r}")
+    vis = tuple(d[f"vis{i}"] for i in range(1, 5))
+    sig = tuple(d[f"sigma{i}"] for i in range(1, 5))
+    return vis, sig, out_dtype
+
+
+def _assign_scan_bin(times_hr, scans):
```
achael, line 103: let's add a comment or TODO at the top here that we are not copying over the logic on dealing with overlapping scans or scans that overlap midnight without a +24 hr adjustment from get_bins_labels in dataframes.py. I'm not sure this will be a problem on any real data but it was probably in the original function for a reason, so we should at least note its absence here

ehtim/statistics/averaging.py
```diff
@@ -0,0 +1,565 @@
...
+        order = np.argsort(times, kind="stable")
+        sorted_rows = rows[order]
+        sorted_times = times[order]
+
+        # For each row, the inclusive window [t - half, t + half]
+        # corresponds to indices [left, right) in the sorted-time array.
+        left = np.searchsorted(sorted_times, sorted_times - half, side="left")
+        right = np.searchsorted(sorted_times, sorted_times + half, side="right")
+
+        for vf, sf in zip(vis_fields, sig_fields):
```
achael, line 624: is there a reason that the coh_moving_avg_vis doesn't use the helper functions (_inverse_variance_sigma, _legacy_sigma, etc) but implements them directly? If it's possible to reformat this function to use the same helpers as coh_avg_vis that would be great

ehtim/statistics/stats.py
```diff
@@ -262,15 +257,17 @@ def inc_sig(amp,sig):
...
     if snrA>0:
         sigma0=amp0/snrA
-    else: sigma0=coh_sig(amp,sig)
+    else:
+        sigma0=coh_sig(amp,sig)
     return sigma0
 
-def coh_sig(amp,sig):
-    amp = np.abs(np.asarray(amp))
+def coh_sig(amp, sig):
```
achael, line 287: we should keep coh_sig as is. I believe this is the correct estimator for the variance of the directly averaged visibilities (while the inverse variance sigma is the correct estimator of the variance for the inverse variance weighted average visibilities). 

For the inverse-variance weighted path, we should add a new function "invvar_coh_sig." 

We should probably check with Maciek on nuances of invvar weighting with incoherent averaging / debiasing.

achael (commented):
this is looking good -- i have some minor comments on the coherent averaging path. 

On the incoherent averaging path, I'm less certain on how inverse variance weighting works. I think we still want to keep the original behavior in invvar_avg=False -- right now I believe the code is doing a hybrid operation where the sigmas are averaged with inverse variance weighting while the amplitudes themselves are not. 

I believe we always need to match behaviors. That is, the standard RMS average \sqrt{(\sum \sigma^2)/N} is the correct error estimate for the standard mean <V> = \sum V  / N, but 1/\sqrt{\sum{1/sigma^2}} is the correct error estimate for the inverse variance weighted mean <V> = \sum{V / sigma^2}/\sum{1/sigma^2}}. For coherent averaging, the inverse variance weighting is more robust to data points with outliers with low SNRs. For incoherent averaging, I'm less certain about how an inverse variance weighted average interacts with the debiasing, which is why we need to make sure the old path is exactly preserved. 

Once we merge, we should open up an email with Maciek to get him to test out the new code.

rohandahale:
Hi @wielgusm, in this PR we remove the pandas dependency from `statistics/dataframes.py` by moving the three core averaging functions into` statistics/averaging.py`. The goal is for `statistics/averaging.py` to become the default implementation for averaging going forward.

We have added unit tests for the new implementation in `tests/test_averaging.py`, and we would really appreciate your feedback on the changes. In particular, we still have some doubts regarding the incoherent averaging discussed in the PR thread above between Andrew and me.

If you have time, it would also be very helpful if you could run some of your existing scripts against this branch to help validate the implementation and check for any discrepancies.

rohandahale:
@achael Double checked the code and found that both branches were keeping the `inc_sig` Rician-SNR sigma, while only the amplitude switched between legacy `deb_amp` and inverse-variance-weighted mean (and` debias=True` was ignored on `inv-var`). Fixed by coupling the two halves under `invvar_avg` so each is internally consistent:

- ` invvar_avg=False`: legacy pair preserved bit-for-bit, `stats.deb_amp` amplitude +` stats.inc_sig` sigma. New parity test 
`test_parity_incoh_avg_vis_invvar_false_matches_legacy_varied_sigma` pins this on varied-sigma data.
- `invvar_avg=True`: inverse-variance Rice-debiased amplitude + inverse-variance sigma. Amplitude is `sqrt(max(<|V|²>_w − (2 − 1/N)·<σ²>_w, 0)) with w_i = 1/σ_i²`. Since `w_i σ_i² = 1, <σ²>_w = N/Σw_i` and the bias correction simplifies to `(2N − 1)/Σw_i `, generalizing deb_amp's `(2 − 1/Nc)·mean(σ²)` to `inv-var ` weights. In the equal-σ limit this reduces exactly to `deb_amp` (new test `test_incoh_avg_vis_invvar_amp_equal_sigma_matches_deb_amp`).

Module docstring now spells the coupling out, and there's a `test_incoh_avg_vis_invvar_debias_subtracts_bias` pinning that `debias=True` is actually respected on the `inv-var `path.

ehtim/statistics/averaging.py
rohandahale: Done:
- _amplitude_per_group_legacy → _legacy_mean_amplitude_group
- _invvar_amp_sigma_per_group → _inverse_variance_mean_amplitude_group

Moved both below the mean-complex pair so the helper order now reads: sigma pair → mean-complex pair → mean-amplitude pair (each as legacy /  inverse-variance).

Both helpers now have full numpy style docstring.

ehtim/statistics/averaging.py
rohandahale: Done!

ehtim/statistics/averaging.py
rohandahale: You are right. With the latest push the two branches no longer share any sigma code:

```
# invvar_avg=False  → _legacy_mean_amplitude_group → mean_incoh_avg
#                  → amp from stats.deb_amp        (arithmetic mean of amp² − bias)
#                  → sigma from stats.inc_sig      (median(sig) + Rician-SNR formula)
```

```
# invvar_avg=True   → _inverse_variance_mean_amplitude_group
#                  → amp = sqrt(max(<|V|²>_w − (2N−1)/Σwᵢ, 0))
#                  → sigma = 1/√Σwᵢ
```

Neither half of the legacy path is inverse-variance-weighted, so invvar_avg=False is now  identical to `ehtim.statistics.dataframes.incoh_avg_vis` (we call the same` stats.mean_incoh_avg`). New parity test pins this on varied-σ data: `test_parity_incoh_avg_vis_invvar_false_matches_legacy_varied_sigma`.

One caveat: `stats.inc_sig` itself has a low-SNR fallback, when the Rician approximation breaks (snrA ≤ 0) it returns `coh_sig(amp, sig) = 1/√Σ(1/σᵢ²)`. That's an inverse-variance sigma sneaking in by way of a legacy branch inside` stats.py`, not from my code; the False path mirrors legacy behavior exactly there.

ehtim/statistics/averaging.py
rohandahale: I expanded the `_assign_scan_bin` docstring with an explicit "Legacy behaviors deliberately NOT ported" section flagging all three things  `dataframes.get_bins_labels` did that this code doesn't:

1. Edge padding (dt=0.00001 hr cushion).
2. Overlap merging. Overlapping intervals here would cause earlier scans to silently shadow later ones, `obs.scans` is expected pre-sorted and non-overlapping.
3. Midnight wraparound. Intervals with tstop < tstart get an empty mask and drop their rows; multi-day or midnight-crossing tables should be pre-normalised.

Each item names the legacy helper name + the concrete failure mode, so a future hit on any of these has a doc-string-level pointer to where to look.

ehtim/statistics/averaging.py
rohandahale: Good catch! The blocker is that the two functions use different reduction kernels: `coh_avg_vis` scatters rows into disjoint bins via  `np.bincount(gids, ...)`, while `coh_moving_avg_vis` sums overlapping sliding windows via cumsum-differences (cs[right] - cs[left]). Overlapping windows can't be expressed as gids, so the window path can't call the gids-based helpers directly.
 
But the arithmetic was duplicated and now isn't. Extracted segment agnostic primitives that both paths share:

- _inverse_variance_weights(sig) — per-row 1/σ² with finite-masking (the masking pattern that appeared 4 times).
- _combine_mean_inverse_variance, _combine_mean_direct, _combine_sigma_inverse_variance, _combine_sigma_legacy — the four combine formulas, operating on already-summed arrays.
- _window_sums(values, left, right) — the cumsum analogue of bincount for windows.

Now both `coh_avg_vis`'s group helpers and the `coh_moving_avg_vis` loop build per-row terms, sum them per segment (`bincount` for groups, `_window_sums` for windows), and call the same four combiners. The inverse-variance and legacy formulas each live in exactly one place.

Added a `test_coh_moving_avg_vis_full_window_matches_coh_avg_vis` that pins the equivalence: a window wide enough to cover all samples on a baseline reproduces, at every row, the value `coh_avg_vis` gives for the bin containing those samples, for both weighting modes. Plus two tests (...`_invvar_mean_nonuniform_sigma`, ...`_legacy_mean_and_sigma_nonuniform`).

Also documented (in the module + coh_moving_avg_vis docstrings) that the centered window [t−dt/2, t+dt/2] intentionally differs from  the legacy trailing .rolling(dt) + −dt/2 shift, and pinned it with test_coh_moving_avg_vis_centered_window_differs_from_legacy.

ehtim/statistics/stats.py
rohandahale: Done. Reverted `coh_sig` to the original s`qrt(Σσᵢ²)/N` (error of the directly-averaged visibility), agreed it's the right estimator there, and inc_sig's low-SNR fallback now matches upstream again. Added `invvar_coh_sig(amp, sig) = 1/√(Σ 1/σᵢ²)` as the inverse-variance sibling. The earlier "fix" in this branch was a mistake, the two undefined-name fixes in that same commit (dicts_TV_report, compare_TV) are unrelated and stay.

averaging.py's inverse-variance path computes its sigma inline via the vectorized `_combine_sigma_inverse_variance` (same formula as `invvar_coh_sig`); the  consolidation TODO flags unifying these.

ehtim/statistics/averaging.py
rohandahale: The `_assign_scan_bin` docstring now has a "Legacy behaviors deliberately NOT ported" section; item 1 covers exactly this.

It also flags the overlap-merge and midnight-wraparound behaviors from `get_bins_labels` that we likewise don't replicate, since you mentioned scan averaging has been a source of past issues. Skipped the zero-default dt offset to avoid adding a knob nobody currently needs, easy to revisit if a real dataset turns up that needs it.

ehtim/statistics/averaging.py
achael: great, it would be good to hear if @wielgusm has thoughts on if these are necessary

ehtim/statistics/averaging.py
achael: looks good. just want @wielgusm to do a sanity check.

achael (dismissed):
this looks good to me. I'd like to see if we can get @wielgusm to take a quick look before merging.

achael (approved):

PR #253, opened 2026-05-15
Migrate NFFT backend from pynfft to finufft

### LINKED ISSUE #26: installation with NFFT on a cluster as a non-sys admin
**Step 1: NFFT**
`git clone https://github.com/NFFT/nfft `
Following the installation instructions provided, there are a couple things to watch out for. You need to run configure with openMP enabled, and if you're installing on a cluster then you probably want to do a local installation. This should do the trick:
`./configure --enable-all --enable-openmp --prefix=/absolute/path/to/nfft`

**Step 2: pyNFFT**
 If you don't have NFFT installed system wide, then you have to run setup.py like this:
`ttfamily python setup.py build_ext -I /absolute/path/to/nfft/include -L /absolute/path/to/nfftl/lib -R /absolute/path/to/nfft/lib`

However, there is an issue which you have to fix ( [https://github.com/ghisvail/pyNFFT/issues/56](https://github.com/ghisvail/pyNFFT/issues/56) ). To do this open  `pynfft/util.c` and either comment out or remove the line  `#include "nfft3util.h"`
You'll probably also have to use the `--inplace` option while running `python setup.py build_ext` and then the `--user` option while running `python setup.py install` unless you're a system administrator. Note that if the install fails because it can't find the site-packages directory in your installation directory, just add it to your $PYTHONPATH and run the install again. 

**Step 3: eht-imaging**
`git clone https://github.com/achael/eht-imaging
cd eht-imaging
pip install . --user`

NB: if you get an error about an unexpected keyword "overwrite" then you need to modify the ehtim/io/save.py script to update `hdulist.writeto(fname, overwrite=True)` to `hdulist.writeto(fname, clobber=True)`)

rohandahale:
Replaces `pynfft` (unmaintained, last `nfft3` C library release 2016) with `finufft` (actively maintained, universal prebuilt wheels, explicit `eps` accuracy knob). Single-chokepoint rewrite: a new `FINUFFTPlan` class inside `obs_helpers.py` exposes pynfft's stateful `.f_hat` / `.f` / `.trafo()` / `.adjoint()` API on top of two `finufft.Plan` objects (type-2 forward + type-1 adjoint). All ~35 NFFT consumer functions in `imager_utils.py` / `pol_imager_utils.py` stay byte-identical.

`finufft >= 2.2` becomes a required dependency; `pynfft` is dropped entirely.

## Behavior changes worth flagging

1. **New required dep `finufft >= 2.2`.** Prebuilt wheels for macOS x86/arm64, Linux x86_64, Windows on PyPI. Source-build needs C++17.
2. **Default NFFT accuracy bumped** from pynfft's historical `m=2` (~1e-3 relative) to `eps=1e-9`. Tighter than before; bracketing all anticipated arrays (EHT/GMVA, ALMA polarimetry, BHEX, SKA-class dynamic range 1e6+).
3. **`nfft_eps` is a new `Imager(...)` kwarg.** Users can override per call: `eps=1e-12` for SKA-class precision, `eps=1e-6` for fast EHT prototyping. Defaults preserve current behavior at higher precision.
4. **`npad` and `p_rad` on `NFFTInfo`** are now accepted for backward compatibility but unused; finufft picks its own oversampling + kernel width from `eps`.
5. **One test atol relaxed** (`test_parity_nfft_pol[m]`: 1e-15 → 1e-12) because multi-threaded finufft reductions are not bit-deterministic across calls; the e-13 noise floor is below the new bound but above the old one.

## Sign / scale convention pinned

- pynfft `trafo()` ≡ finufft type-2 `isign=-1` (forward).
- pynfft `adjoint()` ≡ finufft type-1 `isign=+1` (adjoint).
- pynfft coords `(-0.5, 0.5]` ↔ finufft `(-π, π]` via `uv_finufft = 2π·uv·psize`.
- isign set explicitly in `FINUFFTPlan.__init__` (not relying on defaults).
- x↔u, y↔v axis convention verified analytically (delta-function FT match to 1e-9 on a rectangular 64×48 image); see `tests/test_nfft_migration.py`.

## Test coverage added

- `tests/test_obs_simulate.py` (2 tests, <1s) — direct-vs-nfft parity for `sample_vis` on an asymmetric off-centre elongated polarized Gaussian. Catches axis swaps, centering, sign, polarization-loop bugs.
- `tests/test_nfft_migration.py` (8 tests, <0.1s) — forward-adjoint transpose identity, real-image dtype upcast (regression), centered-delta vs analytic FT (square + rectangular), `eps` accuracy floor parametrized over 1e-6/1e-9/1e-12, plan reuse across cycles (catches stale-cache bugs).

## Out of scope / follow-ups
- `imager_backend_jax.py` mirror will adopt `jax-finufft`'s functional API (`jax_finufft.nufft1` / `nufft2`) separately.
- `ehtim/imaging/clean.py` → `scripts/clean.py` relocation  — separate PR.
- `clean.py` has its own `NFFTInfo` construction sites that do not yet read `nfft_eps`; can be plumbed when needed.

## Note
- `pyproject.toml` dep change `finufft >= 2.2` is now required (not optional).
- Sign convention (`isign=-1` fwd, `+1` adj) , eyeball on commit `62c48b3`.
- `eps=1e-9` default. If you'd prefer `eps=1e-12` for paranoid safety, trivial to flip.
- `FINUFFTPlan` is stateful by design (mirrors pynfft); JAX mirror should bypass it entirely and call `jax_finufft.nufft1` / `nufft2` directly.

### DIFF
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -33,6 +33,13 @@ import ehtim.const_def as ehc
 import ehtim.image
 import ehtim.imaging.imager_utils as imutils
 import ehtim.imaging.pol_imager_utils as polutils
+from ehtim.const_def import (
+    FFT_INTERP_DEFAULT,
+    FFT_PAD_DEFAULT,
+    GRIDDER_CONV_FUNC_DEFAULT,
+    GRIDDER_P_RAD_DEFAULT,
+    NFFT_EPS_DEFAULT,
+)
 from ehtim.imaging.imager_backend import (
     DATATERMS,
     DATATERMS_POL,
@@ -62,12 +69,6 @@ NHIST = 50   # number of steps to store for hessian approx
 MAXLS = 40   # maximum number of line search steps in BFGS-B
 STOP = 1e-6  # convergence criterion
 EPS = 1e-8
-
-GRIDDER_P_RAD_DEFAULT = 2
-GRIDDER_CONV_FUNC_DEFAULT = 'gaussian'
-FFT_PAD_DEFAULT = 2
-FFT_INTERP_DEFAULT = 3
-NFFT_EPS_DEFAULT = 1e-9
 
 REG_DEFAULT = {'simple': 1}
 DAT_DEFAULT = {'vis': 100}
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -67,6 +67,7 @@ GRIDDER_P_RAD_DEFAULT = 2
 GRIDDER_CONV_FUNC_DEFAULT = 'gaussian'
 FFT_PAD_DEFAULT = 2
 FFT_INTERP_DEFAULT = 3
+NFFT_EPS_DEFAULT = 1e-9
 
 REG_DEFAULT = {'simple': 1}
 DAT_DEFAULT = {'vis': 100}
@@ -196,6 +197,7 @@ class Imager:
         self._fft_conv_func = kwargs.get('fft_conv_func', GRIDDER_CONV_FUNC_DEFAULT)
         self._fft_pad_factor = kwargs.get('fft_pad_factor', FFT_PAD_DEFAULT)
         self._fft_interp_order = kwargs.get('fft_interp_order', FFT_INTERP_DEFAULT)
+        self._nfft_eps = kwargs.get('nfft_eps', NFFT_EPS_DEFAULT)
 
         # multifrequency
         mf = kwargs.get('mf', False)
@@ -776,6 +778,7 @@ class Imager:
             fft_conv_func=self._fft_conv_func,
             fft_gridder_prad=self._fft_gridder_prad,
             fft_interp_order=self._fft_interp_order,
+            nfft_eps=self._nfft_eps,
         )
 
     def make_reg_dict(self, imcur):
--- a/ehtim/observing/obs_helpers.py
+++ b/ehtim/observing/obs_helpers.py
@@ -24,17 +24,13 @@ try:
 except ImportError:
     print("Warning: skyfield not installed: cannot simulate space VLBI")
 
-try:
-    from pynfft.nfft import NFFT
-except ImportError:
-    print("Warning: No NFFT installed!")
-
 import copy
 import itertools as it
 import sys
 import warnings
 
 import astropy.time as at
+import finufft
 import numpy as np
 import scipy.ndimage as nd
 import scipy.spatial.distance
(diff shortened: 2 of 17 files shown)


ehtim/observing/obs_helpers.py
```diff
@@ -1328,30 +1324,67 @@ def uimage(iimage, mimage, chiimage):
 ##################################################################################################
 # FFT & NFFT helper functions
 ##################################################################################################
+class FINUFFTPlan:
+    """Stateful 2D NFFT at fixed nonuniform (u, v) points.
```
achael, line 1328: perhaps worth saying in the docstring that we adopt this format for backwards compatibility with old pynfft implementation

ehtim/observing/obs_helpers.py
```diff
@@ -1328,30 +1324,67 @@ def uimage(iimage, mimage, chiimage):
...
+
+    Set .f_hat (shape (xdim, ydim), complex128) and call .trafo() for the
+    forward transform; the result lands in .f. Set .f (length len(uv),
+    complex128) and call .adjoint(); the result lands in .f_hat.
+    """
+
+    def __init__(self, xdim, ydim, uv_finufft, eps=1e-9):
```
achael: use NFFT_EPS_DEFAULT here?

ehtim/observing/obs_helpers.py
```diff
@@ -1328,30 +1324,67 @@ def uimage(iimage, mimage, chiimage):
...
 class NFFTInfo:
-    def __init__(self, xdim, ydim, psize, pulse, npad, p_rad, uv):
+    """Precomputed NFFT plan + per-point pulse/centering factor.
+
+    eps is the requested relative accuracy of the NFFT. Default 1e-9 is
+    safe for high-dynamic-range imaging (ALMA polarimetry, SKA-scale
+    arrays). Tighten to 1e-12 for ~1e6 dynamic range; relax to 1e-6 for
+    faster low-SNR work where data noise dominates.
+
+    npad and p_rad are accepted for backward compatibility but unused:
```
achael: again say in docstring that npad, p_rad are for backcompat with old pynfft implementation

ehtim/imager.py
```diff
@@ -67,6 +67,7 @@
 GRIDDER_CONV_FUNC_DEFAULT = 'gaussian'
 FFT_PAD_DEFAULT = 2
 FFT_INTERP_DEFAULT = 3
+NFFT_EPS_DEFAULT = 1e-9
```
achael: I think we should define NFFT_EPS_DEFAULT only in const_def. 
Looks like FFT_PAD_DEFAULT and FFT_INTERP_DEFAULT, GRIDDER_FUNC_DEFAULT (and others?) are defined both in const_def and imager.py -- let's fix that and have imager import the values from const_def

tests/test_nfft_migration.py
```diff
@@ -0,0 +1,164 @@
...
+
+        plan.f_hat = img_a
+        plan.trafo()
+        out_a_second = plan.f.copy()
+
+        np.testing.assert_allclose(out_a_second, out_a_first, rtol=1e-12, atol=1e-12)
+
+
+class TestNFFTInfo:
+    """End-to-end tests of NFFTInfo against analytic Fourier transforms."""
```
achael, line 117: nice

achael (commented):
fantastic that this was such a simple fix. should have done it a long time ago :) 

This is passing all the tests we did earlier on gradients/chisqs of  rectangular images, right? That would be my one concern, but I think the tests should cover it. 

Only a few minor comments otherwise

rohandahale:
Thanks @achael ! Yes, rectangular tests are fully covered and are passing. `test_centered_delta_matches_analytic_ft` in `tests/test_nfft_migration.py` is parametrized over both `(64, 64)` and `(64, 48)` shapes. Plus the rectangular sweep from PR #225 still runs through the new NFFTInfo backend.

ehtim/observing/obs_helpers.py
rohandahale: Done!

ehtim/observing/obs_helpers.py
rohandahale: Done! `FINUFFTPlan.__init__` now takes `eps=ehc.NFFT_EPS_DEFAULT`, and the same for `NFFTInfo.__init__` further down. All constants are in `const_def`.

ehtim/observing/obs_helpers.py
rohandahale: Done. Rephrased to: "npad and p_rad are accepted for backwards compatibility with the old pynfft.NFFT implementation but are unused under finufft, which chooses its own oversampling and kernel width
from eps."

ehtim/imager.py
rohandahale: Done! Dropped the five duplicated definitions (GRIDDER_P_RAD_DEFAULT, GRIDDER_CONV_FUNC_DEFAULT, FFT_PAD_DEFAULT, FFT_INTERP_DEFAULT, NFFT_EPS_DEFAULT) from imager.py and replaced with `from ehtim.const_def import (...)` at the top of the file.

 The existing test_defaults_preserved in test_imager_backend.py was updated to import from const_def as well.

achael (approved):
looks great!

PR #256, opened 2026-05-19
Support rectangular images for stochastic_optics

rohandahale:
Adds the first unit tests for `stochastic_optics.py`, supports `xdim != ydim` inputs across the module, fixes two latent compat-only bugs. Part of #212 tasks.

## Rectangular-image fixes

- `sqrtQ_Matrix`: separate `Nx, Ny`; per-axis `FOV`/`dq`/`fftfreq`; per-axis Fourier-shift normalisation.
- `Ensemble_Average_Kernel`: split `uvlist` → `ulist` (xdim) + `vlist` (ydim); FIXME comment removed.
- `MakePhaseScreen`: Fourier-shift exp normalised per-axis  (`s*dx/Nx + t*dy/Ny`).
- `Wrapped_Convolve`: `N = sig.shape[0]` → `Ny, Nx = sig.shape`; pad each axis with its own size.
- `Scatter_Movie`: in each of the 3 input-type branches replace `N = ...xdim` with `Nx, Ny = ...xdim, ...ydim`; reshape `(N, N)` → `(Ny, Nx)`; output Movie reshapes flipped from `(im.xdim, im.ydim)` → `(im.ydim, im.xdim)`.

All changes reduce to the original square formulas when `xdim == ydim`, older square scripts are unaffected.

## Compat-only fixes

- `ScatteringModel.__init__` on modern scipy: `xtol` → `xatol` (deprecated kwarg) + `.x` → `.x[0]` (modern scipy returns a 1-element ndarray). Pre-fix, no model could be initialised on scipy >= 1.5.
- `MakeEpsilonScreenFromList`: `np.complex` → builtin `complex` (NumPy 2.x removed `np.complex`).

## Out of scope

- `Scatter_Movie` line 569 (post-cleanup numbering): `tlist_hr = [... for j in range(N_frames)]` uses the function-signature default `N_frames=None` BEFORE the if-elif-else block sets it. Affects list/static-image input paths; the Movie input branch sets `Unscattered_Movie.times` and bypasses this code. The new `TestScatterMovie` exercises the Movie path which avoids the bug.
- `MakePhaseScreen` amplitude prefactor at line 404 still uses x-only `FOV = psize * xdim * D_obs` (pre-existing). For square it reduces cleanly; for rect, the prefactor evaluates as if the screen were `Ny x Ny` rather than `Ny x Nx`. Whether that's physically right on rect needs a re-derivation from Johnson & Narayan 2016 with Michael's input. The new `TestMakePhaseScreen` only checks shape + finiteness so wouldn't catch a magnitude off-factor.
- `Scatter()` `FOV` variable is misleadingly named on rect (it's `FOV_x` but all uses divide by `Nx` giving `psize * D_obs`, equal to `FOV_y / Ny`).

### DIFF
--- a/ehtim/scattering/stochastic_optics.py
+++ b/ehtim/scattering/stochastic_optics.py
@@ -101,19 +101,19 @@ class ScatteringModel(object):
             def avM_Anisotropy(kzeta):                
                 return np.abs( (kzeta*sps.i0(kzeta)/sps.i1(kzeta) - 1.0)**0.5 - A )
 
-            self.kzeta = minimize(avM_Anisotropy, A**2, method='nelder-mead', options={'xtol': 1e-8, 'disp': False}).x
+            self.kzeta = minimize(avM_Anisotropy, A**2, method='nelder-mead', options={'xatol': 1e-8, 'disp': False}).x[0]
             self.P_phi_prefac = 1.0/(2.0*np.pi*sps.i0(self.kzeta))  
         elif model == 'boxcar':
             def boxcar_Anisotropy(kzeta):                
                 return np.abs( np.sin(np.pi/(1.0 + kzeta))/(np.pi/(1.0 + kzeta)) - (theta_maj_mas_ref**2 - theta_min_mas_ref**2)/(theta_maj_mas_ref**2 + theta_min_mas_ref**2) )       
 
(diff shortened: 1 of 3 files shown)


achael (approved):
again this looks good to me, but it would be useful to ping Michael. We could get his ideas on additional tests as well.

PR #258, opened 2026-05-19
Collapse sX + reg_X regularizer pairs and uniformise scm/scmgrad masking

rohandahale:
Closes roadmap **2.26** + **2.27** on issue #212.

Two follow-ups from PR #238:

- **2.27** — `scm`/`scmgrad` took the embed mask as a positional argument (5th arg) and handled masking internally, breaking the embed-pre / mask-post-slice pattern used by every other spatial regularizer. The wrappers were forced into a special-case path. Fix: rewrite `scm`/`scmgrad` to take a full-grid `imvec`; `reg_cm`/`reggrad_cm` now follow the standard pattern.
- **2.26** — every Stokes-I / pol / MF regularizer came as a `sX` + `reg_X` pair where the `reg_X` was a 1:1 thin sign-flip wrapper over `sX`. Collapse: inline the `sX` body into `reg_X` with sign flipped at return; drop the `sX` function. 

Verified zero direct `sX` imports anywhere in `ehtim/`, `tests/`, `examples/`. Public API is dispatcher-mediated (`compute_regularizer_term` / legacy `regularizer` etc. all string-name-driven). 

`reg_tv2log` / `reggrad_tv2log` were calling `embed(imvec, mask, randomfloor=True)` with the default `clipfloor=0`, so masked positions zero-filled and then `np.log(0) = -inf` propagated through `reg_tv2(np.log(imvec), ...)`, yielding NaN/garbage on any partial mask. `reg_tvlog` already handled this with `clipfloor=epsilon_tv`. Both `tv2log` variants now follow the same pattern (`f53c826`). Latent for as long as `tv2log` has existed; would have silently corrupted results whenever a non-trivial mask was applied to `tv2log` imaging.

## Sign convention after collapse

Each collapsed `reg_X` returns the **penalty value** (positive; imager solver minimises it). For `sX` that returned positive entropy: the inlined math gains a leading `-`. For `sX` that already had `-` in the math: the wrapper `-` and leaf `-` cancel.

## Design decisions

- **Math derivation tags** restored in `reggrad_ptv` and the 5 V-pol gradient functions; explicit `# dS/dI`, `# dS/dm`, `# dS/dphi (phi = 2*chi)`, `# dS/dpsi` markers per branch. The `phi = 2*chi` convention (load-bearing per PR #240) is now annotated at the site that uses it.
- **`# TODO check!!`** flag re-added above the V-pol section — the `gradv * (vfimage / np.tan(psiimage))` chain-rule form is unusual and pre-existing test coverage only exercises generic random pol structure.
- **`# TODO: figure out normalizations for compact/compact2`** flag re-added — the `flux * beam_size**2` (compact) and `flux**2 * beam_size**2` (compact2) normalizations carry a legacy TODO from the previous author.
- **Section headers refreshed** in all three regularizer files to describe what the now-unified `reg_X` functions do (the old "Imager-backend wrappers" header was stale).

## Out of scope (deferred follow-ups)

- **`beam_size or psize` semantics** — pre-existing across all spatial regularizers; not introduced here.
- **`TestComputeRegularizerTerm` rename** — the "legacy parity" tests are now self-comparison because the legacy `regularizer()` shim was rewritten to call `compute_regularizer_term` in PR #238. Renaming/replacing them is a follow-up cleanup.


ehtim/imaging/imager_utils.py
```diff
@@ -1882,671 +1739,425 @@ def fAgrad(imvec, I_ref=1.0, alpha_A=1.0):
...
+    norm = flux if kwargs.get('norm_reg', True) else 1
+    return np.sign(imvec) / norm
 
 
 def reg_l1w(imvec, mask, **kwargs):
-    return -sl1w(imvec, kwargs['nprior'], kwargs['flux'],
-                 norm_reg=kwargs.get('norm_reg', True))
+    priorvec = kwargs['nprior']
+    epsilon = ehc.EP
+    # norm is unity for both norm_reg branches in the legacy sl1w; preserved.
```
achael: for transparency, let's put in an explicit norm=1 as in the old code. I doubt it will ever be upgraded but it wouldn't hurt to keep the placeholder

ehtim/imaging/imager_utils.py
```diff
@@ -1882,671 +1739,425 @@ def fAgrad(imvec, I_ref=1.0, alpha_A=1.0):
...
 
 
-# scm / scmgrad take the embed mask as a positional and handle masking internally,
-# so these wrappers skip the embed-pre / mask-post pattern used by other spatial regs.
 def reg_cm(imvec, mask, **kwargs):
-    return -scm(imvec, kwargs['xdim'], kwargs['ydim'], kwargs['psize'],
-                kwargs['flux'], mask,
-                norm_reg=kwargs.get('norm_reg', True),
-                beam_size=kwargs.get('beam_size'))
+    if np.any(np.invert(mask)):
```
achael, line 1873: the previous implementation did not have an embed step but instead masked the xx,yy arrays. Is the idea that this is better for jax? In standard numpy I would naively think masking xx and yy would be faster than the embed, but I'm not sure.

ehtim/imaging/imager_utils.py
```diff
@@ -1882,671 +1739,425 @@ def fAgrad(imvec, I_ref=1.0, alpha_A=1.0):
...
+    im = imvec.reshape(ny, nx)
+    xx, yy = np.meshgrid(range(nx), range(ny))
+    xxpsize = (xx - (nx-1)/2.0) * psize
+    yypsize = (yy - (ny-1)/2.0) * psize
+    g = 2 * im * (xxpsize**2 + yypsize**2)
+    g = g.reshape(-1) / norm
     return g[mask]
 
 
 def reg_rgauss(imvec, mask, **kwargs):
```
achael, line 2107: add comment that reg_rgauss has no norm_reg option and is always normalized

achael (commented):
looks good to me, only minor comments. I manually checked all of the regularizers and did quick sanity checks on the gradients but am mostly relying on the tests to catch any gradient issues.

ehtim/imaging/imager_utils.py
achael: never mind, i see this is addressed in the PR notes

ehtim/imaging/imager_utils.py
rohandahale: Done!

ehtim/imaging/imager_utils.py
rohandahale: Done!

rohandahale:
@achael I have addressed your minor comments.

achael (approved):
looks good

PR #263, opened 2026-05-20
Add initial test suite for image.py

achael:
## Summary
- Adds `tests/test_image.py`: 159 passing, 2 xfailed
- Method-by-method coverage of every public `Image` method plus the module-level helpers (`make_empty`, `make_square`, `load_image`, `load_txt`, `load_fits`, `avg_imlist`, `get_specim`)
- Mirrors the section layout and session-fixture reuse style of `tests/test_obsdata.py`

## Bugs surfaced and fixed
- `Image.evpa()` circ branch was missing 0.5 factor ([image.py:979](https://github.com/achael/eht-imaging/blob/dev-backend/ehtim/image.py#L979)) — caught by `test_evpa_matches_across_polreps`
- `Image.grad(gradtype='x')` aliased `'abs'` due to a missing `elif` ([image.py:1786-1791](https://github.com/achael/eht-imaging/blob/dev-backend/ehtim/image.py#L1786)) — caught by `test_grad_x_differs_from_abs`

## Issues surfaced for future changes
-  `apply_mask` mutates `self` in place (worked around with `.copy()` in the tests). Rarely used but should be fixed. 

## Test plan
- [ ] `pytest tests/test_image.py` — 159 passed, 2 xfailed locally (~3.5s)
- [ ] CI green on `dev-backend`
- [ ] Rohan review before marking ready for Andrew


rohandahale:
Looks good! Test coverage is good. I verified both bugs the xfails pin are real in `image.py`. These two are genuine `image.py` bugs. I think we should fix them in this PR rather than leaving them as xfails.

One lint fix: `ruff check --fix tests/test_image.py` (single `I001` import sort).

tests/test_image.py
```diff
@@ -0,0 +1,979 @@
...
+    assert f_stokes == pytest.approx(f_circ, rel=1e-12)
+
+
+def test_evpa_stokes_recovers_input_angle(gauss_im_pol):
+    # gauss_im_pol has Q = 0.10*I, U = 0.05*I -> EVPA = 0.5*atan2(0.05, 0.10).
+    expected = 0.5 * np.arctan2(0.05, 0.10)
+    assert gauss_im_pol.evpa() == pytest.approx(expected, rel=1e-10)
+
+
+@pytest.mark.xfail(reason="Bug in Image.evpa circ branch: missing 0.5 factor "
```
rohandahale: I think we should fix it in this PR rather than leaving it as xfail.

tests/test_image.py
```diff
@@ -0,0 +1,979 @@
...
+
+def test_grad_zero_for_uniform_image():
+    arr = np.ones((16, 16))
+    im = eh.image.Image(arr, 1e-10, 17.761, -29.0)
+    out = im.grad()
+    # Sobel on a uniform field is zero everywhere.
+    np.testing.assert_allclose(out.imvec, 0.0, atol=1e-12)
+
+
+@pytest.mark.xfail(reason="Bug in Image.grad (image.py:1786-1791): `if gradtype=='y'` "
```
rohandahale: I think we should fix it in this PR rather than leaving it as xfail.

rohandahale (changes requested):

tests/test_image.py
achael: done!

tests/test_image.py
achael: done!

rohandahale:
@achael looks good to merge!

rohandahale (approved):

PR #260, opened 2026-05-19
Array class migration for mixed-pol (feed_type wiring, TarrView, IO feed_type column)

achael:
Closes Phase 2 of `obsdata_mixedpol_plan_v2.md` (Checkpoint 2). Builds on Phase 1 (#254). Phase 2.5 is gated on this PR.

## What changed

- **Array query methods**: `is_homogeneous_feeds`, `feed_types`, `sefd_for_feed`, `dterm_for_feed`.
- **`add_site` / `add_satellite_tle` / `add_satellite_elements`** extended with `feed_type`, `sefd_p1`/`sefd_p2`, `d_p1`/`d_p2` kwargs. Mixing legacy and generic kwargs for the same field raises `ValueError` (`sefd` vs `sefd_p*`, `dr` vs `d_p1`, `dl` vs `d_p2`, `sefd_p1` without `sefd_p2`, `dr`/`dl` with `feed_type != 'rl'`). Bare `add_site('A', coords)` is bit-identical to today.
- **`TarrView`** wraps `Array.tarr`: column-form access of legacy R/L names raises `KeyError` on non-RL arrays, pointing at `sefd_for_feed` / `dterm_for_feed`. Homogeneous-RL arrays unchanged. Pickle, numpy interop, and recarray-equality patterns all preserved.
- **`Array.obsdata(polrep=...)`** accepts `{'stokes', 'circ', 'lin', 'mixed'}` with explicit feed-type validation. `'lin'` and `'mixed'` raise `NotImplementedError` pointing at Phase 5.
- **`save_array_txt`** emits `# ehtim array format v2` + a 14th `feed_type` column. Numeric layout unchanged. **`load_array_txt`** dispatches on column count (5/13/14) — every existing `arrays/*.txt` file loads bit-identically.
- **`save_obs_txt`** appends `feed_type` as the 14th column on the embedded tarr block; `load_obs_txt`'s Phase 1 `len==15` branch handles it.

## Tests

~78 new `test_phase2_*` tests, including a parametrized bit-identical regression over every `arrays/*.txt` file. Full suite: **1102 passed, 1 skipped** (`SITES.txt` — not an Array file). Baseline pre-PR was 1024 passed.

## Out of scope (deferred)

- `MixedPolConventionWarning` firings — Phase 3c/4
- `obs_simulate.py` mixed-feed noise model — Phase 5
- `model.py` hardcoded RR/RL/LR/LL dispatch — Phase 4
[...]

PR #264, opened 2026-05-20
Add initial test suite for array.py

achael:
## Summary
- Adds `tests/test_array.py`: 43 passing, 1 xfailed
- Method-by-method coverage of every public `Array` method + `load_txt` helper
- Reuses the `eht_array` session fixture from `conftest.py`
- Forward-compatible with PR #260: tests avoid the parts of the API that the Phase 2 mixed-pol Array migration is reshaping (full DTARR field tuple, raw recarray semantics of `arr.tarr`, `'lin'`/`'mixed'` polrep)

## Bug surfaced (xfail test pins current behaviour)
- `Array.add_satellite_tle` body references `tlearr` while the kwarg is named `tlelist`, so calling the function always raises `NameError` ([array.py:256-275](https://github.com/achael/eht-imaging/blob/dev-backend/ehtim/array.py#L256)). PR #260 already renames the kwarg, so the xfail will flip to passing on merge of #260 + forward-merge into `dev-backend`.

## Test plan
- [ ] `pytest tests/test_array.py` — 43 passed, 1 xfailed locally (~2.5s)
- [ ] CI green on `dev-backend`
- [ ] Rohan review before marking ready for Andrew

### DIFF
--- a/tests/test_array.py
+++ b/tests/test_array.py
@@ -1,0 +1,467 @@
+"""Tests for Array methods.
+
+Method-by-method coverage of ehtim/array.py. Sections mirror the class
+itself and reuse the session-scoped ``eht_array`` fixture from conftest.py
+so the I/O cost is amortised.
+
+Designed to survive the Phase-2 mixed-pol changes landing in PR #260
+(branch ``feature/mixpol-phase2-array`` on ``dev-backend-mixpol``):
+
+* Legacy field names ('sefdr', 'sefdl', 'dr', 'dl') will become title
+  aliases on RL-feed arrays after #260, so tests using them on
+  homogeneous-RL arrays continue to pass.
+* DTARR will gain a ``feed_type`` column. Tests avoid asserting the full
+  ``dtype.names`` tuple — they only check that the legacy columns are
+  present.
+* ``Array.tarr`` will return a ``TarrView`` wrapper instead of a raw
+  recarray. Tests interact with ``tarr`` only via behaviour that the
+  wrapper preserves (column access on homogeneous-RL arrays, iteration,
+  ``len``, ``copy``, row indexing, numpy interop).
+* ``add_site`` / ``add_satellite_*`` gain extra kwargs; this suite uses
+  the legacy positional/keyword signatures only.
+* ``Array.obsdata(polrep=...)`` will gain ``'lin'`` and ``'mixed'``
+  options with stricter validation; tests here use only ``'stokes'`` and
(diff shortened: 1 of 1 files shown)


rohandahale:
Looks good! I confirmed the xfail bug on `dev-backend`: `add_satellite_tle(self, tlelist, …)` but the body uses `tlearr` (`array.py:264`, `:272`) → `NameError` on every call, the function has effectively never worked. `test_add_satellite_tle_appends_row` pins it, and #260 renames the kwarg to `tlearr`, so the xfail flips on forward-merge. 

One lint fix: `ruff check --fix tests/test_array.py` (unused `tempfile` import).

rohandahale (approved):

tests/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+        assert col in names
+
+
+def test_init_empty_array_has_empty_tkey():
+    arr = eh.array.Array(np.zeros(0, dtype=ehc.DTARR))
+    assert arr.tkey == {}
+    assert len(arr.tarr) == 0
+
+
+def test_init_nan_coords_without_ephemeris_raises():
```
rohandahale, line 74: This is actually a real bug. `Array.__init__ (array.py:65)` catches except NameError, but ephem[sitename] on a missing key raises KeyError, so the except never fires and the intended 'no ephemeris for site %s !' is never raised. The bare pytest.raises(Exception) passes only because KeyError subclasses Exception, which hides this.

The fix is in ehtim/array.py. The except NameError is dead code, ehem is a parameter (never undefined), so ephem[sitename] on a missing site raises KeyError, which NameError doesn't catch.

`ehtim/array.py:65:`


```
                try:
                    elen = len(ephem[sitename])
                except KeyError:
                    raise Exception('no ephemeris for site %s !' % sitename)
```


And the test then asserts :

```
def test_init_nan_coords_without_ephemeris_raises():
    tarr = np.zeros(1, dtype=ehc.DTARR)
    tarr["site"] = ["SAT_BAD"]
    tarr["x"] = tarr["y"] = tarr["z"] = [np.nan]
    with pytest.raises(Exception, match="no ephemeris"):
        eh.array.Array(tarr)
```

rohandahale (changes requested):

achael:
Fixed the lint error, and the remaining bug will be fixed in the mixedpol version in #260. Let me know when it's ok to merge!

rohandahale:
@achael looks good to merge!

rohandahale (approved):

PR #266, opened 2026-05-20
Fix tlist/bllist collapsing to 2-D when scans share a shape

achael:
## Summary

- `Obsdata.tlist()` and `Obsdata.bllist()` returned `np.array(datalist, dtype=object)`, which silently stacks equal-shape recarrays into a 2-D ndarray and drops the structured dtype. Downstream `tdata[0]['time']` then raises `IndexError`.
- Pre-allocate the 1-D object array and assign element-by-element. Preserves numpy semantics (boolean indexing in `dynamical_imaging.py`, `np.concatenate` callers) while guaranteeing shape `(n,)`.
- Adds 4 regression tests in `tests/test_obsdata.py`: single-scan tlist, multi-scan tlist with uniform shape, bllist on a closed triangle (ALMA/APEX/SPT, all share length 72), and an end-to-end `bispectra()` smoke mirroring the repro from #186.

## Why not return a Python list (as #186 proposed)?

`dynamical_imaging.py:1197` boolean-indexes the result (`tlist[[x == j for x in idx_list]]`), which requires an ndarray. The pre-allocate idiom keeps the array interface intact.

## Test plan

- [x] `pytest tests/test_obsdata.py` — 174 passed
- [x] All 4 new tests fail on the pre-fix code, pass on the fixed code
- [x] Manual repro from #186 (`obs.bispectra(...)` on a single-scan obs) no longer raises

Closes #186.

### DIFF
--- a/ehtim/obsdata.py
+++ b/ehtim/obsdata.py
@@ -613,7 +613,12 @@ class Obsdata:
                     data, lambda x: np.searchsorted(self.scans[:, 0], x['time'])):
                 datalist.append(np.array([obs for obs in group]))
 
-        return np.array(datalist, dtype=object)
+        # np.array(datalist, dtype=object) silently stacks into 2-D when all scans
+        # share a shape, dropping the recarray dtype; pre-allocate to force 1-D.
+        out = np.empty(len(datalist), dtype=object)
+        for i, d in enumerate(datalist):
+            out[i] = d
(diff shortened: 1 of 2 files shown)


rohandahale:
@achael looks good to merge!

rohandahale (approved):

PR #267, opened 2026-05-20
Fix reorder_baselines: distinguish conjugate pairs from data loss

### LINKED ISSUE #140: reorder_baselines loses data
The reorder_baselines method of the Obsdata class assumes that any observations obtained on telescopes (t1, t2) and (t2, t1) are conjugates of each other, and throws away one of the two. However, this is not generally they case, in particular if the observations were obtained at different times and/or frequencies.

I discovered this when importing an OIFITS file, only to find that of the 60 or so wavelengths in the file, only a single one remains after reorder_baselines is run.

achael:
## Summary

- The default `Obsdata.reorder_baselines` path deduped same-time, same-baseline rows silently with no warning, collapsing legitimate multi-channel or duplicate data along with the intended `(A,B)`/`(B,A)` conjugate-pair merge. This is the root cause of #140 and affected every Obsdata construction.
- The opt-in `reorder_baselines_trial_speedups` variant (only reached when callers explicitly passed `trial_speedups=True`; off by default everywhere) had a separate dead-code bug: the dedup mask was built into `dat_unique` and then thrown away, so when invoked it printed a `WARNING` that it had removed duplicates but in fact kept them. The two paths had opposite behavior on the same input.
- Unify into one vectorized implementation. A true conjugate pair (exactly two rows with opposite original orderings and matching `(u, v)` after the swap) is silently merged. Anything else — true duplicate, more than two rows, or mismatched `(u, v)` — fires a `UserWarning`. `trial_speedups` stays on the signature as a no-op for API compatibility.

## Why warn instead of raise?

Most UVFITS files load cleanly. A loud warning surfaces the multi-channel/data-loss case (cf. #140's OIFITS reporter losing 59 of 60 wavelengths) without rejecting files that load today, and positions the codebase for future multi-channel support or a re-implemented `load_oifits`.

## Test plan

- [x] `pytest tests/test_obsdata.py` — 173 passed, 3 new regression tests pass
- [x] Full suite across 4 batches: 1000 passed, 0 failed
- [x] All 3 new tests fail on the pre-fix code
- [x] `test_trial_speedups_matches_default` strengthened to seed a conjugate pair and assert byte-identical output across both flag values

Closes #140.

### DIFF
--- a/ehtim/obsdata.py
+++ b/ehtim/obsdata.py
@@ -338,147 +338,112 @@ class Obsdata:
         return newobs
 
     def reorder_baselines(self, trial_speedups=False):
-        """Reorder baselines to match uvfits convention, based on the telescope array ordering
+        """Reorder baselines to canonical order (tkey[t1] < tkey[t2]) and handle duplicates.
+
+        Within each timestep, any row whose baseline is in reversed order is swapped
+        so it appears as (t1, t2) with tkey[t1] < tkey[t2], conjugating the
+        visibilities and negating (u, v) as needed.
+
+        After the swap, rows that collide on (time, t1, t2) are classified:
+          - **Conjugate pair**: one row was originally (A, B), the other (B, A) at
+            the same time. The two rows represent the same measurement; one is
+            silently dropped.
+          - **True duplicate or multi-channel data**: anything else. A
+            ``UserWarning`` is emitted and only the first row is kept. The
+            single-frequency Obsdata schema cannot represent per-channel data,
+            so loaders that flatten multi-channel files lose information here.
+
+        Args:
+            trial_speedups (bool): kept for API compatibility; the implementation
+                is now always vectorized and this flag has no effect.
         """
-        if trial_speedups:
-            self.reorder_baselines_trial_speedups()
-
-        else: # original code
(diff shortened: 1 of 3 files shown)


ehtim/obsdata.py
```diff
@@ -338,147 +338,112 @@ def switch_polrep(self, polrep_out='stokes', allow_singlepol=True, singlepol_han
...
+                )
+                if is_conj_pair:
+                    pass
+                else:
+                    n_true_dupes_dropped += len(members) - 1
 
-        # save the data
-        self.data = dat
+            if n_true_dupes_dropped > 0:
+                warnings.warn(
```
rohandahale, line 436: Consider softening to e.g. "...at matching (time, t1, t2) that aren't a conjugate pair (duplicate or multi-channel data)..." so the message matches all the cases that trigger it.

rohandahale:
Looks good!

rohandahale (dismissed):

ehtim/obsdata.py
achael: done

rohandahale (approved):

PR #270, opened 2026-05-21
'lin' polrep for Image/Movie/Model + pol_conventions module

achael:
- **New module** `ehtim/observing/pol_conventions.py` — basis transforms, sigma propagation, and Jones-matrix scaffolding. Factored into 8 pair-based primitives (parallel/cross for circ, diag/offdiag for lin) with 4-in/4-out
  wrappers on top.
- **New doc** `docs/polarization_conventions.md` (§1-§12) — convention statement (IAU/HBS), derivations, sign tables, Jones formalism. Each section cites the implementing line range in `pol_conventions.py`.
- **`Image`, `Movie`, `Model`** widen to accept `polrep='lin'` with `pol_prim ∈ {XX, YY}`. All cross-polrep accessors (ivec from circ, rrvec from stokes, xxvec from anything, etc.) compose on demand via
  `pol_conventions`.
- **`Obsdata.switch_polrep`** now calls `pol_conventions.circ_to_stokes` / `stokes_to_circ` (numerically bit-identical to the previous inline math).
- **`Movie` harmonization** — every frame accessor (`iframes`, `qframes`,`rrframes`, etc., all 12 incl. lin) now computes on demand for any source polrep. Pre-Phase-3 they raised on wrong polrep. Setters keep their
  storage-basis guard. - **`Movie.rlvec` / `lrvec` renamed** to `rlframes` / `lrframes` (consistent with `iframes`/`rrframes`/etc.). Old names raise `AttributeError` with a migration message. No in-tree callers used
  these on Movie objects.
- **`Model.observe_same_nonoise`** and the 9 `sample_uv*` dispatch sites gain XX/YY/XY/YX clauses.
- **`const_def.py`** `vis_poldict` / `amp_poldict` / `sig_poldict` extended with XX/YY/XY/YX entries.

## Why

- Provides the sky-model side of mixed-pol observations so linear-feed
  Obsdata (coming in Phase 4) has something to image against.
- Centralizes basis-transform math in one place so future convention
  changes are a single-file edit.
- Movie accessor harmonization gives the class a uniform shape that
  matches `Image`'s long-standing pattern.

## Known issues called out in `tests/test_model.py` header

- `Model.__init__` hardcodes `self.polrep = 'stokes'` regardless of the
  `polrep=` constructor argument, and `Model.switch_polrep` is 
  a no-op. The polrep slot on a Model is always 'stokes' in practice.
- `Model.sample_uv` / `observe_same_nonoise` go through `sample_model_uv`
  (model.py:1467, 1470) which uses `np.sum(generator)`, removed in numpy
  2.0. The end-to-end lin path is marked `xfail(strict=True)`/ 

Both are out of scope here.

## Design points worth discussing

1. **Jones factoring `J = G·(I+D)`** in `pol_conventions.jones_matrix` — per plan Open Question 4, verify against existing `pol_cal*` code when Phase 5 (Caltable rewrite) lands; this PR ships the scaffolding ahead
   of its first consumer.
2. **Per-component sigma propagation** drops cross-component covariance. Documented as a known limitation in `pol_conventions.py` and `docs/polarization_conventions.md` §6. Matches existing inline`Obsdata` behavior bit-for-bit, so no regression — but new code that mixes polreps in noise calcs should be aware.

### DIFF
--- a/docs/polarization_conventions.md
+++ b/docs/polarization_conventions.md
@@ -17,9 +17,15 @@ We work in the sky-tangent frame at the source, with the right-handed
 basis $(\hat e_X, \hat e_Y, \hat k)$ where $\hat k$ points from source
 to observer. The electric field of a monochromatic plane wave is
 
-$$\vec E(t) = \mathrm{Re}\,[E_X \hat e_X + E_Y \hat e_Y]\,e^{-i\omega t},$$
+$$\vec E(t) = \mathrm{Re}\,[E_X \hat e_X + E_Y \hat e_Y]\,e^{+i\omega t},$$
 
-with $E_X, E_Y$ complex analytic amplitudes. The "feed" of an antenna
+with $E_X, E_Y$ complex analytic amplitudes. We use the engineering /
+IEEE positive-frequency convention $e^{+i\omega t}$ standard in radio
+interferometry literature (e.g., TMS Ch. 4). The physics convention
+$e^{-i\omega t}$ would conjugate all complex amplitudes and flip the
+sign of $V$ in the §5 linear formulas; everything else is invariant.
+
+The "feed" of an antenna
 projects $\vec E$ onto a basis vector; the two pure-feed cases of
 interest are **linear** (samples $E_X, E_Y$ directly) and **circular**
 (samples $E_R, E_L$ — defined below).
@@ -31,16 +37,17 @@ Implemented at: convention statement in
 
 ## §2. Linear-circular basis transform — the convention choice (IAU/HBS)
 
-ehtim adopts the IAU / Hamaker-Bregman-Sault / CASA / AIPS convention:
+ehtim adopts the IAU / Hamaker-Bregman-Sault / CASA / AIPS convention
+under the §1 engineering time dependence $e^{+i\omega t}$:
 
-$$R = \frac{X - iY}{\sqrt 2}, \quad L = \frac{X + iY}{\sqrt 2}.$$
+$$R = \frac{X + iY}{\sqrt 2}, \quad L = \frac{X - iY}{\sqrt 2}.$$
 
 In matrix form, the basis-change matrix
 $M_{\mathrm{lin}\to\mathrm{circ}}$ acting on the column vector
 $(E_X, E_Y)^\top$ to produce $(E_R, E_L)^\top$ is
 
 $$M_{\mathrm{lin}\to\mathrm{circ}} = \frac{1}{\sqrt 2}
-  \begin{pmatrix} 1 & -i \\ 1 & +i \end{pmatrix}.$$
+  \begin{pmatrix} 1 & +i \\ 1 & -i \end{pmatrix}.$$
 
 It is unitary, so
 $M_{\mathrm{circ}\to\mathrm{lin}} = M_{\mathrm{lin}\to\mathrm{circ}}^\dagger$.
@@ -49,10 +56,11 @@ The IAU/HBS convention is the standard in radio interferometry
 calibration software (CASA, AIPS, DIFX) and is what AIPS POLTYA/POLTYB
 labels assume when they report an X/Y feed.
 
-The opposite choice ($R = (X+iY)/\sqrt 2$) is sometimes used in optics
-texts and pulsar literature; switching to it requires only changing
-the constant `BASIS_LIN_TO_CIRC` in `pol_conventions.py` (and the
-formulas in §5 below). The sign of $V$ flips under that change.
+Under the physics time convention $e^{-i\omega t}$ the same physical
+basis is written $R = (X - iY)/\sqrt 2$ (complex conjugates of the
+amplitudes). Switching to that convention requires changing the
+constant `BASIS_LIN_TO_CIRC` in `pol_conventions.py` and flipping the
+sign of $V$ in §5; §4 is unaffected.
 
 Implemented at: `ehtim/observing/pol_conventions.py:56-60`.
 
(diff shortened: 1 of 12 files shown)


ehtim/observing/pol_conventions.py
```diff
@@ -0,0 +1,359 @@
...
+# 4-in/4-out functions further down compose two pair calls and are
+# convenience wrappers for callers (like Obsdata.switch_polrep) that
+# want the full four-tuple.
+
+def circ_to_stokes_parallel(rr, ll):
+    """(I, V) from (RR, LL) parallel-hand pair.
+
+    Returns ``(0.5*(RR+LL), 0.5*(RR-LL))``.
+    """
+    _maybe_warn_convention()
```
rohandahale, line 119: The convention warning is now inside the pure transform primitives and in `apply_inverse_jones_to_coherency`. For the eventual JAX mirror these should be side-effect-free (a `warnings.warn` in a traced/vmapped path is a problem). May be consider lifting the `_maybe_warn_convention()` call up to the wrapper layer (`switch_polrep`) so the math primitives stay pure.

docs/polarization_conventions.md
```diff
@@ -0,0 +1,351 @@
...
+
+The **sign of V** depends on the §2 choice: under the opposite basis
+convention, $V \to -i(XY - YX)/2$ and accordingly $XY \to U + iV$,
+$YX \to U - iV$.
+
+Implemented at: pair primitives in
+`ehtim/observing/pol_conventions.py:140-146` (`lin_to_stokes_diag`),
+`:149-158` (`lin_to_stokes_offdiag`), `:160-163`
+(`stokes_to_lin_diag`), `:166-169` (`stokes_to_lin_offdiag`).
+Four-component wrappers at `:200-210` (`lin_to_stokes`) and
```
rohandahale, line 170: 1) Cite TMS (3rd ed) Eq. 4.28 here as "the opposite-sign convention" (it gives `XY = ½(U+jV)`, i.e. `V = −i(XY−YX)`), so the cross-reference is explicit.
2) State the XY correlator labeling explicitly (`XY ≡ ⟨E_x E_y*⟩`). The algebra is all here, but pinning which physical product is "XY" removes the last degree of freedom that sets the V sign.

ehtim/image.py
rohandahale, line 570: `obsdata.py` now routes circ<->stokes through `pol_conventions`, but image.py still has the inline `0.5*(lr+rl)` / `0.5j*(lr-rl)` here. Routing this through `pol_conventions.circ_to_stokes_cross` too would give a single source for the convention.

ehtim/observing/pol_conventions.py
```diff
@@ -0,0 +1,359 @@
...
+
+import numpy as np
+
+from ehtim.warnings import MixedPolConventionWarning
+
+# ---------------------------------------------------------------------------
+# Basis matrices
+# ---------------------------------------------------------------------------
+
+# Maps a feed-basis vector (X, Y) -> (R, L) under the IAU/HBS convention.
```
rohandahale: These are defined but unused, the transforms use the explicit arithmetic forms, and `apply_inverse_jones_to_coherency` inlines `np.linalg.inv` rather than calling `invert_jones`. Fine to keep as documented reference/scaffolding for the deferred single matrix Jones form, but a one line comment noting they are reference only would be great.

ehtim/observing/pol_conventions.py
```diff
@@ -0,0 +1,359 @@
...
+                    [d_p2, 1.0]], dtype=complex)
+    return G @ IpD
+
+
+def invert_jones(J):
+    """Return the inverse of a Jones matrix.
+
+    Broadcasts over leading axes.
+    """
+    return np.linalg.inv(J)
```
rohandahale, line 344: This is defined but unused, the transforms use the explicit arithmetic forms, and `apply_inverse_jones_to_coherency` inlines `np.linalg.inv` rather than calling invert_jones.  A one line comment noting they are reference only would be great.

rohandahale (changes requested):

rohandahale:
@achael Looks good! I just have a couple of comments.

ehtim/observing/pol_conventions.py
achael: I think we should leave this as is for now. 
1. we may not need to jaxify these functions. The imaging/modeling code should use a fixed polrep -- it will need the Jones matrices, but probably not these change of basis functions.
2. we use these transformations in different places across different classes, and if we move the warning we have to be careful to re-implement it every time we call them. 

so for now i'd leave as-is;  if we do need to jaxify these later it will not be hard to move the warning then.

ehtim/observing/pol_conventions.py
achael: done

ehtim/observing/pol_conventions.py
achael: done

ehtim/image.py
achael: done

ehtim/observing/pol_conventions.py
rohandahale: Sounds good!

ehtim/image.py
rohandahale: Is this in a commit that is not pushed?

ehtim/observing/pol_conventions.py
rohandahale: Is this in a commit that is not pushed?

ehtim/image.py
achael: yes, just pushed sorry!

ehtim/observing/pol_conventions.py
achael: just pushed

docs/polarization_conventions.md
achael: After more review, I think we want our linear to circular basis transforms to be R = (X+iY)/sqrt2, L=(X-iY)/sqrt2, which makes our Linear<->Stokes consistent with TMS while keeping the Circular<->Stokes consistent with the historical conventions (used in paper 7 and on). Can you check the updated doc?

achael:
Hi @rohandahale , I've addressed your comments and made 2 other adjustments: 
1. switched the circ<->lin convention. This keeps the lin<->stokes convention in line with TMS while preserving our circ<->lin convention. Can you check the new docs and see if they make sense? 
2. routed all the polrep changes in model.py through pol_conventions.py --> previously they were all inline and could get decoupled from our new central file

docs/polarization_conventions.md
rohandahale: I checked the new docs against TMS (3rd ed. section 4.7). They make sense and the linear <--> stokes is now sign consistent with TMS Eq. 4.32 (XY = U+iV, YX = U−iV, V = −i(XY−YX)/2), which fixes the V sign mismatch I had flagged earlier. Circular parallel matches Table 4.3 (RR=I+V, +V=RCP) and the circular cross-hand keeps ehtim's existing RL=Q+iU.

ehtim/image.py
rohandahale: evec routes the circ branch through `pol_conventions.circ_to_stokes_cross`, so the convention lives in one place. Resolved, thanks!

ehtim/observing/pol_conventions.py
rohandahale: Resolved, thanks!

ehtim/observing/pol_conventions.py
rohandahale: Great, `invert_jones` now has the reference only note (that `apply_inverse_jones_to_coherency` inlines np.linalg.inv). Resolved, thanks!

rohandahale:
Thanks @achael, both edits look great!

rohandahale (approved):

PR #271, opened 2026-05-21
Support the latest stable packages (NumPy 2 / SciPy 1.17 / Python 3.11-3.12)

rohandahale:
`import ehtim` currently fails on the latest scientific stack: SciPy 1.14 removed `scipy.ndimage.filters`, and NumPy 2 removed `np.float`, `np.complex_`, `np.string_` and deprecated `numpy.matlib`. This PR replaces the dead/deprecated APIs, makes pandas optional, raises the Python floor to 3.11 (required by the latest NumPy/SciPy/Astropy/pandas/NetworkX), and removes the remaining SyntaxWarning/DeprecationWarning emitted on import.

Validated in a clean environment with the current stable releases (numpy 2.4.6, scipy 1.17.1, astropy 7.2.0, matplotlib 3.10.9, h5py 3.16.0, networkx 3.6.1, finufft 2.5.1, pandas 3.0.3) on Python 3.12: `import ehtim` is warning-free and the full suite passes (1294 passed, 1 xfailed).

### Changes
- SciPy: `scipy.ndimage.filters`/`.interpolation` to top-level `scipy.ndimage`; REX migrated from the removed `interp2d` to `RegularGridInterpolator`.
- NumPy 2: `np.float` to `float`, `np.complex_` to `np.complex128`, `np.string_` to `np.bytes_`, `np.row_stack` to `np.vstack`, `numpy.matlib.repmat` to `np.tile`, and `np.sum(generator)` to `sum(...)` in the model sampling path.
- pandas is now optional: removed from core dependencies (moved to the `dev` extra) and the eager import in `survey.py` is guarded; also fixed the removed `.ix` accessor.
- Raw strings for LaTeX plot labels (silences the invalid-escape SyntaxWarnings).
- Cleared the pre-existing ruff debt in the touched files (whitespace, import order, compound-statement splits, printf to f-strings, dewildcarded `import *`) so the lint job is green. This surfaced a few latent bugs, now fixed: the numpy-2 `np.sum(generator)` error above (which broke all model imaging), undefined names in `rex.FindProfileSingle`, and an undefined name in the model-parameter dispatch.
- Packaging: `requires-python = ">=3.11,<3.13"`; floors raised (`numpy>=2.0`, `scipy>=1.13`, `astropy>=6.0`, `h5py>=3.0`, `networkx>=3.0`) with no upper caps; CI matrix moved from 3.10 to 3.11.

### Testing
- New `tests/test_rex.py`: the interpolation migration end to end (axis-convention contract, ring-radius recovery, polarized profiles, `findCenter`, rectangular image).
- New `tests/test_model.py` and `tests/test_movie.py`: smoke coverage for the previously-untested model and movie modules (every mring variant, the model dispatch, image-list merge, HDF5 round-trip); this is what caught the `np.sum` numpy-2 bug.
- Full suite green on Python 3.11 and 3.12.

### Out of scope
- The pre-existing `Array.add_satellite_tle` NameError (still xfail in test_array.py) is unrelated and left for a focused fix.
- 63 numerical RuntimeWarnings (divide-by-zero in chisq paths) and 2 UserWarnings are not Syntax/Deprecation and were left untouched.

### DIFF
--- a/ehtim/features/rex.py
+++ b/ehtim/features/rex.py
@@ -104,7 +104,9 @@ class Profiles(object):
 
         self.xs = np.arange(im.xdim)*im.psize/ehc.RADPERUAS
         self.ys = np.arange(im.ydim)*im.psize/ehc.RADPERUAS
-        self.interp = scipy.interpolate.interp2d(self.ys, self.xs, self.imarr, kind=interptype)
+        self.interp = scipy.interpolate.RegularGridInterpolator(
+            (self.ys, self.xs), self.imarr,
+            method=interptype, bounds_error=False, fill_value=None)
 
         self.profiles = np.array(profs)
         self.profilesQ = np.array(profsQ)
@@ -724,13 +726,14 @@ def compute_ring_profile(im, x0, y0, title="",
     ys = np.arange(im.ydim)*im.psize/ehc.RADPERUAS
 
     # TODO: test fiducial images with linear?
-    interp = scipy.interpolate.interp2d(ys, xs, imarr, kind=interptype)
+    interp = scipy.interpolate.RegularGridInterpolator(
+        (ys, xs), imarr, method=interptype, bounds_error=False, fill_value=None)
 
     def ringVals(theta):
         xxs = x0 - rs*np.sin(theta)
         yys = y0 + rs*np.cos(theta)
 
-        vals = [interp(xxs[i], yys[i])[0] for i in np.arange(len(rs))]
+        vals = interp(np.column_stack([yys, xxs]))
         return vals
 
     profs = []
@@ -744,21 +747,23 @@ def compute_ring_profile(im, x0, y0, title="",
     if len(im.qvec) > 0 and len(im.uvec > 0) and pol_profs:
         qarr = im.qvec.reshape(im.ydim, im.xdim)[::-1] * factor  # in brightness temperature K
         uarr = im.uvec.reshape(im.ydim, im.xdim)[::-1] * factor  # in brightness temperature K
-        interpQ = scipy.interpolate.interp2d(ys, xs, qarr, kind=interptype)
-        interpU = scipy.interpolate.interp2d(ys, xs, uarr, kind=interptype)
+        interpQ = scipy.interpolate.RegularGridInterpolator(
+            (ys, xs), qarr, method=interptype, bounds_error=False, fill_value=None)
+        interpU = scipy.interpolate.RegularGridInterpolator(
+            (ys, xs), uarr, method=interptype, bounds_error=False, fill_value=None)
 
         def ringValsQ(theta):
             xxs = x0 - rs*np.sin(theta)
             yys = y0 + rs*np.cos(theta)
 
-            vals = [interpQ(xxs[i], yys[i])[0] for i in np.arange(len(rs))]
+            vals = interpQ(np.column_stack([yys, xxs]))
             return vals
 
         def ringValsU(theta):
             xxs = x0 - rs*np.sin(theta)
             yys = y0 + rs*np.cos(theta)
 
-            vals = [interpU(xxs[i], yys[i])[0] for i in np.arange(len(rs))]
+            vals = interpU(np.column_stack([yys, xxs]))
             return vals
 
         for j in range(nrays):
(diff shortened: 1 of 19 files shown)


ehtim/features/rex.py
```diff
@@ -724,13 +720,14 @@ def compute_ring_profile(im, x0, y0, title="",
...
-    interp = scipy.interpolate.interp2d(ys, xs, imarr, kind=interptype)
+    interp = scipy.interpolate.RegularGridInterpolator(
+        (ys, xs), imarr, method=interptype, bounds_error=False, fill_value=None)
 
     def ringVals(theta):
         xxs = x0 - rs*np.sin(theta)
         yys = y0 + rs*np.cos(theta)
 
-        vals = [interp(xxs[i], yys[i])[0] for i in np.arange(len(rs))]
+        vals = interp(np.column_stack([yys, xxs]))
```
achael, line 730: double check that this is the right order of ys, xs for rectangular images? Looking at the docs it seems right but let's sanity check one more time.

ehtim/image.py
```diff
@@ -23,9 +23,8 @@
 import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
-import numpy.matlib as matlib
 import scipy.interpolate
-import scipy.ndimage.filters as filt
+import scipy.ndimage as filt
```
achael: can we replace the duplicated "import scipy.ndimage as filt" and "from scipy import ndimage as ndi" into one import? filt is only used once (for gaussian_filter).

ehtim/imaging/dynamical_imaging.py
```diff
@@ -31,7 +31,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import requests
-import scipy.ndimage.filters as filt
+import scipy.ndimage as filt
```
achael: for consistency, can we "import scipy.ndimage as ndi" and replace the single call to "filt.gaussian_filter" in dynamical_imaging with "ndi.gaussian_filter"

ehtim/survey.py
```diff
@@ -16,16 +16,23 @@
...
 
+import os
+
 import matplotlib.pyplot as plt
 import numpy as np
-import pandas as pd
-import os
 
-import ehtim as eh
+try:
```
achael, line 25: add a TODO to replace this pandas implementation as well. All it is doing is using pandas for is to save csv files.

tests/test_movie.py
```diff
@@ -0,0 +1,27 @@
+"""Smoke tests for ehtim.movie: image-list merge and HDF5 round-trip."""
```
achael, line 1: good for now. 

The new tests in the mixpol branch for test_movie.py really only touch the polarization transforms, so we could use more comprehensive coverage. Let's make an inventory of what test modules still need to be filled out before we push to dev.

achael (commented):
looks great! only 3 comments
1. sanity check the new interp functions in rex.py take the right order of ys,xs one last time. 
2. clean up scipy.ndimage imports to be consistent (import scipy.ndimage as ndi)
3. we should make a list of what modules still need full test suites developed. Most of the main modules now have test files, but not all of them are filled out to the same extent. Worth tracking somewhere as the test suite gets more complex.

ehtim/features/rex.py
rohandahale: I confirmed that it is correct. ` RegularGridInterpolator((ys, xs), imarr)` with imarr shape `(ydim, xdim)`, so axis 0 <--> ys, axis 1 <--> xs. The `np.column_stack([yys, xxs]) `samples physical `(y, x)`. For the new `test_compute_ring_profile_rectangular`, a swapped (xs, ys) would raise on the shape mismatch there, and it passes.

ehtim/image.py
rohandahale: Done! I did the same in movie.py and dynamical_imaging.py so it's consistent across all three.

ehtim/imaging/dynamical_imaging.py
rohandahale: Done!

ehtim/survey.py
rohandahale: The two to_csv calls now use stdlib csv (identical output, including the leading index column).

And since survey.py was the only thing pulling in paramsurvey (which transitively requires pandas via pandas-appender), I made paramsurvey optional too, guarded the import and moved it to the dev extra. So import ehtim now needs neither `paramsurvey` nor `pandas`. The survey feature still works. It just asks for pip install paramsurvey when you actually use it.

rohandahale:
@achael Thanks! I have addressed all three comments.

tests/test_movie.py
rohandahale: I have added a coverage md file in the docs folder to track the tests coverage.

achael (approved):
looks great, thanks for making the updates

PR #272, opened 2026-05-21
Vectorize diagonalized closure gradients and tidy the imaging backend for JAX

rohandahale:
Prep for the JAX backend port: removes NumPy patterns that are both poor NumPy and JAX friction. No behavior change.

Diagonalized closure terms (cphase_diag, logcamp_diag):
- The inner per-closure Python loop in the gradients is an exact matvec (weight @ tform_mat); replaced.
- For the nfft (production) path, the outer per-time-block loop is also collapsed: the per-block decorrelating transforms are stacked once into a block-diagonal sparse operator in chisqdata, so chisq/grad apply them as a single `T @ samples` / `T.T @ weight`. The block-diagonal layout is vmap/jit-ready for the JAX port.
- direct diag (interleaves the per-block DFT matrices) and plain-FFT `fast` (slated for deprecation) are left as-is.

Other tidy-ups:
- Removed a dead range-check from mcv_grad/vcv_grad: `mfrac_max = 1 - |vfrac| > 1` is unreachable, so the raise never fired; it was only a data-dependent branch in the gradient path.
- pack_imarr: hstack-in-loop replaced by a single concatenate.

Tests: new finite-difference tests pin the diagonalized closure gradients (previously untested) on direct + nfft, plus a cross-check tying the block-diagonal nfft chisq to the direct path. FD diffs are byte-identical before/after the inner-loop change; the block-diagonal form is exact by construction.

Testing: full suite (1286 passed, 1 xfailed); memray within noise (~481 MB peak RSS).

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -355,10 +355,8 @@ def pack_imarr(imarr, which_solve):
     if nsolve != len(which_solve):
         raise Exception("in pack_imarr, imarr has inconsistent shape with which_solve!")
 
-    vec = np.array([])
-    for kk in range(nsolve):
-        if which_solve[kk]!=0:
-            vec = np.hstack((vec,imarr[kk]))
+    blocks = [imarr[kk] for kk in range(nsolve) if which_solve[kk] != 0]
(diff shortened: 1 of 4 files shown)


achael (approved):
looks great, thanks!

PR #273, opened 2026-05-21
Vectorize a few per-visibility loops and deprecate the unused resample_square

rohandahale:
Small perf tidy-ups from a profile-driven audit of the non-imaging-backend code. Each is a behavior-preserving rewrite of a Python loop that scales with data size.

- `obsdata.rescale_zbl`: replace the per-visibility loop (uvdist recomputed each row, 8 per-element writes) with one uvdist + mask and per-field masked scaling.
- `load_obs_uvfits` remove_nan: replace the per-row nan-sigma fill with vectorized `np.where`, preserving the original rr -> ll -> rl/lr fill order.
- `caltable.applycal`: collect per-baseline rows in a list and `hstack` once after the loop instead of growing the table each iteration (was O(N_baseline^2) copies).
- `Image.resample_square`: add a `DeprecationWarning` pointing to `regrid_image`. It is O(N_pix^2) (~30s on a 512->256 resample) and has no callers in the codebase; `regrid_image` is the fast, used path. Flagging for your call on deprecating it - one-line revert if you would rather keep it silent.

This PR deliberately stops here. A profile-driven pass over obsdata/image/caltable showed the other static-audit candidates were either fine (<1s: closures, regrid, blur, observe), legacy/unused (resample_square), or already handled elsewhere (the `avg_coherent` pandas path is PR #252). modeling/scattering/plotting were left alone (parked / heritage / not perf-critical).

Testing: new `tests/test_caltable.py` (unity-gain applycal preserves all visibilities) + a resample_square deprecation-warning test; the existing rescale_zbl test still passes. Full suite 1279 passed, 1 xfailed; memray flat (~481 MB peak RSS).

### DIFF
--- a/ehtim/image.py
+++ b/ehtim/image.py
@@ -1180,8 +1180,16 @@ class Image:
     def resample_square(self, xdim_new, ker_size=5):
         """Exactly resample a square image to new dimensions using the pulse function.
 
-           Deprecated: use :meth:`regrid_image`, which is far faster. This
-           pulse-function resampler is O(N_pix^2) and is not used internally.
+           Deprecated in favor of :meth:`regrid_image`. The two differ:
+           ``resample_square`` convolves the pixel values with the pulse
+           function and resamples the continuous pixel*pulse image
+           representation exactly at the new pixel locations, while
+           ``regrid_image`` interpolates the pixel values directly with
+           scipy.interpolate.RectBivariateSpline. In practice ``regrid_image``
+           is the correct choice and is far faster; use ``resample_square``
+           only if you specifically need the exact continuous pixel*pulse
+           representation at intermediate points. This implementation is
+           O(N_pix^2) and is not used internally.
 
            Args:
                 xdim_new  (int): new pixel dimension
@@ -1191,8 +1199,12 @@ class Image:
                 im_resampled (Image): resampled image
         """
         warnings.warn(
-            "Image.resample_square is deprecated and very slow (O(N_pix^2)); "
-            "use Image.regrid_image instead.",
+            "Image.resample_square is deprecated and slow (O(N_pix^2)). It "
+            "resamples the continuous pixel*pulse representation exactly at the "
+            "new pixel locations; for nearly all uses prefer Image.regrid_image, "
+            "which interpolates the pixel values directly and is far faster. Use "
+            "resample_square only if you need the exact pixel*pulse "
+            "representation at intermediate points.",
             DeprecationWarning, stacklevel=2,
         )
 
--- a/ehtim/image.py
+++ b/ehtim/image.py
@@ -19,6 +19,7 @@
 
 import copy
 import math
+import warnings
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
@@ -1181,6 +1182,9 @@ class Image:
     def resample_square(self, xdim_new, ker_size=5):
         """Exactly resample a square image to new dimensions using the pulse function.
 
+           Deprecated: use :meth:`regrid_image`, which is far faster. This
+           pulse-function resampler is O(N_pix^2) and is not used internally.
+
            Args:
                 xdim_new  (int): new pixel dimension
                 ker_size  (int): kernel size for resampling
@@ -1188,6 +1192,11 @@ class Image:
            Returns:
                 im_resampled (Image): resampled image
         """
+        warnings.warn(
+            "Image.resample_square is deprecated and very slow (O(N_pix^2)); "
+            "use Image.regrid_image instead.",
+            DeprecationWarning, stacklevel=2,
+        )
 
         if self.xdim != self.ydim:
             raise Exception("Image must be square to use Image.resample_square!")
--- a/ehtim/io/load.py
+++ b/ehtim/io/load.py
@@ -1033,7 +1033,7 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
                     allow_singlepol=True, force_singlepol=None,
                     channel=all, IF=all, remove_nan=False,
                     ignore_pzero_date=True,
-                    trial_speedups=False,
+                    trial_speedups=True,
                     invvar_channel_avg=True):
     """Load observation data from a uvfits file.
 
@@ -1049,7 +1049,9 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
            remove_nan (bool): whether or not to remove entries with nan data
            ignore_pzero_date (bool): if True, ignore the offset parameters in DATE field
                                      TODO: what is the correct behavior per AIPS memo 117?
-           trial_speedups (bool): if True, use faster array/telescope handling paths
+           trial_speedups (bool): if True (default), use faster vectorized
+                                  array/telescope handling paths; produces data
+                                  and tarr identical to the legacy paths
            invvar_channel_avg (bool): if True, average IFs/channels with inverse-variance
                                       weighting. If False, use the original simple averaging.
        Returns:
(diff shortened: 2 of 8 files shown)


rohandahale:
1. `ttype='fast'` (plain FFT) gradients are wrong ~30% for every data term

While adding finite-difference tests for the diagonalized closure gradients (#272) I found that **all `ttype='fast'` (plain FFT) gradients are off by ~30% for every data term** (vis, amp, bs, cphase, camp, logcamp, and the `_diag` variants). `direct` and `nfft` are both correct.

Proven two independent ways:
- **FD step-size sweep:** the fractional error is constant (~0.30 median) across step sizes dx = 1e-2 … 1e-8. A real analytic-gradient error stays constant; FD noise would shrink with larger steps. (`nfft` converges linearly to ~1e-6, as expected.)
- **Dot-product adjoint test** on the FFT forward/adjoint: `<y, sampler(g)>` = -39-9j vs `<gridder(y), g>` = -16+14j (|ratio| 1.87). If they were an adjoint pair these would match.

**Root cause:** the forward `sampler` reads the FFT grid with `map_coordinates` (cubic spline, order 3), while the adjoint `gridder` scatters with a gaussian convolution kernel (p_rad 2). They are not adjoint operators, so `chisqgrad_*_fft` is the gradient of a *different* forward model than `chisq_*_fft` computes. `nfft` is exact because finufft type-2/type-1 are adjoint pairs by construction.

Since `fast` is being deprecated I left it alone (the #272 diag FD tests cover `direct`+`nfft` only), but flagging it as a silent multi-year bug. Fix if ever wanted: make the forward sampler use the gridder's conv kernel (or vice-versa) so they form a true transpose pair.

---

2. `self_cal(..., processes=1)` (serial) errors on output assembly

Hit while profiling calibration serially: `self_cal` with `processes=1` runs the per-scan solves fine but then raises `Exception('Data table dtype should be DTPOL_STOKES or DTPOL_CIRC')` assembling the output. The multiprocessing path (`processes=0`/auto) works. Latent serial-path bug.

---

3. `Image.resample_square` deprecation (#273)

In #273 I added a `DeprecationWarning` to `Image.resample_square` pointing to `regrid_image`. It's O(N_pix^2) (~30s on a 512->256 resample) and has **zero callers** in the codebase (only its own tests); `regrid_image` is the fast path everyone uses. Non-breaking — the method still works. Your call on deprecating public API; one-line revert if you'd rather keep it silent.

---

4. `mcv_grad`/`vcv_grad` dead check, possible latent validation gap (#272)

In #272 I removed a dead check from `mcv_grad`/`vcv_grad`: `if np.any(mfrac_max > 1): raise`, where `mfrac_max = 1 - |vfrac|`, so `> 1` means `|vfrac| < 0` — unreachable. The same check sits in the forward `mcv`/`vcv`. Possible it was *meant* to be `< 0` (i.e. `|vfrac| > 1`, a genuinely invalid state)? Pre-existing; flagging in case it's a real validation gap worth fixing in the forward transforms.

---

Calibration perf, for the cal/Jones redesign

Profiling (`scripts/profile_calibration.py`): `applycal` ~5 ms (fine), `self_cal`/`network_cal` are `scipy.minimize`-bound (no vectorizable hotspot). The one real hotspot is `leakage_cal`: `obs_simulate.make_jones`/`apply_jones` get rebuilt once per optimizer iteration via Python comprehensions + ~165k `numpy.linalg` calls. Worth vectorizing the per-baseline Jones construction/application whenever the Jones framework is touched for MixPol.

ehtim/io/load.py
```diff
@@ -1013,6 +1013,20 @@ def load_obs_txt(filename, polrep='stokes'):
...
+    rr_filled = np.where(np.isnan(rr), ll, rr)
+    ll_filled = np.where(np.isnan(ll), rr_filled, ll)
+    rl_filled = np.where(np.isnan(rl), rr_filled, rl)
+    lr_filled = np.where(np.isnan(lr), rr_filled, lr)
+    return rr_filled, ll_filled, rl_filled, lr_filled
+
+
 # TODO can we save new telescope array terms and flags to uvfits and load them?
 # TODO uv coordinates, multiply by IF freqs and not header FREQ?
 def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
```
achael, line 1032: the 'trial_speedups' option is my first vectorization attempt from a few years ago. Is there a test to verify that the 'trial_speedups==True' version of importing a default uvfits agrees with the 'trial_speedups==False' version? If so, we should make trial_speedups==True the default

ehtim/image.py
```diff
@@ -1181,13 +1182,21 @@ def pad(self, fovx, fovy):
     def resample_square(self, xdim_new, ker_size=5):
         """Exactly resample a square image to new dimensions using the pulse function.
 
+           Deprecated: use :meth:`regrid_image`, which is far faster. This
```
achael: regrid_image and resample_square are doing slightly different things: 
1. regrid_image uses scipy.interpolate.RectBivariateSpline directly on the pixel values.  
2. resample_square is convolving the pixel values with the pulse function and then exactly resampling the continuous image representation at the new pixel points. 

I agree that in practice regrid_image is always the correct choice to use, unless you need to know exactly what the continuous pixel*pulse representation is at intermediate points. But I would rewrite the warning to clarify the difference between the functions. 

Probably there is a way to speed up the current implementation of the resample_square function as well, though that is a low-priority item.

ehtim/obsdata.py
```diff
@@ -1535,16 +1535,12 @@ def rescale_zbl(self, totflux, uv_max, debias=True):
...
-                obs.data['uvis'][j] *= totflux / orig_totflux
-                obs.data['vvis'][j] *= totflux / orig_totflux
-                obs.data['sigma'][j] *= totflux / orig_totflux
-                obs.data['qsigma'][j] *= totflux / orig_totflux
-                obs.data['usigma'][j] *= totflux / orig_totflux
-                obs.data['vsigma'][j] *= totflux / orig_totflux
+        scale = totflux / orig_totflux
+        uvdist = np.sqrt(obs.data['u']**2 + obs.data['v']**2)
+        mask = uvdist < uv_max
+        for field in ('vis', 'qvis', 'uvis', 'vvis',
```
achael: this function is I guess so rarely used it didn't get updated to work on arbitrary polreps -- update it to work with the self.poldict logic

achael: and raise a warning at the top if polrep!='stokes' that orig_totalflux may be incorrectly estimated

ehtim/io/load.py
achael: looks like there is already a test -- can you run it on all of the uvfits files in the data directory?

achael (commented):
looks great overall, just a few comments

ehtim/io/load.py
rohandahale: Yes, now `test_load_obs_uvfits_trial_speedups_matches_default` covers this. I extended it to parametrize over every .uvfits in data/ (sample, tutorial2, hops_lo) and both polreps. All 6 cases pass, so I have made `trial_speedups=True` the default.

ehtim/image.py
rohandahale: I have rewritten both the warning and the docstring. The message now says to prefer `regrid_image` for nearly all uses and reach for `resample_square` only when you need the exact pixel pulse values at intermediate points.

ehtim/obsdata.py
rohandahale: Done! The scaling loop now pulls field names from self.poldict. So it works on circ too, previously it would have shown KeyError on a non-Stokes obs. I added a warning at the top when polrep != 'stokes' flagging that `orig_totflux` may be incorrectly estimated, and a circ-polrep test that checks both the scaling and the warning.

achael (approved):

PR #274, opened 2026-05-26
Deprecate ttype='fast' with a DeprecationWarning

rohandahale:
`ttype='fast'` (plain FFT) imaging gradients are inaccurate. Tthe forward sampler (cubic-spline interpolation of the FFT grid) and the adjoint gridder (Gaussian convolution) are not transpose operators, so `chisqgrad_*_fft` is the gradient of a different forward model than `chisq_*_fft` evaluates (found via FD + dot-product adjoint tests while adding the #272 diag-gradient tests; `direct` and `nfft` are exact). Rather than rework the FFT path we're moving off, this deprecates it.

Adds `obs_helpers.warn_fast_ttype_deprecated()`, fired once at the two user entry points where `fast` is selected: `Imager.__init__` (imaging) and `sample_vis` (all forward sampling, incl. `Image.observe` / `Movie`). Points users to `nfft` / `direct`. The code path is kept and still runs.

### DIFF
--- a/ehtim/observing/obs_helpers.py
+++ b/ehtim/observing/obs_helpers.py
@@ -47,18 +47,28 @@ warnings.filterwarnings("ignore", message="divide by zero encountered in double_
 def warn_fast_ttype_deprecated():
     """Emit a DeprecationWarning for the ``ttype='fast'`` (plain FFT) path.
 
-    The plain-FFT transform is deprecated: its imaging gradients are
-    inaccurate because the forward sampler (cubic-spline interpolation of the
-    FFT grid) and the adjoint gridder (Gaussian convolution) are not transpose
-    operators, so the FFT gradient does not match the FFT chi-squared. Use
-    ``ttype='nfft'`` (accurate, recommended) or ``ttype='direct'`` (small
-    problems) instead.
+    Used at forward-sampling sites where the path is deprecated but otherwise
+    correct. For imaging callers, use :func:`warn_fast_ttype_deprecated_imaging`,
+    which additionally notes that the FFT imaging gradients are inaccurate.
     """
     warnings.warn(
         "ttype='fast' (plain FFT) is deprecated and will be removed in a "
-        "future release. Its imaging gradients are inaccurate (the FFT sampler "
-        "and gridder are not adjoint operators). Use ttype='nfft' (recommended) "
-        "or ttype='direct' instead.",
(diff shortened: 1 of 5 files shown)


ehtim/observing/obs_simulate.py
```diff
@@ -227,6 +227,7 @@ def sample_vis(im_org, uv, sgrscat=False, polrep_obs='stokes',
 
     # Get visibilities from straightforward FFT
     if ttype == "fast":
+        obsh.warn_fast_ttype_deprecated()
```
achael, line 230: the fact that the gradients aren't exact isn't relevant for obs_simulate --> I would leave that out of the warning in this case

ehtim/observing/obs_helpers.py
```diff
@@ -44,6 +44,25 @@
...
+
+    The plain-FFT transform is deprecated: its imaging gradients are
+    inaccurate because the forward sampler (cubic-spline interpolation of the
+    FFT grid) and the adjoint gridder (Gaussian convolution) are not transpose
+    operators, so the FFT gradient does not match the FFT chi-squared. Use
+    ``ttype='nfft'`` (accurate, recommended) or ``ttype='direct'`` (small
+    problems) instead.
+    """
+    warnings.warn(
+        "ttype='fast' (plain FFT) is deprecated and will be removed in a "
```
achael, line 55: i think we should shorten this warning and/or have two versions. In obs_simulate, the user doesn't need to know the gradients aren't exact. And even in the "imager" warning I would remove the parenthetical about the sampler and gridder not being adjoint

achael (commented):
looks good, but i would make minor adjustments to the warning.

ehtim/observing/obs_simulate.py
rohandahale: Done! I dropped the gradient note

ehtim/observing/obs_helpers.py
rohandahale: Done, I have split into two helpers now.

rohandahale:
@achael I have addressed both of your comments

achael (approved):
great, this looks good for now. In the mixpol branch we made a warnings module -- i've made a note that once we move primary development over there it would be good to consolidate this warning and other warnings throughout the code in that new module. but for now I don't think this is worth holding up merging this branch.

PR #276, opened 2026-05-28
Calibration test suite + bug fixes

rohandahale:
Test site for calibration, before mixedpol and JAX backend. 67 new tests across `caltable`, `self_cal`, `network_cal`, `leakage_cal`, `polgains_cal`. Five latent bugs surfaced while writing and got fixed in dedicated commits along the way.

### Tests added

| File | Tests |
|---|---|
| `tests/test_caltable.py` | 44 |
| `tests/test_self_cal.py` | 8 |
| `tests/test_network_cal.py` | 7 |
| `tests/test_pol_cal.py` | 3 |
| `tests/test_polgains_cal.py` | 5 |

### Bug fixes

- `Caltable.copy` was a shallow copy aliasing `self.data` / `self.tarr`; now `copy.deepcopy(self)` matching `Obsdata.copy`.
- `make_caltable` used `gains[s*ntele + t]` instead of `s*ntimes + t`, silently transposing the gain mapping for `ntele != ntimes`.
- `scan_avg` did `sites[s]` on a `dict_keys` view, raising `TypeError` on every call.
- `self_cal` and `network_cal` multiprocessing path wrapped `pool.map(...)` in `np.array(..., dtype=object)`, forcing an object outer dtype that corrupted the subsequent `np.concatenate(scans_cal)` and tripped the Obsdata dtype check on the `caltable=False` return.

### DIFF
--- a/ehtim/calibrating/network_cal.py
+++ b/ehtim/calibrating/network_cal.py
@@ -17,24 +17,20 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from __future__ import division
-from __future__ import print_function
 
-from builtins import str
-from builtins import range
-from builtins import object
+import copy
+import time
+from multiprocessing import Pool, cpu_count
 
 import numpy as np
 import scipy.optimize as opt
-import time
-import copy
-from multiprocessing import cpu_count, Pool
 
-import ehtim.obsdata
-import ehtim.parloop as parloop
-from . import cal_helpers as calh
-import ehtim.observing.obs_helpers as obsh
 import ehtim.const_def as ehc
+import ehtim.obsdata
+import ehtim.observing.obs_helpers as obsh
+import ehtim.parloop as parloop
+
+from . import cal_helpers as calh
 
 ZBLCUTOFF = 1.e7
 MAXIT = 5000
@@ -109,14 +105,14 @@ def network_cal(obs, zbl, sites=[], zbl_uvdist_max=ZBLCUTOFF, method="amp", mini
         counter = parloop.Counter(initval=0, maxval=len(scans))
         if processes > len(scans):
             processes = len(scans)
-        print("Using Multiprocessing with %d Processes" % processes)
+        print(f"Using Multiprocessing with {processes} Processes")
         pool = Pool(processes=processes, initializer=init, initargs=(counter,))
     elif processes == 0:
         counter = parloop.Counter(initval=0, maxval=len(scans))
         processes = int(cpu_count())
         if processes > len(scans):
             processes = len(scans)
-        print("Using Multiprocessing with %d Processes" % processes)
+        print(f"Using Multiprocessing with {processes} Processes")
         pool = Pool(processes=processes, initializer=init, initargs=(counter,))
     else:
         print("Not Using Multiprocessing")
@@ -143,7 +139,7 @@ def network_cal(obs, zbl, sites=[], zbl_uvdist_max=ZBLCUTOFF, method="amp", mini
                                             pad_amp=pad_amp, gain_tol=gain_tol, debias=debias)
 
     tstop = time.time()
-    print("\nnetwork_cal time: %f s" % (tstop - tstart))
+    print(f"\nnetwork_cal time: {tstop - tstart:f} s")
 
     if caltable:  # create and return  a caltable
         allsites = obs.tarr['site']
@@ -305,7 +301,7 @@ def network_cal_scan(scan, zbl, sites, clustered_sites, polrep='stokes', pol='I'
     n_gains = len(sites)
     n_clusterbls = len(clusterbls)
     if show_solution:
-        print('%d Gains; %d Clusters' % (n_gains, n_clusterbls))
+        print(f'{n_gains} Gains; {n_clusterbls} Clusters')
 
     gpar_guess = np.ones(n_gains, dtype=np.complex128).view(dtype=np.float64)
     vpar_guess = np.ones(n_clusterbls, dtype=np.complex128)
--- a/ehtim/calibrating/network_cal.py
+++ b/ehtim/calibrating/network_cal.py
@@ -124,12 +124,15 @@ def network_cal(obs, zbl, sites=[], zbl_uvdist_max=ZBLCUTOFF, method="amp", mini
     # loop over scans and calibrate
     tstart = time.time()
     if processes > 0:  # with multiprocessing
-        scans_cal = np.array(pool.map(get_network_scan_cal, [[i, len(scans), scans[i],
-                                                              zbl, sites, cluster_data, obs.polrep, pol,
-                                                              method, pad_amp, gain_tol,
-                                                              caltable, show_solution, debias, msgtype
-                                                             ] for i in range(len(scans))]),
-                                                             dtype=object)
+        # Keep a plain list. Wrapping in np.array(..., dtype=object) used to
+        # force an object outer dtype, which corrupted the subsequent
+        # np.concatenate(scans_cal) for the caltable=False return path and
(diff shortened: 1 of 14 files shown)


ehtim/calibrating/network_cal.py
```diff
@@ -109,27 +105,30 @@ def network_cal(obs, zbl, sites=[], zbl_uvdist_max=ZBLCUTOFF, method="amp", mini
...
     # loop over scans and calibrate
     tstart = time.time()
     if processes > 0:  # with multiprocessing
-        scans_cal = np.array(pool.map(get_network_scan_cal, [[i, len(scans), scans[i],
-                                                              zbl, sites, cluster_data, obs.polrep, pol,
-                                                              method, pad_amp, gain_tol,
-                                                              caltable, show_solution, debias, msgtype
-                                                             ] for i in range(len(scans))]),
-                                                             dtype=object)
+        # Keep a plain list. Wrapping in np.array(..., dtype=object) used to
```
achael, line 123: I'm confused, was this only an issue after recent upgrades to Obsdata.__init__? Because the caltable=False path has definitely worked with multiprocessing in the past. Happy to  make the change, just want to understand when the issue arose.

ehtim/calibrating/self_cal.py
```diff
@@ -151,15 +151,18 @@ def self_cal(obs, im, sites=[], pol='I', apply_singlepol=False, method="both",
...
-        scans_cal = np.array(pool.map(get_selfcal_scan_cal, [[i, len(scans), scans[i],
-                                                              im, V_scans[i], sites,
-                                                              obs.polrep, pol, apply_singlepol,
-                                                              method, minimizer_method,
-                                                              show_solution, pad_amp, gain_tol,
-                                                              debias, caltable, msgtype,
-                                                              use_grad
-                                                              ] for i in range(len(scans))]),
-                                                              dtype=object)
+        # Keep the per-scan results as a plain list. Wrapping in
```
achael, line 154: see above comment

ehtim/caltable.py
```diff
@@ -810,7 +808,7 @@ def make_caltable(obs, gains, sites, times):
     for s in range(0, ntele):
         datatable = []
         for t in range(0, ntimes):
-            gain = gains[s * ntele + t]
+            gain = gains[s * ntimes + t]
```
achael, line 811: don't think this function has ever been used, but good catch

tests/conftest.py
```diff
@@ -240,3 +240,115 @@ def _factory(obs, im, data_term, reg_term=None, pol="I", ttype="direct",
         imcur = transform_imarr(imcur, imgr._config.transforms, imgr._which_solve)
         return imgr, imcur
     return _factory
+
```
achael, line 243: important TODO: currently the mixpol branch has a time dependent d-term field, so we will need to merge these tests into dev-backend-mixpol and update them.

achael (commented):
the bug fixes look good. three comments
1. do the new calibration tests add significant time to the test suite? With the end to end imaging test it already takes about 12 minutes on my machine. Is there a way to define a minimal vs full set of tests so we can leave out tests that take most of the time for most spot checks? 
2. self_cal has a use_grad option that plugs in an explicit gradient. It would be good to test that this gradient is accurate. 
3. pol_cal_new is the updated version of pol_cal that includes explicit gradients for the D-term calibration. I had a test file somewhere that it was producing consistent results with pol_cal, but now I can't find it. I think we should define a test to make sure the gradients are accurate and if so deprecate the old pol_cal in favor of pol_cal_new

tests/conftest.py
rohandahale: Good call! I have added a TODO comment now.

ehtim/calibrating/network_cal.py
rohandahale: It wasn't the Obsdata.__init__ upgrade. That dtype check has been there at obsdata.py:131-132 since commit b89ba6fab in 2018. The trigger is a NumPy 2 semantics change due to PR #271.

Timeline:

1. Oct 2023 (585351c, "fixed dtype=object issue for latest numpy in selfcal"):  You added dtype=object to silence the NumPy 1.x DeprecationWarning about ragged sequence-to-array conversion. Under NumPy 1, np.concatenate still flattened the resulting object array back to the structured DTPOL_* dtype.

2. NumPy 2 changed object-array creation: when inner sequences have uniform length, np.array(list_of_seqs, dtype=object) now gives a 2D object array of shape (n_scans, n_per_scan) instead of a 1D object array of recarrays. On a 2D object array, np.concatenate preserves the outer object dtype. It doesn't drop down to the inner structured dtype. 1D object arrays still flatten correctly.

3. Real obs scans are ragged (different baselines/times per scan) --> 1D object array --> concat still produces DTPOL_* --> has been fine. The new cal test fixtures use a synthetic dense window obs where per-scan rows happen to be uniform --> 2D object array --> concat returns object dtype --> Obsdata.__init__ rejects with "Data table dtype should be DTPOL_STOKES or DTPOL_CIRC".

On numpy 2.4.4:
```
u1 = np.empty(5, dtype=DTPOL_STOKES)
u2 = np.empty(5, dtype=DTPOL_STOKES)
np.concatenate(np.array([u1, u2], dtype=object)).dtype.kind        # 'O'  <-- bug
np.concatenate([u1, u2]).dtype.kind                                # 'V'  <-- works
```

So dropping the np.array(..., dtype=object) wrap (pool.map already returns a list) is robust under both NumPy 1 and 2 and both uniform/ragged scan sizes. The self_cal.py:154 change is the same fix for the same reason.

ehtim/calibrating/self_cal.py
rohandahale: same as above

rohandahale:
@achael Thanks, I addressed all three:

1. Test-suite runtime. I registered a slow marker in pyproject.toml and tagged the optimizer files at module level: test_self_cal.py, test_network_cal.py, test_pol_cal.py, test_polgains_cal.py, test_imager_e2e.py, test_imager_history.py. test_caltable.py stays unmarked (fast, no optimizer). Default pytest tests/ runs the full suite unchanged; `pytest -m "not slow"` skips the 105 tagged tests for dev spot  checks. CI keeps running the full suite.

2. self_cal use_grad gradient. New TestErrfuncGradient class with FD parity against errfunc_grad_full at 3 perturbation points (|g|<1, |g|>1, third sub-unity), each as its own parametrised case.

3. pol_cal_new gradient + deprecation. New TestLeakageCalNewGradient in tests/test_pol_cal.py. The errfunc / errfunc_grad are inner closures capturing FT-of-image + gains + field rotation, so the test spies on pol_cal_new.opt.minimize to grab the closures, then runs FD parity at 3 random Dpar seeds. Gradient verified accurate, so I added the deprecation in this PR. I added a one-line DeprecationWarning at the top of `leakage_cal` pointing to to `leakage_cal_new`.

ehtim/calibrating/network_cal.py
achael: ok, this makes sense. Unclear to me why numpy 2 had to change all of these things :)

achael:
Sounds great. We can probably soon remove the old leakage_cal and fully replace it with leakage_cal_new, but this works for now.

ehtim/calibrating/pol_cal.py
```diff
@@ -64,6 +65,11 @@ def leakage_cal(obs, im=None, sites=[], leakage_tol=.1, pol_fit=['RL', 'LR'], dt
        Returns:
            (Obsdata): the calibrated observation, with computed leakage values added to the obs.tarr
     """
+    warnings.warn(
```
achael, line 68: great

achael (approved):

PR #278, opened 2026-05-28
rex: FindProfile + RingProfile plotting + off-centre findCenter tests

rohandahale:
Fills the partial-coverage gaps on `ehtim/features/rex.py` per `docs/coverage_status.md`. 8 new tests:

- `FindProfile` single-image entry point: returns a `RingProfile` with `meanprof` populated and shape-consistent with `rs`.
- `findCenter` on an off-centre ring: the search actually moves to the right pixel (within one psize), pinning the contract that REX can localise rings outside the geometric centre.
- All six `RingProfile` plot methods (`plot_img`, `plot_unwrapped`, `plot_profs`, `plot_prof_band`, `plot_meanprof`, `plot_meanprof_theta`) — Agg-backend smoke, `save_png=False`.

### DIFF
--- a/tests/test_rex.py
+++ b/tests/test_rex.py
@@ -123,3 +123,78 @@ def test_findcenter_locates_ring_center(ring_image):
     assert np.all(np.isfinite(res))
     assert abs(res[0] - center) < 8.0
     assert abs(res[1] - center) < 8.0
+
+
+# ---------------------------------------------------------------------------
+# Off-centre ring + FindProfile entry point
+# ---------------------------------------------------------------------------
+
+
+# Offset for the off-centre ring test: shift the ring by 8 uas in (+x, -y)
+# so findCenter has to actually search rather than pick the geometric centre.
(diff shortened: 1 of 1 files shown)


achael (commented):
looks good, but findCenter is a very slow function (it's a brute force search). Does this new test add significant time to our test suite? Maybe we can define a minimal test suite that doesn't include optimizer runs for running frequently during development and leave the full suite including the optimizer runs to only be run before pushing PRs, during CI?

rohandahale:
@achael The two new `findCenter` tests are 0.07 s and 0.08 s on the small EHT-sized fixtures (gauss_im and a synthetic 64 pix ring). The  whole `tests/test_rex.py` runs in 0.77 s for 14 tests. The brute force search only gets expensive on much larger images.
 
The broader minimal/full split applies much more to #276 . I'll register a slow marker + tag the genuinely slow tests over there in the same PR where you raised it.

achael:
makes sense, thanks for checking!

achael (approved):

PR #280, opened 2026-05-28
array: plot_dterms + plot_satellite_orbits + satellite-elements tests + add_satellite_tle kwarg fix

rohandahale:
Fills the partial-coverage gaps on `array.py` per `docs/coverage_status.md`. 3 new tests + 1 unxfailed + 1 source fix.

### Bonus fix

`Array.add_satellite_tle` had its kwarg name (`tlelist`) mismatched against the body's `tlearr` references, so every call raised `NameError`. Renamed the kwarg to `tlearr` to match the body (and matches the rename Andrew made on `dev-backend-mixpol` in #260). Docstring already said `tlearr (str)`.

### Tests added

- `plot_dterms` Agg-backend smoke.
- `add_satellite_elements_populates_keplerian_ephem` — verifies the 6-tuple ephem entry.
- `plot_satellite_orbits` Agg-backend smoke (uses `add_satellite_elements` to populate ephem first, then propagates a short LEO orbit).

### Unxfailed

- `test_add_satellite_tle_appends_row` now passes courtesy of the kwarg fix above.

PR #281, opened 2026-05-28
Clean up ruff debt in network_cal.py

rohandahale:
Pre-existing ruff debt in `ehtim/calibrating/network_cal.py` (the touch-debt-laden-file pattern) gets inherited by any PR that touches the file. PR #276 hit this during the cal-cluster work and bundled the cleanup; extracting it here so it lands independently and the next touch is clean.

Pure ruff fixes, no functional change:

- Removed unused `from __future__ import division` / `print_function` (`UP010`).
- Removed unused `from builtins import str`, `range`, `object` (`UP029`).
- 4× `'...' % var` → `f"..."` (`UP031`).
- Import block sort (`I001`).
- Whitespace and trailing-blank-line strips (`W291`, `W293`).

Identical cleanup as the one bundled in PR #276; that commit becomes a no-op on rebase after this lands. The matching cleanup for `array.py` was bundled with the kwarg fix in PR #280 and isn't extracted here because the cleanup uncovers a latent `F821` that needs the kwarg fix to resolve.

### DIFF
--- a/ehtim/calibrating/network_cal.py
+++ b/ehtim/calibrating/network_cal.py
@@ -17,24 +17,20 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from __future__ import division
-from __future__ import print_function
 
-from builtins import str
-from builtins import range
-from builtins import object
+import copy
+import time
+from multiprocessing import Pool, cpu_count
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


rohandahale:
Redundant — the network_cal.py cleanup is already bundled in #276, and array.py cleanup is in #280. Closing.

PR #283, opened 2026-06-01
MixPol: data_conj, closures, and unpack widening for lin/mixed

achael:
Widen Obsdata's data_conj, closure quantities, and unpack to the 'lin' and 'mixed' polreps. 

- data_conj: generic poldict-driven conjugate/swap for circ/lin/mixed; flips the polbasis halves on conjugate rows.
- Closures (bispectra / c_phases / c_amplitudes + diag + tri/quad): make_bispectrum / make_closure_amplitude consolidated onto one per-correlation helper (vis_component) that routes basis synthesis through pol_conventions. Mixed: skip feed-mixing triangles/quads with MixedPolClosureSkipWarning (+ minimal-set caveat); Stokes/generic-slot vtypes raise.
- pol_conventions: new coherency_to_stokes / stokes_to_coherency (+ sigma) and correlation_slot. Under ideal feeds every baseline measures the full coherency, so Stokes is recoverable from any feed pairing, including
  heterogeneous and hybrid feeds.
- unpack: lin via the vis_component consolidation (matches switch_polrep exactly); mixed via row-aligned NaN-fill (physical names -> matching slot or NaN with a per-field MixedPolUnpackNaNWarning; Stokes-derived recovered on
  every row via coherency_to_stokes; generic slots direct). All vis-family dispatch lives in obs_helpers.unpack_vis
- load_uvfits: detect non-R/L POLTYA/POLTYB and raise (full parse is Phase 7)

## transforms in current unpack and closure behavior are somewhat inconsistent: worth more discussion
mixed -> Stokes synthesis is now defined (per-row, ideal-feed) and used by unpack, but raises in closures
mixed -> coherency synthesis only returns native components and nans (unpack) or skips (closures) undefined components rather than performing a feed transform 'circ' polrep will not return 'lin' quantities and vice versa. 'Stokes' polrep will return both.  object-level switch_polrep involving input our output polrep=='mixed' still raises.

## Verification
- Full suite green (test_mixedpol, test_obsdata, test_pol_conventions, test_io, test_imager_e2e): 412 passed, 1 skipped.
- Cross-checked against v1.3 on data/sample.uvfits and the HOPS M87 file: closure tables and unpack outputs are bit-identical for stokes and circ across all legacy fields; only synthesized-vtype bispectrum sigmas differ at
  ~1e-13 (sqrt/square round-trip, machine precision).
- New e2e test: Stokes-I closure imaging is polrep-transparent (circ and lin reproduce the stokes-obs reconstruction).

### DIFF
--- a/ehtim/obsdata.py
+++ b/ehtim/obsdata.py
@@ -1007,122 +1007,15 @@ class Obsdata:
                 tdata = self.tarr[keys]
                 out = sites
                 ty = 'U32'
-            elif self.polrep == 'mixed':
-                # MIXED has no single column formula (slot meaning varies per row
-                # with feed_type), so dispatch per row via polbasis: row-aligned
-                # NaN-fill, Stokes-derived recovered with coherency_to_stokes.
-                out, sig, ty = obsh.unpack_vis_mixed(data, field)
-            elif field in ['vis', 'amp', 'phase', 'snr', 'sigma', 'sigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'vis', self.polrep)
-            elif field in ['qvis', 'qamp', 'qphase', 'qsnr', 'qsigma', 'qsigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'qvis', self.polrep)
-            elif field in ['uvis', 'uamp', 'uphase', 'usnr', 'usigma', 'usigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'uvis', self.polrep)
-            elif field in ['vvis', 'vamp', 'vphase', 'vsnr', 'vsigma', 'vsigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'vvis', self.polrep)
-            elif field in ['pvis', 'pamp', 'pphase', 'psnr', 'psigma', 'psigma_phase']:
-                ty = 'c16'
-                if self.polrep in ('stokes', 'circ'):
-                    out, sig = obsh.vis_component(data, 'rlvis', self.polrep)  # P = RL
-                elif self.polrep == 'lin':
-                    q, qsig = obsh.vis_component(data, 'qvis', 'lin')
-                    u, usig = obsh.vis_component(data, 'uvis', 'lin')
-                    out = q + 1j * u
-                    sig = np.sqrt(qsig**2 + usig**2)
-            elif field in ['m', 'mamp', 'mphase', 'msnr', 'msigma', 'msigma_phase']:
-                ty = 'c16'
-                if self.polrep == 'stokes':
-                    out = (data['qvis'] + 1j * data['uvis']) / data['vis']
-                    sig = obsh.merr(data['sigma'], data['qsigma'], data['usigma'], data['vis'], out)
-                elif self.polrep == 'circ':
-                    out = 2 * data['rlvis'] / (data['rrvis'] + data['llvis'])
-                    sig = obsh.merr2(data['rlsigma'], data['rrsigma'], data['llsigma'],
-                                     0.5 * (data['rrvis'] + data['llvis']), out)
-                elif self.polrep == 'lin':
-                    ivis, isig = obsh.vis_component(data, 'vis', 'lin')
-                    q, qsig = obsh.vis_component(data, 'qvis', 'lin')
-                    u, usig = obsh.vis_component(data, 'uvis', 'lin')
-                    out = (q + 1j * u) / ivis
-                    sig = obsh.merr(isig, qsig, usig, ivis, out)
-            elif field in ['evis', 'eamp', 'ephase', 'esnr', 'esigma', 'esigma_phase']:
-                ty = 'c16'
-                ang = np.arctan2(data['u'], data['v'])  # TODO: correct convention EofN?
-                q, qsig = obsh.vis_component(data, 'qvis', self.polrep)
-                u, usig = obsh.vis_component(data, 'uvis', self.polrep)
-                out = (np.cos(2 * ang) * q + np.sin(2 * ang) * u)
-                sig = np.sqrt(0.5 * ((np.cos(2 * ang) * qsig)**2 + (np.sin(2 * ang) * usig)**2))
-            elif field in ['bvis', 'bamp', 'bphase', 'bsnr', 'bsigma', 'bsigma_phase']:
-                ty = 'c16'
-                ang = np.arctan2(data['u'], data['v'])  # TODO: correct convention EofN?
-                q, qsig = obsh.vis_component(data, 'qvis', self.polrep)
-                u, usig = obsh.vis_component(data, 'uvis', self.polrep)
-                out = (-np.sin(2 * ang) * q + np.cos(2 * ang) * u)
-                sig = np.sqrt(0.5 * ((np.sin(2 * ang) * qsig)**2 + (np.cos(2 * ang) * usig)**2))
-            elif field in ['rrvis', 'rramp', 'rrphase', 'rrsnr', 'rrsigma', 'rrsigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'rrvis', self.polrep)
-            elif field in ['llvis', 'llamp', 'llphase', 'llsnr', 'llsigma', 'llsigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'llvis', self.polrep)
-            elif field in ['rlvis', 'rlamp', 'rlphase', 'rlsnr', 'rlsigma', 'rlsigma_phase']:
-                ty = 'c16'
-                out, sig = obsh.vis_component(data, 'rlvis', self.polrep)
-            elif field in ['lrvis', 'lramp', 'lrphase', 'lrsnr', 'lrsigma', 'lrsigma_phase']:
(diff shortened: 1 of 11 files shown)


tests/test_imager_e2e.py
```diff
@@ -152,6 +154,39 @@ def test_recovers_gaussian_stokes_i_diag_nfft(self, gauss_im, observe):
...
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")  # polrep vs feed_type mismatch + convention
+            out_other = image(obs_stokes.switch_polrep(polrep, singlepol_hand=hand))
+
+        (errors, _, _) = out_other.compare_images(
+            gauss_im, metric=["nxcorr"], blur_frac=0.0)
+        assert errors[0] > STOKES_I_NXCORR_MIN
+        # unpolarized source -> exact I synthesis -> identical reconstruction
+        np.testing.assert_allclose(out_other.imvec, out_stokes.imvec,
+                                   rtol=0, atol=1e-10)
```
rohandahale: `test_stokes_i_closure_imaging_polrep_transparent[circ]` and `[lin] `both fail in my env (numpy 2.4 / scipy 1.17): 1024/1024 pixels mismatch, max rel diff ~0.26-0.37. I verified the data products fed to the imager are identical to machine precision (amp 1e-16, cphase 4e-14, logcamp  ~1e-12).  

The I-synthesis is exact. But L-BFGS-B amplifies those ~1e-12 input deltas into ~1e-3 image differences. This is the exact BLAS/scipy-trajectory drift that made PR #232 floor the nxcorr threshold at 0.80. The test then contradicts that with a 1e-10 exact imvec assertion. May be it passed for you by luck of scipy 1.13's reduction order. 
  
Fix: drop the `assert_allclose(imvec, atol=1e-10)` line. Assert the real invariant on the data products (`unpack('amp'/'cphase'/'logcamp')` equal to ~1e-12 between the stokes obs and its switch_polrep) before imaging, plus the existing nxcorr check. 

If you want an image-level check, use `nxcorr(out_other, out_stokes) > 0.99,` not raw imvec.

ehtim/obsdata.py
```diff
@@ -3540,7 +3497,8 @@ def c_log_amplitudes_diag(self, vtype='vis', mode='all', count='min',
                 design_mat[il, ind4] = -1.0
 
             # construct the covariance matrix
-            logvisamp_err = viss_here[sigmatype] / np.abs(viss_here[vtype])
+            vis_here, sig_here = obsh.vis_component(viss_here, vtype, self.polrep)
+            logvisamp_err = sig_here / np.abs(vis_here)
             sigma_mat = np.diag(logvisamp_err**2.0)
```
rohandahale, line 3521: `logcamp_diag` crashes:

The diag covariance is assembled over all baselines at a timestamp with no polbasis filter. On a mixed obs, vis_component NaN fills the heterogeneous baselines. Those NaNs land on sigma_mat's diagonal, and design_mat @ sigma_mat @ design_mat.T does 0*NaN=NaN, for the surviving triangle. 

`c_log_amplitudes_diag` raises IndexError. 

The existing test only asserts len(d)>=1, so it passes despite the NaN.

Fix: restrict `viss_here` to the baselines actually referenced by the surviving (polbasis filtered) closures before building sigma_mat, the same filter the maximal set paths use. Then strengthen `test_c_phases_diag_mixed_skips_cross_feed` to assert n`p.all(np.isfinite(sigmacp))` and add a `c_log_amplitudes_diag` mixed  test.

ehtim/obsdata.py
```diff
@@ -2911,8 +2830,9 @@ def c_phases_diag(self, vtype='vis', count='min', ang_unit='deg',
                 ind3 = ((viss_here['t1'] == cp['t3']) & (viss_here['t2'] == cp['t1']))
                 design_mat[ic, ind3] = 1.0
 
-            # construct the covariance matrix
-            visphase_err = viss_here[sigmatype] / np.abs(viss_here[vtype])
+            # construct the covariance matrix (per-baseline visibility, sigma pair)
+            vis_here, sig_here = obsh.vis_component(viss_here, vtype, self.polrep)
+            visphase_err = sig_here / np.abs(vis_here)
             sigma_mat = np.diag(visphase_err**2.0)
             covar_mat = np.matmul(np.matmul(design_mat, sigma_mat), np.transpose(design_mat))
```
rohandahale, line 2845: Mixed diagonal closures produce NaN sigmas:

The diag covariance is assembled over all baselines at a timestamp with no polbasis filter. On a mixed obs, vis_component NaN fills the heterogeneous baselines. Those NaNs land on sigma_mat's diagonal, and design_mat @ sigma_mat @ design_mat.T does 0*NaN=NaN, for the surviving triangle. 

`c_phases_diag(vtype='rrvis')` returns cphase=63.43° but `sigmacp=NaN`

The existing test only asserts len(d)>=1, so it passes despite the NaN.

Fix: restrict `viss_here` to the baselines actually referenced by the surviving (polbasis filtered) closures before building sigma_mat, the same filter the maximal set paths use. Then strengthen `test_c_phases_diag_mixed_skips_cross_feed` to assert n`p.all(np.isfinite(sigmacp))` and add a `c_log_amplitudes_diag` mixed  test.

rohandahale (changes requested):
@achael Looks good overall! 

Two things in the inline comments:
 1. New e2e test fails deterministically, atol=1e-10 on a post-optimizer image. tests/test_imager_e2e.py:185-189. test_stokes_i_closure_imaging_polrep_transparent[circ] and [lin] both fail in my env (numpy 2.4 / scipy 1.17).
2. Mixed-feed diagonal closures silently produce NaN sigmas (and logcamp_diag crashes).
 
Other issues:
- Closure-amplitude family untested on lin/mixed (tests/test_mixedpol.py). make_closure_amplitude is one of only two changed bodies and c_amplitudes/camp_quad got new mixed quad-skip logic, but no test calls c_amplitudes/camp_quad/c_log_amplitudes_diag on a lin or mixed obs (bispectra and c_phases each got both a  lin-equality and a mixed-skip test; closure amps got neither). Add test_c_amplitudes_lin_matches_stokes + test_c_amplitudes_mixed_skips_cross_feed_quad.
- Backward-compat for moved unpack fields not regression-tested (obsdata.py:1010-1045). The ~130 lines of inline stokes/circ unpack for evis/bvis/m/rrllvis were deleted and routed through unpack_vis; the stokes-vs-circ equality test only covers amp/qamp/uamp/vamp/pvis. Add a parametrized equality test over [evis,bvis,m,rrllvis,rrvis,llvis,rlvis,lrvis] to ~1e-12.
- Mixed unpack tests are self-consistent, not independent (test_mixedpol.py:1846-1871). pvis==q+iu, m==(q+iu)/i, rrll==(i+v)/(i-v) all draw q,u,i,v from the same path, so a systematic feed-matrix error on rlxy baselines would pass. Fix: build the mixed obs from known Stokes via stokes_to_coherency per row, then  assert unpack recovers them on the heterogeneous rows.
- numpy>=2.0 floor + requires-python change bundled into a data PR (pyproject.toml). The numpy 1.24->2.0 / scipy / astropy / matplotlib bumps ride in on Phase 4c; the commit justifies numpy>=2.0 as a scalar-squeeze workaround, but the code already has explicit single-row handling. requires-python is also relaxed from #271's decided >=3.11,<3.13 back to >=3.10 with a 3.13 classifier added, while CI only tests 3.10/3.12 (3.11/3.13 unverified). Worth splitting the packaging change out, or calling it out explicitly for sign-off, especially with the v1.4 release merge pending.

Suggestions:
- feed_matrix docstring says "Always invertible" but accepts degenerate same-feed codes ('rr') -> singular LinAlgError deep in linalg. Not reachable via the validated pipeline, but either reject degenerate feeds at the boundary or soften the docstring + add the negative test (pol_conventions.py:319-331).
- correlation_slot uses str.find: add a 2-distinct-letter guard so a corrupt polbasis fails loudly rather than mis-resolving (pol_conventions.py:414-419).
-  data_conj re-validates polrep against a hardcoded tuple (obsdata.py:674): use the central definition or drop it (the constructor already enforces it).

ehtim/obsdata.py
achael: good catch, fixed by adding a filter on viss_here to limit it to rows with nonzero design matrix entries. Also added early return to catch cases where no closures survive.

ehtim/obsdata.py
achael: same fix as above: filter on viss_here to limit it to rows with nonzero design matrix entries. Also added early return to catch cases where no closures survive.

tests/test_imager_e2e.py
achael: fixed as suggested

achael:
addressed all your comments: 

- adjusted pass criterion on e2e test to 0.99 nxcorr, also added 1e-12 test on data products before imaging
- added filtering of rows without any closures in the design matrix in diagonal closure functions. added catch to return empty list if no closures exist (previously gave IndexError)
- new tests on closure amplitude family
- improved stokes-vs-circ regression test on r[evis,bvis,m,rrllvis,rrvis,llvis,rlvis,lrvis] 
- new test for mixed unpack building from known stokes
- numpy 2.0 not required; reverted pyproject.toml. Will later update to new requirements when merging in dev-backend
- added catch to feed_matrix to reject degenerate feeds
- added guard on correlation_slot

rohandahale:
Thanks @achael! Looks good to me!

rohandahale (approved):

PR #284, opened 2026-06-02
Release prep for v1.4.0

rohandahale:
Documentation + small fixes ahead of the v1.4.0 release merge.

- CHANGELOG.md with the v1.4.0 entry (highlights, breaking changes, bug fixes).
- Status badges (PyPI, license, docs, Zenodo DOI, CI, codecov, ruff) on README and docs index.
- obsdata.py: use warnings.warn instead of print when pandas is missing.
- array.py: catch KeyError (not NameError) so the "no ephemeris for site X" error fires.
- obs_simulate.py: thermal noise is now seed-aware via cerror_hash, so add_jones_and_noise / add_noise are reproducible with a fixed seed. Default seed=False noise is now deterministic per datapoint rather than from process-global RNG.

Full suite green; +2 reproducibility tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,33 @@
+# Changelog
+
+## v1.4.0 (2026-06-02)
+
+### Highlights
+
+- **Pure-function imaging backend.** `Imager` now delegates the math to `ehtim.imaging.imager_backend`, paving the way for JAX support.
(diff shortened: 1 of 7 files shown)


achael (approved):
looks great, fixing the thermal noise with the seed will help Alex's efforts to promote the noise generation to mixed pol while keeping things identical for circular.

PR #282, opened 2026-05-29
Pre-release prep: bump to 1.4.0 + README update + self_cal ComplexWarning fix

rohandahale:
Release prep for the `dev-backend --> dev`, plus a small follow-up agreed on PR #276.

1. **Bump version 1.3.0 → 1.4.0.** : No breaking removals on dev-backend yet, so a minor bump is enough.

2. **README + docs/source/index.rst cleanup.** The pyNFFT installation section is obsolete since PR #253 swapped pyNFFT for FiNUFFT in May 2026, and the auto-installed dep list still listed `ephem` and `pandas` (gone after #271 and #252). Reconciled with `pyproject.toml`. New `Optional Dependencies` section documents `pandas` and `paramsurvey` install paths. Dead `scikit-image` mention removed (not imported anywhere in the codebase).

3. **self_cal `ComplexWarning` fix.** Follow-up on the side note from PR #276: `dchisq_dgr[i] = np.sum(complex_quantity)` at `self_cal.py:557-558` silently discarded the imaginary part. The FD parity in PR #276 confirmed the imaginary parts are numerically zero, so the warning was benign — but explicit `.real` at the source is cleaner than relying on the cast. Comment at line 583 (`# any imaginary parts??? should all be real`) acknowledged the uncertainty.

PR #279, opened 2026-05-28
diagnostics: add sumdown_lin + sumdown_img tests

rohandahale:
Closes the small `diagnostics.py` gap per `docs/coverage_status.md`. 6 new tests across `sumdown_lin` (1D segment summer) and `sumdown_img` (2D block summer), covering output shape, total conservation, and uniform-input correctness.

PR #277, opened 2026-05-28
Image regrid/rotate/shift/blur/plotting tests + matplotlib 3.10 contour fix

rohandahale:
Fills the partial-coverage gaps on `image.py` per `docs/coverage_status.md`. 12 new tests across `regrid_image`, `rotate`, `shift_fft`, `blur_mf`, `contour`, `display`, `overlay_display`.

### Bonus fix

`Image.contour()` was using `cs.collections[0].set_label(...)`, which raises `AttributeError: 'QuadContourSet' object has no attribute 'collections'` on matplotlib 3.10+ (the attribute was deprecated in 3.8 and removed in 3.10). Replaced with `cs.set_label(...)` — `QuadContourSet` is itself a legend artist in current matplotlib.

### Tests added

- `regrid_image` with linear/cubic/quintic interp + multi-pol Q/U/V flux conservation + psize/dim checks.
- `rotate` polarimetric tensor transform: 90 deg → Q' = -Q, U' = -U (TMS Ch. 4 / spin-2 polarization); arbitrary angle → `Q' + iU' = exp(2iα)(Q + iU)`.
- `shift_fft` sub-pixel multi-pol flux conservation.
- `blur_mf` with `fit_order=2` populates `specvec`.
- `contour`, `display`, `overlay_display` Agg-backend smoke tests.

PR #269, opened 2026-05-21
Add interactive Plotly plotting module

Lftw:
Closes part of #212.

Adds `ehtim/plotting/interactive.py`, a Plotly-based interactive companion to `comp_plots.py` / `summary_plots.py`. Plotly is an optional dependency and can be installed with `pip install plotly`.

Public functions:

- `plot_bl(obs, site1, site2, field)` - single-baseline time series
- `plotall(obs, field1, field2)` - any vs any scatter (uv coverage, radplot, etc.)
- `plot_gains(caltable, ...)` - per-site gain time series, amp/phase x R/L dropdown
- `dashboard(im, obs, caltable, ...)` - 2x2 reconstruction summary (image with EVPA ticks, data fit with 16 product selector, gains, D-terms)
- `display(fig)` / `write_html(fig, path)` - Color toolbar wired up for Jupyter and standalone HTML

`tutorials/ehtim_tutorial_interactive.ipynb` is a short walkthrough hitting every function on `sample.uvfits`.

`tests/test_interactive.py` has the smoke tests (tagged `slow`).

PR #285, opened 2026-06-04
Merge dev-backend into dev (v1.4.0 release)

rohandahale:
Merges the v1.4.0 from `dev-backend` into `dev`. 32 commits, all individually reviewed and merged on `dev-backend`; CI green at `2f703bb`. Full details in `CHANGELOG.md`.

Recommend merging with a **merge commit** (not squash) to preserve the per-PR history on `dev`.

## Highlights
- Pure-function imaging backend (`imager.py` -> `imager_backend.py`), JAX-ready
- NFFT migrated PyNFFT -> finufft (now in core deps)
- pandas + paramsurvey moved to the `[dev]` extra; pandas-free averaging in `ehtim.statistics.averaging`
- Calibration cluster characterised (caltable / self_cal / leakage / polgains / network test suites)
- New interactive Plotly plotting module (`ehtim.plotting.interactive`)
- Many latent-bug fixes: polchisq, `_diag` chisq NumPy >= 1.24, IP/IV chain rule, inverse-variance averaging, load_uvfits format typos, and more

## Breaking changes
- Python 3.11 or 3.12 (3.10 dropped)
- numpy >= 2.0, scipy >= 1.13, astropy >= 6.0
- finufft required for `ttype='nfft'`; PyNFFT removed
- pandas / paramsurvey now optional (`[dev]` extra)
- `ttype='fast'` deprecated; use `direct` or `nfft`

## Tests
1460+ tests green; memory baseline ~482 MB peak RSS.

## PRs in this release
#238 #239 #243 #246 #249 #251 #252 #253 #255 #256 #257 #258 #262 #263 #264 #265 #266 #267 #269 #271 #272 #273 #274 #276 #277 #278 #279 #280 #282 #284


rohandahale:
@achael Ready for v1.4.0 on dev! Please do merge instead of squash merge this time.

achael (commented):
looks good overall! just some minor notes
- updating README to be a .md file would be very useful and this seems like a good time to do it
- the PR notes that the stv_pol_grad bug is fixed, but this is already patched into v1.3

achael:
a more-annoying suggestion: I forgot we kept the dataframes.py module but wired all the averaging through the new module. I think that is fine by default, but i'm wondering if we should put back in the original averaging functions into obsdata.py behind a "use_pandas=True" flag. That way people would have access to the original functions if necessary for backwards compatibility. We could also add a one-time warning when people use the averaging functions for the first time that the default has switched off the pandas code. 

I don't think a similar guard is necessary for nfft since we can guarantee the chisqs and gradients are identical to before. 

Feel free to push back on this if you think its overkill.

rohandahale:
Thanks @achael! On the `use_pandas` flag: I think it's already covered without re-wiring pandas into `obsdata`. The new averaging reproduces the legacy result exactly with `invvar_avg=False`, and the original `dataframes.py` functions are still intact and importable directly, so a `use_pandas=True` flag would mostly duplicate `invvar_avg=False`. I added the one-time `FutureWarning` you suggested instead.

That, plus the README --> Markdown conversion and CI-on-`dev` (so the CI/codecov badges show real status on the release branch), is in #289 onto `dev-backend`, so this PR will pick it up. And you're right that the `stv_pol_grad` fix is already in v1.3, not new to v1.4. I have removed it from the notes above. 


So we have to merge #289 first and then this PR.

rohandahale:
@achael #289  is now merged and I have rebased this PR. Now it is ready to be merged.

PR #286, opened 2026-06-05
Add a NumPy/JAX backend switch and a differentiable visibility objective

rohandahale:
Adds the first piece of the JAX port: a backend switch and a differentiable Stokes-I visibility objective. The numpy backend is unchanged.

- `ehtim/backends.py` — `set_backend` / `get_backend` / `backend` (context manager) and `array_namespace`. Selects numpy or jax; jax is imported lazily, only when selected.
- `ehtim/imaging/objective.py` — backend-agnostic `chisq_vis` / `loss_vis_log`, and a `vis_objective(A, vis, sigma)` factory returning a scipy `fun(x) -> (f, g)`. Under jax the gradient is `jax.grad` (jitted, GPU-capable); under numpy it is the analytic `chisqgrad_vis`. The image is parameterized as `imvec = exp(x)`.

First step of the JAX port (extract -> simplify -> JAXify). It proves the core premise: autodiff reproduces the hand-written gradient, so the `*grad` functions become a validation oracle rather than code to maintain by hand.

## Tests
`tests/test_backends.py` (the switch) and `tests/test_objective.py`:
- three-way gradient: `jax.grad` == analytic `chisqgrad_vis * exp(x)` == central finite difference, in float64
- jit == eager, no-retrace, x64-holds-through-jit
- numpy/jax value+grad parity
- L-BFGS-B recovers the source image (both backends)
- GPU device tests (local only)

CI runs numpy only (`-m "not jax"`); JAX (CPU and GPU) is validated locally.

## Install
`pip install -e ".[dev]"` (CPU jax) or `pip install -e ".[gpu]"` (CUDA, installs `jax[cuda12]` + GPU jaxlib).

## Local GPU run (RTX 6000 Ada, jax 0.10.1, float64)
```
jax 0.10.1 [CudaDevice(id=0)]
pytest -m gpu tests/test_objective.py  ->  3 passed
```
`value_and_grad` on npix=1024, nvis=1030: **GPU 0.89 ms vs CPU 15.2 ms** (median, 100 runs, post-warmup). A tiny problem; the gap grows with image/data size.

Memray peak RSS 516.1 MB, identical to the parent commit (the new modules are not imported by `import ehtim`, so the numpy imaging path is unchanged).

## Note
The switch is `from ehtim.backends import set_backend` for now. Exposing it as `ehtim.set_backend` touches `__init__.py`'s pre-existing star-import lint, so it is deferred to a separate cleanup.


rohandahale:
Superseded by #291, which contains this (the backend switch + vis objective) plus the broadening to all six direct Stokes-I chisq terms, the embed-free regularizers, and the top-level `ehtim.set_backend`. Closing in favor of the combined PR.

PR #291, opened 2026-06-05
Add a NumPy/JAX backend switch and differentiable Stokes-I imaging objectives

rohandahale:
Adds a NumPy/JAX backend switch and makes the direct Stokes-I imaging kernels backend-agnostic, with gradients from `jax.grad` validated against the analytic gradients and finite differences. The NumPy backend is byte-identical.

- `ehtim/backends.py` — `set_backend("jax")` / `get_backend()` / `backend()` context manager + `array_namespace`. Kernels read their array module from their inputs (numpy or jax.numpy), so one source runs on both backends; jax is imported lazily, and `array_namespace` memoizes jax-availability so the numpy hot path pays nothing.
- `ehtim/imaging/imager_utils.py` — the existing direct Stokes-I `chisq_*` (vis/amp/bs/cphase/camp/logcamp) and embed-free `reg_*` (flux/simple/l1/l1w/lA/gs) value kernels are now backend-agnostic (`np.` → `array_namespace`). **Single source:** `compute_chisq_dict` / `compute_objective` already call them, so the production objective is jax-ready. The analytic `chisqgrad_*` / `reggrad_*` stay as the numpy gradient path and the jax validation oracle.
- `ehtim/imaging/imager_backend.py` — a `Notes` block on `compute_objective` + `TODO(jax)` at `unpack_imarr` / `transform_imarr`: `jax.grad(compute_objective)` becomes the jax objective once those two are rewritten without in-place assignment (next PR — we reuse `compute_objective`, not build a parallel one).
- `ehtim/__init__.py` / `ehtim/const_def.py` — `const_def` declares `__all__`; `__init__` re-exports it via a star import so new constants auto-propagate, and the accidentally-public names left by the old star imports are pruned.

Under jax the gradient is `jax.grad` (jitted `value_and_grad`, GPU-capable); under numpy it is the analytic `chisqgrad_*` / `reggrad_*`. Each kernel is validated three ways: **`jax.grad` == analytic == finite difference**, in float64.

## Tests
`test_backends.py` (the switch), `test_chisq_jax.py` (6 chisq terms), `test_regularizers_jax.py` (6 regularizers): three-way gradient, jit==eager, no-retrace, numpy/jax value parity, and GPU device tests — all validating the `imager_utils` kernels directly. All `@pytest.mark.jax`, so CI runs numpy-only (`-m "not jax"`); JAX (CPU and GPU) runs locally.

## Note for reviewers
This supersedes the earlier per-term `objective.py` / `regularizers.py` modules on this branch: their kernels duplicated the `imager_utils` ones, so they were removed and the kernels were single-sourced in place. Inline comments on those (now-deleted) files are therefore moot — the equivalent functions live in `imager_utils.py`.


ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
+"""Direct Stokes-I chi-squared objectives with a NumPy/JAX backend switch.
```
achael: it might help me see where things are going if you put a little more signposting in the docstring re what the next steps are on adding the nfft ttype and polarization -- will these live here or elsewhere / how much will need to be modified?

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    chisqgrad_amp,
+    chisqgrad_bs,
+    chisqgrad_camp,
+    chisqgrad_cphase,
+    chisqgrad_logcamp,
+    chisqgrad_vis,
+)
+
+
+def chisq_vis(imvec, A, vis, sigma):
```
achael: do we want these to have the same names as the equivalent functions in imager_utils.py? If it's not necessary for the architecture, maybe add a _backend label? That might make things easier while we maintain both the original and the new versions, though perhaps the different module prefixes are enough.

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    "vis": chisq_vis, "amp": chisq_amp, "bs": chisq_bs,
+    "cphase": chisq_cphase, "camp": chisq_camp, "logcamp": chisq_logcamp,
+}
+_CHISQGRAD = {
+    "vis": chisqgrad_vis, "amp": chisqgrad_amp, "bs": chisqgrad_bs,
+    "cphase": chisqgrad_cphase, "camp": chisqgrad_camp, "logcamp": chisqgrad_logcamp,
+}
+
+
+def _loss_log(dtype, x, A, data, sigma):
```
achael: it might be useful to think about if we want a more flexible image transformation architecture for when we implement polarization. This is probably enough for now, but maybe note as a todo item that we want a uniform wrapper? Then we could also more easily experiment with different Stokes I transforms as well.

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    """chisq[dtype](exp(x), ...) - 1 in log-image coordinates."""
+    xp = array_namespace(x)
+    return _CHISQ[dtype](xp.exp(x), A, data, sigma) - 1.0
+
+
+def loss_vis_log(x, A, vis, sigma):
```
achael: is this the only one currently defined for now because we are just testing with 'vis' and not closures?

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    """Visibility objective in log-image coordinates: chisq_vis(exp(x)) - 1."""
+    return _loss_log("vis", x, A, vis, sigma)
+
+
+@functools.cache
+def _jitted_value_and_grad(dtype):
+    """Per-dtype jit(value_and_grad(loss)); built once, dtype baked into the closure."""
+    import jax
```
achael: just for my curiosity not having worked with jax before, why is it imported separately inside each function here?

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    if dtype not in _CHISQ:
+        raise ValueError(f"dtype must be one of {sorted(_CHISQ)}, got {dtype!r}")
+
+    if get_backend() == "jax":
+        import jax
+        import jax.numpy as jnp
+
+        A_d = _device_put_A(A, device)
+        data_d = jax.device_put(jnp.asarray(data), device)
```
achael: again more for my curiosity, why is the dtype defined for sigma and not for the data? Is it just because the data can be either complex or real?

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@functools.cache
+def _jitted_value_and_grad(dtype):
+    """Per-dtype jit(value_and_grad(loss)); built once, dtype baked into the closure."""
+    import jax
+    return jax.jit(jax.value_and_grad(functools.partial(_loss_log, dtype)))
+
+
+def _device_put_A(A, device):
```
achael: does it make sense to extend this wrapper to put all of A,data,sigma on the device since we always expect to work with all 3?

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    return _CHISQ[dtype](xp.exp(x), A, data, sigma) - 1.0
+
+
+def loss_vis_log(x, A, vis, sigma):
+    """Visibility objective in log-image coordinates: chisq_vis(exp(x)) - 1."""
+    return _loss_log("vis", x, A, vis, sigma)
+
+
+@functools.cache
+def _jitted_value_and_grad(dtype):
```
achael: probably calling this "_jitted_value_and_grad_log_loss" would be clearer, unless the intent is to extend this to future transforms/other functions beyond the log loss.

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+        data_d = jax.device_put(jnp.asarray(data), device)
+        sigma_d = jax.device_put(jnp.asarray(sigma, jnp.float64), device)
+        value_and_grad = _jitted_value_and_grad(dtype)
+
+        def fun(x):
+            value, grad = value_and_grad(jnp.asarray(x), A_d, data_d, sigma_d)
+            return float(value), np.asarray(grad, dtype=np.float64)
+
+        return fun
+
```
achael: idk what the official preferred style here is, but i always prefer explicit elses on the if statements, even if its not necessary here because the if statement has a return under it.

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    grad_fn = _CHISQGRAD[dtype]
+
+    def fun(x):
+        value = float(_loss_log(dtype, x, A, data, sigma))
+        grad = transform_gradients(grad_fn(np.exp(x), A, data, sigma), x, ["log"], np.array([1]))
```
achael: comment on the signature of transform_gradients for reference (i forgot what the np.array([1]) is doing here.

ehtim/imaging/objective.py
```diff
@@ -0,0 +1,148 @@
...
+    a1, a2, a3, a4 = (xp.abs(A[k] @ imvec) for k in range(4))
+    samples = xp.log(a1) + xp.log(a2) - xp.log(a3) - xp.log(a4)
+    return xp.sum(xp.abs((log_clamp - samples) / sigma) ** 2) / log_clamp.shape[0]
+
+
+_CHISQ = {
+    "vis": chisq_vis, "amp": chisq_amp, "bs": chisq_bs,
+    "cphase": chisq_cphase, "camp": chisq_camp, "logcamp": chisq_logcamp,
+}
+_CHISQGRAD = {
```
achael: maybe _CHISQGRAD_ANALYTIC to indicate these are only used when we don't use jax?

ehtim/imaging/objective.py
achael: also a comment on the eventual goal of these functions vis-a-vis the ones in imager_utils --> are we eventually going to replace/merge or will they live side by side?

achael: other minor nitpick --- I interpret the objective as being the sum of the chisqs and regularizers, so maybe this should be renamed likelihoods.py or dataterms.py or chisquareds.py?

ehtim/imaging/regularizers.py
```diff
@@ -0,0 +1,99 @@
...
+    flux = kwargs["flux"]
+    norm = flux if kwargs.get("norm_reg", True) else 1
+    return -xp.sum(imvec - priorvec - imvec * xp.log(imvec / priorvec)) / norm
+
+
+_REG = {
+    "flux": reg_flux, "simple": reg_simple, "l1": reg_l1,
+    "l1w": reg_l1w, "lA": reg_lA, "gs": reg_gs,
+}
+_REGGRAD = {
```
achael: as above, maybe _REGGRAD_ANALYTIC?

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
-from builtins import object
-
-import ehtim.observing
-from ehtim.const_def import *
+# The import order below is constrained by circular dependencies between ehtim
+# submodules; alphabetical sorting breaks `import ehtim`, so isort is disabled.
+# isort: off
+import ehtim.observing as observing
+from ehtim.const_def import (
+    BHIMAGE,
```
achael: is there a way to define the list of globals to be imported at the top of const_def and then pull them all in here? in this version if we define something new there we'd have to make sure to add it to __init__ manually to make sure it shows up in the global namespace (which is the intention for everything in const_def.py)

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+    RADPERUAS,
+    RA_DEFAULT,
+    RA_M87,
+    RA_SGRA,
+    RF_DEFAULT,
+    SCOLORS,
+    SOURCE_DEFAULT,
+    TAUDEF,
+    amp_poldict,
+    show_noblock,
```
achael: i don't think show_noblock is necessary anymore -- probably not necessary to pull in here

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+    sig_poldict,
+    trianglePulse2D,
```
achael: I think we just want PULSE_DEFAULT -- do we need to pull in trianglePulse2D to make that work?

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+)
 from ehtim.plotting import comparisons
 from ehtim.plotting import comp_plots
-import ehtim.plotting
+import ehtim.plotting as plotting
 from ehtim.calibrating.network_cal import network_cal as netcal
 from ehtim.calibrating.self_cal import self_cal as selfcal
-from ehtim.calibrating.pol_cal import *
-from ehtim.calibrating.pol_cal_new import *
+from ehtim.calibrating.pol_cal import (
```
achael, line 56: add a TODO comment above here that we should replace pol_cal with pol-cal in the very near future

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+    leakage_cal,
+    plot_leakage,
+)
+from ehtim.calibrating.pol_cal_new import (
+    MAXIT,
```
achael: i don't think we want MAXIT,MAXLS,NHIST,STOP in the global namespace (I realize they were before because of the * import but they shouldn't be).

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+import ehtim.movie as movie
+import ehtim.image as image
+import ehtim.model as model
+import ehtim.survey as survey
 
 
 import warnings
+# isort: on
+
+__all__ = [
```
achael, line 83: what does __all__ do here?

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+import ehtim.imaging as imaging
+from ehtim.backends import backend, get_backend, set_backend
 from ehtim.features import rex
-import ehtim.features
-from ehtim.plotting.summary_plots import *
-from ehtim.plotting.comparisons import *
-from ehtim.plotting.comp_plots import *
+import ehtim.features as features
+from ehtim.plotting.summary_plots import (
+    FONTSIZE,
```
achael: I don't think we want FONTSIZE, HSPACE,MARGINS,MARKERSIZE,PROCESSES,WSPACE in the global namespace -- i realize they were before, but they shouldn't have been

ehtim/__init__.py
```diff
@@ -6,46 +6,275 @@
...
+)
+from ehtim.plotting.comparisons import (
+    change_cut_off,
+    generate_consistency_plot,
+    get_psize_fov,
+    image_agreements,
+    image_consistency,
+)
+from ehtim.plotting.comp_plots import (
+    COLORLIST,
```
achael: super minor but could move COLORLIST to const_def as the placeholder for unified library color palette.

achael (commented):
looks good overall! just some minor comments, mostly to help me figure out where the larger architecture is going.

ehtim/imaging/objective.py
rohandahale: 1) I removed this module. The kernels are now the agnostic ones in `imager_utils`. Roadmap signposting for nfft/pol + the objective lives in the Notes block on compute_objective (imager_backend.py).

2) I resolved by single-sourcing, the kernels are imager_utils' own (made backend-agnostic in place); this separate module is gone, so there's no side-by-side anymore.

3) Same as above.

ehtim/imaging/objective.py
rohandahale: No second copy now. These are imager_utils.chisq_* (made agnostic), so there's nothing to disambiguate.

ehtim/__init__.py
rohandahale: The dict went with the module. The analytic gradients are imager_utils.chisqgrad_*, dispatched via imager_backend._CHISQ_DISPATCH (numpy path + jax oracle).

ehtim/imaging/objective.py
rohandahale: The production transform is already transform_imarr (imager_backend), it handles log + the pol transforms (mcv/vcv). I added a TODO (jax) there to make it functional for the jax objective.

ehtim/imaging/objective.py
rohandahale: Gone now

ehtim/__init__.py
rohandahale: Moved into const_def (= SCOLORS); comp_plots references ehc.COLORLIST.

ehtim/__init__.py
rohandahale: Defines the public API (from ehtim import *) and the explicit re-export contract now the star imports are gone; const_def's constants are appended from const_def.__all__.

ehtim/__init__.py
rohandahale: Dropped all four

ehtim/__init__.py
rohandahale: Added above the pol_cal import.

ehtim/__init__.py
rohandahale: Dropped, PULSE_DEFAULT is its own binding, works standalone.

ehtim/__init__.py
rohandahale: Dropped

ehtim/__init__.py
rohandahale: Done, const_def declares __all__
 __init__ does from const_def import * and appends const_def.__all__ to ehtim.__all__, so new constants auto-propagate.

ehtim/imaging/regularizers.py
rohandahale: Module removed. The analytic reg gradients are imager_utils.reggrad_*.

ehtim/imaging/objective.py
rohandahale: The dict went with the module, The analytic gradients are imager_utils.chisqgrad_*, dispatched via imager_backend._CHISQ_DISPATCH (numpy path + jax oracle).

ehtim/imaging/objective.py
rohandahale: Kept where it's used (the test): it applies the imvec = exp(x) chain rule; the last arg (which_solve = [1]) marks the single Stokes-I image active.

ehtim/imaging/objective.py
rohandahale: That if/else was removed with the module.

ehtim/imaging/objective.py
rohandahale: Moved into test_chisq_jax.py as _value_and_grad. It is no longer library code

ehtim/imaging/objective.py
rohandahale: Done, the test's _to_device(A, data, sigma) places all three.

ehtim/imaging/objective.py
rohandahale: Right,  data can be complex (vis, bs), so jnp.asarray infers it. sigma is always real, pinned float64.

ehtim/imaging/objective.py
rohandahale: Lazy import keeps import ehtim jax-free. jax loads only on set_backend('jax') (or when a jax array reaches array_namespace). It's in backends.py now.

rohandahale:
@achael Thanks for the review! Your comments pushed me to think more deeply about the implementation and I had to restructure it. So a heads-up on the new state:

Single-sourced the kernels. objective.py / regularizers.py on this branch held backend-agnostic copies of the chisq/reg kernels that already live in imager_utils.py. Rather than maintain two sets, I made the existing imager_utils kernels backend-agnostic in place (np. --> array_namespace(imvec), identical on numpy) and removed both modules. So _CHISQ_DISPATCH / compute_chisq_dict / compute_objective now run the same kernels jax uses. 

compute_objective already is the weighted chisq+reg loss, so the plan is to reuse it. jax.grad(compute_objective) becomes the jax gradient, rather than build a parallel objective. The only blockers are the two in-place-mutation spots (unpack_imarr, transform_imarr). I added a Notes block on compute_objective and TODO(jax) markers there. That's the next PR.

  __init__ / const_def: done as you suggested. const_def now declares __all__ and __init__ re-exports it via a star import so new constants auto-propagate. I pruned the accidentally-public names (show_noblock, trianglePulse2D, the FONTSIZE/MARKERSIZE/… plotting scalars, MAXIT/MAXLS/NHIST/STOP), moved COLORLIST into const_def, and added the pol_cal-->pol-cal TODO.

Tests now validate the imager_utils kernels directly (test_chisq_jax.py, test_regularizers_jax.py): three-way jax.grad == analytic == finite-difference, jit, numpy/jax parity, GPU.

ehtim/imaging/imager_backend.py
```diff
@@ -483,9 +488,15 @@ def transform_imarr_inverse(imarr, transforms, which_solve):
...
-    """Apply chain rule gradients for solver values for all polarizations
-       gradarr is objective func gradients w/r/t physical variables
-       imarr is the current image in solver variables """
+    """Apply chain rule gradients for solver values for all polarizations.
+
+    gradarr is objective func gradients w/r/t physical variables.
+    imarr is the current image in solver variables.
+    transforms lists the active solver transforms: 'log' applies the
+    imvec = exp(x) chain rule, 'mcv'/'vcv'/'polcv' the polarization ones.
+    which_solve is the per-Stokes solve mask; np.array([1]) means a single
```
achael: I think the line about np.array([1]) being for a single Stokes I image is wrong -- when nimage==1, transform_gradients just ignores which_solve entirely. probably the ``correct'' default for stokes I imaging should be None or np.array((1,0,0,0))

ehtim/imaging/imager_utils.py
```diff
@@ -1785,7 +1792,7 @@ def reg_lA(imvec, mask, **kwargs):
         norm = (norm_l1 * weight_l1 + norm_l0 * weight_l0) / (weight_l0 + weight_l1)
     else:
         norm = 1
-    return np.sum(fA(imvec, flux, alpha_A)) / norm
+    return array_namespace(imvec).sum(fA(imvec, flux, alpha_A)) / norm
```
achael: this is the only grad you've added the array_namespace to -- is that intentional?

ehtim/__init__.py
achael: my comment was garbled by typing too fast and not checking :/ what I meant was to note that we should replace pol_cal with pol_cal_new in the near future, since pol_cal_new (D term cal version with manual gradients) has been validated at this point. Then hopefully in the next few months both can be folded into the new calibration package that can be run simultaneously with the imager.

ehtim/__init__.py
achael: great, thanks for cleaning up the __init__ !

achael (dismissed):
looks great! The new design has really streamlined things :) just a few super minor typo corrections, but overall looks good to go.

ehtim/imaging/imager_backend.py
rohandahale: Good catch, I fixed it! which_solve is overwritten to (1,0,0,0) for nimage==1/3 and the nimage==1 branch doesn't  touch it. Corrected the docstring and switched the tests to the (1,0,0,0) mask.

ehtim/imaging/imager_utils.py
rohandahale: reg_lA is the value function (gradients stay numpy). The different thing was the inline array_namespace(imvec).sum(...) --> now it is xp = array_namespace(imvec) and uses xp.sum(...) like every other kernel.

achael (approved):

PR #292, opened 2026-06-09
Add PyPI publish workflow on tagged releases

achael:
## What
Adds `.github/workflows/publish.yml`: builds an sdist + wheel and uploads to PyPI when a GitHub **release** is published.

## How it works
- Trigger: `on: release: [published]` (creating a GitHub Release with a tag fires it).
- Uses **PyPI trusted publishing (OIDC)** via `pypa/gh-action-pypi-publish` — no API token stored in secrets.
- Build job pinned to **Python 3.11** (current `main` only supports <=3.11; bump when `dev` merges).
- `publish` job runs in a `pypi` environment, giving an optional manual-approval gate.

## Required one-time setup before first publish
1. **PyPI**: add a trusted publisher for project `ehtim` — owner `achael`, repo `eht-imaging`, workflow `publish.yml`, environment `pypi`.
2. **GitHub**: create an Environment named `pypi` (optionally with a required-reviewer protection rule).

## How to test
After the setup above, draft a GitHub Release on a tag and publish it; the workflow builds and uploads to PyPI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,33 @@
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - run: pip install build
+      - run: python -m build
+      - uses: actions/upload-artifact@v4
+        with:
+          name: dist
+          path: dist/
+
+  publish:
+    needs: build
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write   # required for trusted publishing (OIDC)
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist/
+      - uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,9 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


rohandahale:
@achael this looks great! A few things before we push the release:

1) Release ordering (the main one): for a release: published event the workflow runs from main, and the build job's checkout resolves to the tag. So everything needs to be on main first. Concretely: merge #292 + #293 + #294, then draft a new release --> create tag v1.3.1 0 --> Target main HEAD --> Publish. If the tag is push from a commit missing #294, python -m build stamps the artifact 1.3.0 instead of 1.3.1.

2) Pre-releases also publish: types: [published] fires for pre-releases too, so a release flagged "pre-release" would still push 1.3.1 to PyPI as a final. Worth guarding the publish job:
    `  if: ${{ !github.event.release.prerelease }}`
  (or switch to types: [released]).

 3) SHA-pin the publish action: pypa/gh-action-pypi-publish@release/v1 is a branch ref (mutable), and it runs in the id-token: write job, so I would pinn to a commit SHA (`@<sha> # release/v1`).

4)  skip-existing for safe re-runs:
```
          with:
            skip-existing: true
```
  so an accidental re-trigger of the same version is a no operation instead of a error 400.
  
3) Zenodo: since part of the goal is getting 1.3.1 into Zenodo (1.3 missed it), just a reminder that's the GitHub <--> Zenodo webhook, not this workflow, worth confirming the integration is enabled on the repo so it doesn't miss again.

Minor/optional: add a top-level` permissions: contents: read` so the build job is least rights regardless of repo defaults.

rohandahale (dismissed):

achael:
Thanks @rohandahale , all addressed: 
1. I will not draft v1.3.1 until all 3 PRs #292 , #293, #294 are merged
2. pre-release guard added
3. publish action SHA - pinned
4. added skip-existing
5. verified separate zenodo integration is fixed (it was broken due to lapsed credentials so the last few releases did not get published).

rohandahale:
@achael Thanks! Good to merge!

rohandahale (approved):

PR #293, opened 2026-06-09
Add docs build/publish workflow on tagged releases

achael:
## What
Automates documentation builds/publishing on release, and makes the docs self-update their version.

### Workflow
Adds `.github/workflows/docs.yml`: builds the Sphinx HTML docs and deploys them to the `gh-pages` branch when a GitHub **release** is published. Replaces the manual `make htmlpush` step.

- Triggers: `on: release: [published]`, plus `workflow_dispatch` for manual rebuilds from the Actions tab.
- Installs `ehtim` (autodoc imports the package) + `sphinx` + `sphinx_rtd_theme`, pinned to **Python 3.11** (current `main` only supports <=3.11; bump when `dev` merges).
- Builds with `sphinx-build` directly (no Makefile / sibling-checkout needed).
- Deploys via `peaceiris/actions-gh-pages@v4` using the built-in `GITHUB_TOKEN` — no extra secret required.

### Version auto-tracking (conf.py / index.rst)
- `version`/`release` in `conf.py` are now read from the installed `ehtim` package metadata instead of being hardcoded (was stale at `1.2.11`). Single source of truth = the version in `pyproject.toml`.
- The "latest stable version" link in `index.rst` now tracks `release` via an `rst_epilog` `|stable_release|` substitution, so it points at `releases/tag/v<release>` automatically.
- Fixed `intersphinx_mapping` to the new dict format — the old format is fatal under modern Sphinx, so the build (and this workflow) would otherwise fail.

## Notes
- `sphinx.ext.githubpages` is already enabled in `conf.py`, so `.nojekyll` is handled automatically.
- GitHub Pages stays configured to serve from the `gh-pages` branch (no settings change).
- Pre-existing, non-fatal autodoc errors remain (`|m|`/`|v|` in `Image.mavg`/`vavg` docstrings read as substitutions) — out of scope here.

## How to test
Build locally with `sphinx-build -b html docs/source <out>` (verified: `index.html` is produced and the stable-version link renders as the installed package version). After merge, run the workflow manually via **Actions → Build and Publish Docs → Run workflow**, or publish a GitHub Release; the built HTML lands on `gh-pages`.


rohandahale:
@achael Looks great! 

Few suggestions:
  - SHA-pin `peaceiris/actions-gh-pages@v4`,  it's the one third-party action and it runs with repo-wide contents: write
  - Add a concurrency group so a manual `workflow_dispatch` rebuild and a release build can't race the gh-pages push:
  ```
concurrency:
    group: docs-deploy
    cancel-in-progress: true
```
  - Pin the docs deps (sphinx, sphinx_rtd_theme) or move them into a [project.optional-dependencies] docs extra and pip install -e .[docs], so a future Sphinx  major doesn't break release-time deploys.
  
The `|stable_release| s`ubstitution rebuilds releases/tag/v{release}, so it assumes the tag is exactly v + the pyproject version. Fine for v1.3.1; just worth keeping that tag convention consistent going forward (an old-style v1.4 tag for version 1.4.0 would 404).

rohandahale (dismissed):

achael:
thanks @rohandahale , all done

rohandahale:
@achael Thanks! Good to merge!

rohandahale (approved):

PR #295, opened 2026-06-09
JAX-differentiable imaging objective on GPU: direct + NFFT, Stokes-I + pol + mf

rohandahale:
> **Update (2026-06-17) — rebased onto current `dev-backend`.**
> The pol-gradient correctness fixes this PR originally surfaced are now **upstream**, so #295 is scoped to the **JAX implementation + tests**:
> - `epsilon_tv` on `reg_ptv`/`reg_vtv` and the `reggrad_ptv` first-row/col boundary masking → now on `dev-backend` via **#306** (alongside `physical_grad_slots`).
> - The V-pol (V/IV) gradient fix → **#296**, already merged to `main`.
> - This PR's own part-B gating patches were **stripped** in the rebase; the canonical gating (`physical_grad_slots`) is inherited from `dev-backend`.
>
> Validated on the rebased branch: **1784 numpy + 71 jax** green. The sections below are retained as history of where those fixes originated.

---

Makes the full imaging objective differentiable under JAX: `jax.grad(compute_objective)` now equals the hand-written `compute_objective_grad`, on CPU **and GPU**, across Stokes-I + polarization + multifrequency, `ttype` direct + NFFT, and all regularizers. Builds on #291's backend switch (`array_namespace`) by making the rest of the objective backend-agnostic. NumPy behavior is byte-identical throughout.

**Now also includes** the `use_jax` flag on `Imager.make_image` — an end-to-end jax reconstruction through scipy L-BFGS-B that recovers the source and matches the numpy recon — plus explicit GPU device + CPU↔GPU-parity tests for the full objective (direct + nfft). **Remaining (draft):** a full *multifrequency* imaging recon test (the mf forward has unit coverage; mf is deprioritized this week) and where to document the GPU jax-finufft build.

### What changed
- **direct objective**: `unpack_imarr`/`transform_imarr` made functional (no in-place mutation); new `make_objective_jax(...)` factory returns a scipy `fun(x)->(value, grad)` (closure over all-but-`x`, lazy jax import so `import ehtim` stays jax-free).
- **OO integration**: `make_image(use_jax=True)` builds the factory and runs scipy L-BFGS-B on the jax objective (numpy path unchanged when off; GPU by default on a CUDA box).
- **NFFT under jax-finufft**: `nufft2_backend` dispatches finufft (numpy, byte-identical) vs `jax_finufft.nufft2` (jax, differentiable, GPU). NumPy NFFT still uses **only finufft**. 6 chisq kernels.
- **all regularizers agnostic**: 6 embed-free (#291) + 9 spatial + full `embed()`/`embed_imarr()` functional scatter (full & partial mask) + 8 pol regs + 2 spectral.
- **two pol-TV gradient correctness fixes** (both surfaced by the autodiff-vs-analytic comparison, see below): `epsilon_tv` on `reg_ptv`/`reg_vtv` (default 0, byte-identical, matching `reg_tv`); and `reggrad_ptv` now zeros its back-neighbor terms on the first row/col — the boundary masking `reggrad_tv`/`reggrad_vtv` already have.
- **pol/mf forward**: `make_*_image`, `polcv/mcv/vcv` (legacy `mcv/vcv` dead `raise` guards kept numpy-only), `chisq_p/m/vvis`, `image_at_freq`.
- **tutorial**: `tutorials/ehtim_tutorial_jax.ipynb` — install (jax + the GPU jax-finufft source build), the `jax.grad(objfunc) == objgrad` demo (direct + nfft on the GPU), and an end-to-end `make_image(use_jax=True)` recon.

### Validation (`tests/test_objective_jax.py`, `@pytest.mark.jax`)
direct+nfft × {value parity, gold-standard grad parity, FD, factory, no-retrace} + embed/partial-mask + pol IP/IV (grad vs **FD ground truth**) + `make_image(use_jax=True)` recovery (matches numpy) + **GPU device & CPU↔GPU parity** (direct + nfft) — 22 tests, green on calypso (CPU + GPU). NumPy regressions green (458 backend+e2e, 146 pol, 102 pol-reg). Ruff clean.
- NFFT grad parity is **eps-limited, not a bug** (proven: residual scales 1e-9→~1e-6, 1e-12→~7e-10); tests use `nfft_eps=1e-12`.

###  Pre-existing bugs surfaced by the autodiff-vs-analytic comparison
1. **pol TV regularizers had no epsilon in their `sqrt`** (unlike `reg_tv`'s `epsilon_tv`) → singular gradient at smooth pixels. **Fixed here** via `epsilon_tv` (default 0, so byte-identical off).
2. **`reggrad_ptv` was missing the first-row/col boundary masking** that `reggrad_tv`/`reggrad_vtv` have → the entire first row + column of the magnitude-gradient slots (I/m/psi) was wrong (corner 4× off vs finite-difference; phase slot was fine because its numerators self-zero at the pad). **Fixed here**; FD-matched to ~1e-10 everywhere after, and a new full-grid regression test (`test_reggrad_ptv_matches_fd_on_boundary`) fails pre-fix.
3. **The V-pol analytic gradient is wrong** (`chisqgrad_vvis` / `vcv_grad`): for IV imaging jax autodiff **and** finite differences agree, but the hand-written analytic dropped the physical `rho` gradient the `vcv` chain consumes (solver `v'`-gradient ~87% off). Pre-existing (the jax port didn't touch the analytic grad kernels) — like `stv_pol_grad` (#240). **Fixed in standalone PR #296** (independent function, no overlap with this PR — kept separate so the correctness fix can merge fast).

### DIFF
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -442,7 +442,8 @@ class Imager:
                grads (bool): whether or not to use image gradients
                use_jax (bool): use the jax autodiff objective (GPU-capable) in place
                    of the numpy analytic gradient; the objective is identical
-               jax_device: jax device for the objective when use_jax=True (None = jax default)
+               jax_device: jax device for the objective when use_jax=True; None uses the
+                   jax default. e.g. jax.devices('gpu')[0] or jax.devices('cpu')[0]
 
                show_updates (bool): whether or not to show imager progress
                update_interval (int): step interval for plotting if show_updates=True
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -58,6 +58,7 @@ from ehtim.imaging.imager_backend import (
     compute_objective_grad,
     compute_reg_dict,
     compute_reggrad_dict,
+    make_objective_jax,
     transform_imarr,
     unpack_imarr,
     validate_limits,
@@ -439,6 +440,9 @@ class Imager:
            Args:
                pol (str): which polarization to image
                grads (bool): whether or not to use image gradients
+               use_jax (bool): use the jax autodiff objective (GPU-capable) in place
+                   of the numpy analytic gradient; the objective is identical
+               jax_device: jax device for the objective when use_jax=True (None = jax default)
 
                show_updates (bool): whether or not to show imager progress
                update_interval (int): step interval for plotting if show_updates=True
@@ -508,7 +512,19 @@ class Imager:
 
 
         tstart = time.time()
-        if grads:
+        if kwargs.get('use_jax', False):
+            # jax autodiff objective (GPU-capable). make_objective_jax returns a
+            # scipy fun(x) -> (value, grad); jax is imported lazily inside it.
+            fun = make_objective_jax(
+                self._init_arr, self._config, self._which_solve, self._data_tuples,
+                self._logfreqratio_list, len(self.obslist_next),
+                self.dat_term_next, self.reg_term_next,
+                self._prior_arr, self.norm_reg, self._regparams(),
+                self._embed_mask, device=kwargs.get('jax_device', None),
+            )
+            res = opt.minimize(fun, self._init_vec, method='L-BFGS-B', jac=True,
+                               options=optdict, callback=callback_func)
+        elif grads:
             res = opt.minimize(self.objfunc, self._init_vec, method='L-BFGS-B', jac=self.objgrad,
                                options=optdict, callback=callback_func)
         else:
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -429,12 +429,13 @@ def transform_imarr(imarr, transforms, which_solve):
     else:
         raise Exception("transform_imarr requires imarr.shape[0] be either 1, 3, 4, or 10!")
 
-    # TODO(jax): the in-place assignments below (outarr[0] = ...) block tracing;
-    # rewrite functionally for the jax objective (Stokes-I just needs xp.exp).
+    xp = array_namespace(imarr)
+    if nimage==1:  # single-pol Stokes I: functional so jax.grad/jit can trace it
+        return xp.exp(imarr) if ('log' in transforms) else imarr.copy()
+
+    # pol / mf (nimage 3, 4, 10): in-place numpy; made functional in Phase 4.
     outarr = imarr.copy()
-    if nimage==1 and ('log' in transforms):
-        outarr = np.exp(outarr)
-    elif nimage==3 and ('log' in transforms):
+    if nimage==3 and ('log' in transforms):
         outarr[0] = np.exp(outarr[0])
     else:
 
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -1,6 +1,7 @@
 # imager_backend.py
 # Pure functional backend for imager.py
 
+import warnings
 from collections.abc import Sequence
 from typing import NamedTuple
 
@@ -1941,35 +1942,43 @@ def compute_objective_grad(imvec, initvec, config,
 def make_objective_jax(initvec, config, which_solve, data_tuples, logfreqratio_list,
                        n_obs, dat_term, reg_term, priorvec, norm_reg, reg_params,
                        embed_mask, device=None):
-    """Return a scipy fun(x) -> (value, grad) for the imaging objective.
+    """Return a scipy fun(x) -> (value, grad) for the imaging objective, via jax.
 
     Same arguments as compute_objective minus the solver vector x. The value is
-    compute_objective and the gradient is its jax autodiff, which equals
-    compute_objective_grad. Mode-agnostic: it just wraps compute_objective, so it
-    works for any pol mode / ttype the backend supports. jax is imported lazily so
-    `import ehtim` stays jax-free.
-
-    The only traced argument is x; everything else is captured in the closure, so
-    the unhashable data_tuples / dat_term dicts never reach jit and one trace is
-    reused across solver steps. Numeric data is moved onto `device` once.
+    compute_objective; the gradient is its jax autodiff, which matches the
+    hand-written compute_objective_grad. Works for any pol mode / ttype the
+    backend supports. jax is imported lazily so `import ehtim` stays jax-free.
     """
     import jax
     import jax.numpy as jnp
 
+    # Double precision is essential: in float32 the gradient is only ~1e-3
+    # accurate (and the nfft tolerance clamps), so warn rather than silently
+    # return a bad gradient.
+    if not jax.config.read("jax_enable_x64"):
+        warnings.warn("jax x64 is disabled -- the objective runs in float32 and "
+                      "gradients will be inaccurate. Set "
+                      "jax.config.update('jax_enable_x64', True) before imaging.",
+                      stacklevel=2)
+
+    # Move arrays onto the chosen device (a GPU if `device` is set) once, up front.
     def put(a):
         a = jnp.asarray(a)
         return jax.device_put(a, device) if device is not None else a
 
-    # A non-array A (the nfft NFFTInfo) is left as a host object for the kernel.
+    # Each data tuple may hold a non-array entry (the nfft NFFTInfo); leave it be.
     data_d = {key: tuple(put(v) if hasattr(v, "shape") else v for v in val)
               for key, val in data_tuples.items()}
     prior_d, init_d = put(priorvec), put(initvec)
 
+    # x is the only traced input; capturing everything else lets jit compile once
+    # and reuse that compilation across every L-BFGS-B step.
     def loss(x):
         return compute_objective(x, init_d, config, which_solve, data_d,
                                  logfreqratio_list, n_obs, dat_term, reg_term,
                                  prior_d, norm_reg, reg_params, embed_mask)
 
+    # jit the value-and-grad so scipy gets both from one compiled call.
     value_and_grad = jax.jit(jax.value_and_grad(loss))
 
     def fun(x):
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -705,9 +705,9 @@ def validate_params(prior, init, config, dat_term_keys, reg_term_keys, freq_list
         raise Exception("Initial image polrep is 'circ': pol_next must be 'RR' or 'LL'")
 
     if (prior.polrep == 'stokes'
-        and pol not in ['I', 'Q', 'U', 'V', 'P','IP','IQU','IV','IQUV']):
+        and pol not in ['I', 'Q', 'U', 'V', 'P','IP','IQU','IV','IQUV','IPV']):
         raise Exception(
-            "Initial image polrep is 'stokes': pol_next must be in 'I', 'Q', 'U', 'V', 'P','IP','IQU','IV','IQUV'!")
+            "Initial image polrep is 'stokes': pol_next must be in 'I', 'Q', 'U', 'V', 'P','IP','IQU','IV','IQUV','IPV'!")
 
     if ('log' in transforms and pol in ['Q', 'U', 'V']):
         raise Exception("Cannot image Stokes Q, U, V with log image transformation!")
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -10,6 +10,7 @@ import ehtim.imaging.imager_utils as imutils
 import ehtim.imaging.multifreq_imager_utils as mfutils
 import ehtim.imaging.pol_imager_utils as polutils
 import ehtim.observing.obs_helpers as obsh
+from ehtim.backends import array_namespace
 
 # -----------------------------------------------------------------------------
 # Naming convention for image arguments throughout this module
@@ -369,6 +370,7 @@ def unpack_imarr(vec, init_arr, which_solve):
       - if which_solve[k] == 0, fall back to `init_arr[k]` (the *initial*
         image, NOT a regularizer prior).
     """
+    xp = array_namespace(vec)
 
     imarrdim = len(init_arr.shape)
     if imarrdim==2:
@@ -384,17 +386,20 @@ def unpack_imarr(vec, init_arr, which_solve):
     if nsolve != len(which_solve):
         raise Exception("in unpack_imarr, init_arr has inconsistent shape with which_solve!")
 
+    # Build the rows then stack, rather than np.empty + in-place assignment, so
+    # jax.grad/jit can trace through: solved slots take the next nimage values of
+    # `vec`, not-solved slots fall back to the initial image. On numpy this is
+    # identical to the old buffer-fill; on jax the host init_arr rows promote to
+    # constants alongside the traced vec slices.
     imct = 0
-    # TODO(jax): build this functionally (e.g. xp.stack of the rows) so
-    # jax.grad(compute_objective) can trace it -- np.empty + in-place assignment
-    # is not jax-traceable.
-    imarr = np.empty((nsolve, nimage))
+    rows = []
     for kk in range(nsolve):
         if which_solve[kk]==0:
-            imarr[kk] = init_arr[kk]
+            rows.append(init_arr[kk])
         else:
-            imarr[kk] = vec[imct*nimage:(imct+1)*nimage]
+            rows.append(vec[imct*nimage:(imct+1)*nimage])
             imct += 1
+    imarr = xp.stack(rows)
 
     if imarrdim==1:
         imarr = imarr[0]
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -433,23 +433,28 @@ def transform_imarr(imarr, transforms, which_solve):
     if nimage==1:  # single-pol Stokes I: functional so jax.grad/jit can trace it
         return xp.exp(imarr) if ('log' in transforms) else imarr.copy()
 
-    # pol / mf (nimage 3, 4, 10): in-place numpy; made functional in Phase 4.
-    outarr = imarr.copy()
-    if nimage==3 and ('log' in transforms):
-        outarr[0] = np.exp(outarr[0])
+    # pol / mf (nimage 3, 4, 10): functional (stack/concatenate) so jax can trace.
+    if nimage==3:  # multifrequency Stokes I: log on I only
+        if 'log' in transforms:
+            return xp.stack([xp.exp(imarr[0]), imarr[1], imarr[2]])
+        return imarr.copy()
+
+    # Build the 4 Stokes rows (log on I, then polcv/mcv/vcv), keeping any trailing
+    # multifrequency rows (4:nimage) unchanged.
+    row0 = xp.exp(imarr[0]) if (pol_which_solve[0]==1 and ('log' in transforms)) else imarr[0]
+    pol_in = xp.stack([row0, imarr[1], imarr[2], imarr[3]])
+    if (pol_which_solve[1]==1 and pol_which_solve[3]==1 and ('polcv' in transforms)):
+        pol_out = polutils.polcv(pol_in)
+    elif (pol_which_solve[1]==1) and ('mcv' in transforms):
+        pol_out = polutils.mcv(pol_in)
+    elif (pol_which_solve[3]==1) and ('vcv' in transforms):
+        pol_out = polutils.vcv(pol_in)
     else:
+        pol_out = pol_in
 
-        if pol_which_solve[0]==1 and ('log' in transforms):  # full polarization, including stokes I imaging
-            outarr[0] = np.exp(outarr[0])
-
-        if (pol_which_solve[1]==1 and pol_which_solve[3]==1 and ('polcv' in transforms)):
-            outarr[0:4] = polutils.polcv(outarr)
-        elif (pol_which_solve[1]==1) and ('mcv' in transforms):
-            outarr[0:4] = polutils.mcv(outarr)
-        elif (pol_which_solve[3]==1) and ('vcv' in transforms):
-            outarr[0:4] = polutils.vcv(outarr)
-
-    return outarr
+    if nimage==4:
+        return pol_out
+    return xp.concatenate([pol_out, imarr[4:nimage]])
 
 def transform_imarr_inverse(imarr, transforms, which_solve):
     """Apply inverse transformation from physical to solver values for all polarizations"""
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -1931,3 +1931,44 @@ def compute_objective_grad(imvec, initvec, config,
     grad = datterm + regterm
     grad = transform_gradients(grad, imcur_prime, transforms, which_solve)
     return pack_imarr(grad, which_solve)
+
+
+def make_objective_jax(initvec, config, which_solve, data_tuples, logfreqratio_list,
+                       n_obs, dat_term, reg_term, priorvec, norm_reg, reg_params,
+                       embed_mask, device=None):
+    """Return a scipy fun(x) -> (value, grad) for the imaging objective.
+
+    Same arguments as compute_objective minus the solver vector x. The value is
+    compute_objective and the gradient is its jax autodiff, which equals
+    compute_objective_grad. Mode-agnostic: it just wraps compute_objective, so it
+    works for any pol mode / ttype the backend supports. jax is imported lazily so
+    `import ehtim` stays jax-free.
+
+    The only traced argument is x; everything else is captured in the closure, so
+    the unhashable data_tuples / dat_term dicts never reach jit and one trace is
+    reused across solver steps. Numeric data is moved onto `device` once.
+    """
+    import jax
+    import jax.numpy as jnp
+
+    def put(a):
+        a = jnp.asarray(a)
+        return jax.device_put(a, device) if device is not None else a
+
+    # A non-array A (the nfft NFFTInfo) is left as a host object for the kernel.
+    data_d = {key: tuple(put(v) if hasattr(v, "shape") else v for v in val)
+              for key, val in data_tuples.items()}
+    prior_d, init_d = put(priorvec), put(initvec)
+
+    def loss(x):
+        return compute_objective(x, init_d, config, which_solve, data_d,
+                                 logfreqratio_list, n_obs, dat_term, reg_term,
+                                 prior_d, norm_reg, reg_params, embed_mask)
+
+    value_and_grad = jax.jit(jax.value_and_grad(loss))
+
+    def fun(x):
+        value, grad = value_and_grad(jnp.asarray(x))
+        return float(value), np.asarray(grad, dtype=np.float64)
+
+    return fun
--- a/ehtim/imaging/imager_utils.py
+++ b/ehtim/imaging/imager_utils.py
@@ -26,6 +26,7 @@ import scipy.sparse as sps
 import ehtim.const_def as ehc
 import ehtim.observing.obs_helpers as obsh
 from ehtim.backends import array_namespace
+from ehtim.observing.obs_helpers import nufft2_backend
 
 ##################################################################################################
 # Constants & Definitions
@@ -887,26 +888,6 @@ def chisqgrad_logamp_fft(vis_arr, A, amp, sigma):
 ##################################################################################################
 # NFFT Chi-squared and Gradient Functions
 ##################################################################################################
-
-
-def nufft2_backend(imvec, nfft_info):
-    """Type-2 NUFFT: uniform image -> nonuniform visibility samples.
-
-    Dispatches on imvec's array type -- the stateful finufft plan on numpy
-    (byte-identical to the legacy path), jax_finufft.nufft2 on jax (differentiable,
-    GPU-capable). Conventions match: isign=-1, (-pi, pi] points, (xdim, ydim) modes.
-    Callers multiply the result by nfft_info.pulsefac.
-    """
-    xp = array_namespace(imvec)
-    f_hat = imvec.reshape((nfft_info.ydim, nfft_info.xdim)).T
-    if xp is np:
-        nfft_info.plan.f_hat = f_hat
-        nfft_info.plan.trafo()
-        return nfft_info.plan.f.copy()
-    from jax_finufft import nufft2
-    uvf = nfft_info.uv_finufft
-    return nufft2(f_hat.astype(xp.complex128), uvf[:, 0], uvf[:, 1],
-                  iflag=-1, eps=nfft_info.eps)
 
 
 def chisq_vis_nfft(imvec, A, vis, sigma):
@@ -3372,8 +3353,9 @@ def embed(imvec, mask, clipfloor=0., randomfloor=False):
     """Embeds a 1d image vector into the size of boolean embed mask
     """
     xp = array_namespace(imvec)
-    on = mask.nonzero()[0]
-    off = (mask - 1).nonzero()[0]
+    mbool = mask.astype(bool)
+    on = mbool.nonzero()[0]
+    off = (~mbool).nonzero()[0]
     if randomfloor:  # prevent total variation gradient singularities
         floor = clipfloor * np.abs(np.random.normal(size=len(off)))
     else:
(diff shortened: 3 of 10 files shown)


ehtim/imaging/imager_utils.py
```diff
@@ -889,24 +889,31 @@ def chisqgrad_logamp_fft(vis_arr, A, amp, sigma):
...
+    GPU-capable). Conventions match: isign=-1, (-pi, pi] points, (xdim, ydim) modes.
+    Callers multiply the result by nfft_info.pulsefac.
+    """
+    xp = array_namespace(imvec)
+    f_hat = imvec.reshape((nfft_info.ydim, nfft_info.xdim)).T
+    if xp is np:
+        nfft_info.plan.f_hat = f_hat
+        nfft_info.plan.trafo()
+        return nfft_info.plan.f.copy()
+    from jax_finufft import nufft2
```
achael: minor: i'd prefer an explicit else statement here

ehtim/imaging/imager_utils.py
```diff
@@ -1851,7 +1728,7 @@ def reg_cm(imvec, mask, **kwargs):
     xx, yy = np.meshgrid(range(nx//2, -nx//2, -1), range(ny//2, -ny//2, -1))
```
achael, line 1738: is it ok to have a bare np when using  jax here for meshgrid?

ehtim/imaging/imager_utils.py
```diff
@@ -1924,7 +1803,7 @@ def reg_tvlog(imvec, mask, **kwargs):
     logflux = npix * np.abs(np.log(flux / npix))
```
achael, line 1832: ok to have np.log here?

achael: i guess yes since flux/npix is a float?

ehtim/imaging/imager_utils.py
```diff
@@ -1991,7 +1872,7 @@ def reg_tv2log(imvec, mask, **kwargs):
...
```
achael, line 1920: same as above

ehtim/imaging/imager_utils.py
```diff
@@ -2021,9 +1903,9 @@ def reg_compact(imvec, mask, **kwargs):
     xx, yy = np.meshgrid(range(nx), range(ny))
```
achael, line 1959: im just flagging all the bare np. uses -- i think the are probably all fine but want to be sure

ehtim/imaging/imager_utils.py
```diff
@@ -3487,20 +3371,23 @@ def plot_i(im, Prior, nit, chi2_dict, **kwargs):
...
     else:
-        out[(mask-1).nonzero()] = clipfloor
+        floor = np.full(len(off), clipfloor)
 
-    return out
+    if xp is np:
+        out = np.zeros(len(mask))
+        out[on] = imvec
+        out[off] = floor
+        return out
```
achael, line 3450: again, unless there is a good reason not to, i think an explicit else statement would help with readability

ehtim/imaging/imager_utils.py
```diff
@@ -3558,17 +3445,24 @@ def embed_imarr(imarr, mask, clipfloor=0., randomfloor=False):
...
     nimage_out = len(mask)
-    outarr = np.empty((nsolve, nimage_out))
-    # Vectorized over the nsolve axis: scatter imarr into the masked columns
-    # of outarr, then fill non-mask columns with clipfloor (or random).
-    not_mask = ~mask.astype(bool)
-    outarr[:, mask.astype(bool)] = imarr
+    # Vectorized over the nsolve axis: scatter imarr into the masked columns,
+    # then fill non-mask columns with clipfloor (or random). Functional on jax.
+    xp = array_namespace(imarr)
+    on_cols = mask.astype(bool).nonzero()[0]
```
achael: any reason the syntax for on/off is different in embed vs embed_imarr? (mask.nonzero and (mask-1).nonzero vs mask.astype(bool).nonzero , ~mask.astype(bool).nonzero)

ehtim/imaging/imager_utils.py
```diff
@@ -889,24 +889,31 @@ def chisqgrad_logamp_fft(vis_arr, A, amp, sigma):
 ##################################################################################################
```
achael, line 896: consider moving nufft2_backend to imager_backend.py

achael: or obs_helpers.py -- since there it would be next to the rest of the NFFT machinery

ehtim/observing/obs_helpers.py
```diff
@@ -1466,6 +1466,8 @@ def __init__(self, xdim, ydim, psize, pulse, npad, p_rad, uv,
         # (-pi, pi] form is what FINUFFTPlan expects.
         uv_scaled = uv * psize
         uv_finufft = 2 * np.pi * uv_scaled
+        self.eps = eps
+        self.uv_finufft = uv_finufft  # (-pi, pi] nonuniform points, for jax_finufft.nufft2
         self.plan = FINUFFTPlan(self.xdim, self.ydim, uv_finufft, eps=eps)
```
achael, line 1499: maybe put a comment that plan isn't used for jax

achael: moving nufft_backend here would help make the relationship clear

ehtim/imager.py
```diff
@@ -508,7 +512,19 @@ def callback_func(xcur):
 
 
         tstart = time.time()
-        if grads:
+        if kwargs.get('use_jax', False):
```
achael, line 516: it's very nice that it turns out to be so simple in imager.py

ehtim/imager.py
```diff
@@ -439,6 +440,9 @@ def make_image(self, pol=None, grads=True, mf=False, **kwargs):
            Args:
                pol (str): which polarization to image
                grads (bool): whether or not to use image gradients
+               use_jax (bool): use the jax autodiff objective (GPU-capable) in place
+                   of the numpy analytic gradient; the objective is identical
+               jax_device: jax device for the objective when use_jax=True (None = jax default)
```
achael: maybe put examples of what jax_device might be if not None?

ehtim/imaging/pol_imager_utils.py
```diff
@@ -68,15 +70,17 @@ def make_p_image(imarr):
     """
 
     # NOTE! We replaced EVPA chi with phi=2chi in the imarr
```
achael: can probably get rid of this NOTE -- best would be to add a comment under #CONVENTIONS at the top of the file re what imarr[0], imarr[1], imarr[2], imarr[3] are

ehtim/imaging/pol_imager_utils.py
```diff
@@ -193,18 +202,19 @@ def mcv(imarr):
        input is solver values, output is physical values
     """
 
+    xp = array_namespace(imarr)
     vfrac = imarr[3] # when using this transform, we interpret transformed imarr[3] as mfrac=\rho sin(\psi)
-    mfrac_max = 1-np.abs(vfrac)
-    if np.any(mfrac_max>1):
+    mfrac_max = 1-xp.abs(vfrac)
+    if xp is np and np.any(mfrac_max>1):  # dead guard (mfrac_max<=1 always); kept on numpy
```
achael: let's just delete this check for both numpy and jax unless there is a good reason not to

ehtim/imaging/pol_imager_utils.py
```diff
@@ -266,18 +276,19 @@ def vcv(imarr):
     """change of variables for v(v') from range (-inf,inf) to (-1+|m|,1-|m|) while keeping m=P/I fixed
        input is solver values, output is physical values"""
 
+    xp = array_namespace(imarr)
     mfrac = imarr[1] # when using this transform, we interpret transformed imarr[1] as mfrac=\rho cos(\psi)
-    vfrac_max = 1-np.abs(mfrac)
-    if np.any(vfrac_max>1):
+    vfrac_max = 1-xp.abs(mfrac)
+    if xp is np and np.any(vfrac_max>1):  # dead guard (vfrac_max<=1 always); kept on numpy
```
achael: same as above

ehtim/imaging/pol_imager_utils.py
```diff
@@ -266,18 +276,19 @@ def vcv(imarr):
...
-    rho = np.sqrt(mfrac**2 + vfrac**2)
-    psi = np.arcsin(vfrac/rho)
+    rho = xp.sqrt(mfrac**2 + vfrac**2)
+    psi = xp.arcsin(vfrac/rho)
 
-    out = np.array((imarr[0], rho, imarr[2], psi))
+    out = xp.stack((imarr[0], rho, imarr[2], psi))
     return out
 
 def vcv_r(imarr):
```
achael, line 312: why aren't vcv, vcv_r modified? are they not used in the current implementation?

ehtim/imaging/pol_imager_utils.py
```diff
@@ -428,9 +439,10 @@ def chisq_p(imarr, Amatrix, p, sigmap):
...
 
+    xp = array_namespace(imarr)
     pimage = make_p_image(imarr)
-    psamples = np.dot(Amatrix, pimage)
-    chisq =  np.sum(np.abs(p - psamples)**2/(sigmap**2)) / (2*len(p))
+    psamples = xp.dot(Amatrix, pimage)
+    chisq =  xp.sum(xp.abs(p - psamples)**2/(sigmap**2)) / (2*len(p))
     return chisq
 
 def chisqgrad_p(imarr, Amatrix, p, sigmap,pol_solve=POL_SOLVE_DEFAULT):
```
achael, line 478: I thought we weren't adding jax support to the gradient functions? I'm fine either way, but we should be consistent -- we didn't in the nfft ones in imager_utils.py

ehtim/imaging/pol_imager_utils.py
```diff
@@ -524,9 +537,10 @@ def chisq_vvis(imarr, Amatrix, v, sigmav):
     """V visibility chi-squared
     """
 
+    xp = array_namespace(imarr)
     vimage = make_v_image(imarr)
-    vsamples = np.dot(Amatrix, vimage)
-    chisq =  np.sum(np.abs(v - vsamples)**2/(sigmav**2)) / (2*len(v))
+    vsamples = xp.dot(Amatrix, vimage)
+    chisq =  xp.sum(xp.abs(v - vsamples)**2/(sigmav**2)) / (2*len(v))
     return chisq
```
achael, line 580: what about all the other  polchisq functions / nfft versions?

ehtim/imaging/pol_imager_utils.py
```diff
@@ -887,24 +907,38 @@ def reggrad_ptv(imarr, mask, **kwargs):
...
-    d2 = np.sqrt(np.abs(im_r1 - im)**2 + np.abs(im_r1l2 - im_r1)**2)
-    d3 = np.sqrt(np.abs(im_r2 - im)**2 + np.abs(im_l1r2 - im_r2)**2)
+    # epsilon_tv (default 0, matching reg_ptv) keeps these finite at zero-|P| diffs.
+    epsilon = kwargs.get('epsilon_tv', 0.)
+    d1 = np.sqrt(np.abs(im_l1 - im)**2 + np.abs(im_l2 - im)**2 + epsilon)
+    d2 = np.sqrt(np.abs(im_r1 - im)**2 + np.abs(im_r1l2 - im_r1)**2 + epsilon)
+    d3 = np.sqrt(np.abs(im_r2 - im)**2 + np.abs(im_l1r2 - im_r2)**2 + epsilon)
     # Numerators below use cos/sin of the single-angle difference between
     # neighbors, from d|P_l1 - P|^2/d|P| = 2|P| - 2|P_l1|*cos(angle(P_l1) - angle(P)).
+    # The back-neighbor magnitude numerators m2/m3 keep a |P| term that does not
```
achael: make this a one-line comment (e.g.     # mask the first row column gradient terms that don't exist)

ehtim/imaging/imager_backend.py
```diff
@@ -1925,3 +1936,44 @@ def compute_objective_grad(imvec, initvec, config,
...
+                       embed_mask, device=None):
+    """Return a scipy fun(x) -> (value, grad) for the imaging objective.
+
+    Same arguments as compute_objective minus the solver vector x. The value is
+    compute_objective and the gradient is its jax autodiff, which equals
+    compute_objective_grad. Mode-agnostic: it just wraps compute_objective, so it
+    works for any pol mode / ttype the backend supports. jax is imported lazily so
+    `import ehtim` stays jax-free.
+
+    The only traced argument is x; everything else is captured in the closure, so
```
achael: this last part of the docstring is a bit jargony and i don't think necessary -- i would remove or rephrase

ehtim/imaging/imager_backend.py
```diff
@@ -1925,3 +1936,44 @@ def compute_objective_grad(imvec, initvec, config,
...
+    reused across solver steps. Numeric data is moved onto `device` once.
+    """
+    import jax
+    import jax.numpy as jnp
+
+    def put(a):
+        a = jnp.asarray(a)
+        return jax.device_put(a, device) if device is not None else a
+
+    # A non-array A (the nfft NFFTInfo) is left as a host object for the kernel.
```
achael: this comment in particular is a bit confusing

ehtim/imaging/imager_backend.py
achael: and i'd add some more one-line explanatory comments throughout for non jax experts to be able to follow what's going on here

achael (commented):
awesome stuff, I'm amazed this has been so simple so far! lots of comments but they boil down to 3 points. 

1. a few places where there seem to be calls to np. where i'm wondering if you want xp. for the flexible backend. But not being fluent in jax i'm not sure if it's necessary
2. coverage of jaxifying the chisqs and gradient functions -- in pol_imager_utils.py in particular, it doesn't seem like all chisq functions are converted, while on the flip side some gradient functions are converted, when gradients aren't touched 
3. reworking the comments to be clear (some things deserve more comments so those not familiar with jax can follow) but a bit more terse (in bug fixes the comments don't need to signpost that there used to be a bug there).

ehtim/imaging/imager_utils.py
rohandahale: `reg_cm` meshgrid / `reg_tvlog` & `reg_tv2log` `np.log` / `reg_compact`: These are all constants as far as the solver's concerned (coordinate grids, or  a scalar float). So they never get traced. When they multiply the traced image, jax folds them in as constants on its own, so `xp.` isn't needed (and for the meshgrid it'd just force a device array we don't want). The actual log of the image already uses `xp.log`. I left these as `np.` on purpose.

ehtim/imaging/imager_utils.py
rohandahale: Yes, same reply as above

ehtim/imaging/imager_utils.py
rohandahale: Same reply as above

ehtim/imaging/imager_utils.py
rohandahale: No worries, I addressed your concern in the top inline comment

ehtim/imaging/imager_utils.py
rohandahale: Done, explicit `else`

ehtim/imaging/imager_utils.py
rohandahale: Nope, they were just doing the same thing two different ways. Unified both to `mbool = mask.astype(bool)` then `mbool.nonzero()` / `(~mbool).nonzero()`.

ehtim/imaging/imager_utils.py
rohandahale: Moved it into `obs_helpers.py`, right after `NFFTInfo` and next to `FINUFFTPlan`. `imager_utils` and `pol_imager_utils` import it from there now.

ehtim/observing/obs_helpers.py
rohandahale: Added it, and moved `nufft2_backend` here too so it's clear the finufft `plan` is the numpy-only path while jax goes through `uv_finufft`.

ehtim/imager.py
rohandahale: Added: `jax.devices('gpu')[0]` or `jax.devices('cpu')[0]`.

ehtim/imaging/pol_imager_utils.py
rohandahale: Dropped it. The `# CONVENTIONS` block at the top of the file already has `imarr = (I, ρ, φ, ψ)` with `φ = 2χ`, so the inline NOTE was just repeating that

ehtim/imaging/pol_imager_utils.py
rohandahale: Deleted in both. `mfrac_max = 1 - |vfrac|` (and the vcv equivalent) is always ≤ 1, so the check could never fire (it was dead code either way).

ehtim/imaging/pol_imager_utils.py
rohandahale: Deleted in both. `mfrac_max = 1 - |vfrac|` (and the vcv equivalent) is always ≤ 1, so the check could never fire (it was dead code either way).

ehtim/imaging/pol_imager_utils.py
rohandahale: `vcv` is converted (it's agnostic now). `vcv_r` is the reverse transform (physical --> solver) and only runs once at setup. `imager_backend.py:497` uses  it to build the solver vector from the physical prior. It never touches the traced path, so there's nothing to gain from jax-ifying it. Same for `mcv_r`.

ehtim/imaging/pol_imager_utils.py
rohandahale: This is actually `chisq_p` (the value function). None of the gradient functions are converted

ehtim/imaging/pol_imager_utils.py
rohandahale: All converted now: `chisq_p_nfft`, `chisq_m_nfft`, and `chisq_vvis_nfft` go through `nufft2_backend`. So it's consistent: every value chisq is backend-agnostic, every gradient chisq stays numpy.

ehtim/imaging/pol_imager_utils.py
rohandahale: Done

ehtim/imaging/imager_backend.py
rohandahale: Cleaned up, dropped the dense paragraphs, reworded the confusing NFFTInfo line, and added short comments through the body

ehtim/imaging/imager_backend.py
rohandahale: fixed it

rohandahale:
Thanks @achael  this was really useful! 

**1. np. vs xp.** Every bare `np.` you flagged is on something that doesn't depend on the image we're solving for — coordinate grids from `meshgrid`, or  scalar arithmetic like `np.log(flux/npix)` where `flux/npix` is just a Python float. None of that gets traced. And when one of these constant numpy arrays multiplies the traced image (e.g. `imvec*xx`), jax folds it in as a constant on  its own, so switching to `xp.` wouldn't get anything — for the meshgrid it'd actually be worse, since `xp.meshgrid` would force a device array where a baked-in constant is exactly what we want. The one spot where it matters, the log of the *image* itself, already uses `xp.log(imvec)`. So the bare `np.`s are all intentional.

**2. chisq vs grad conversion.** Good catch, it really was inconsistent. The rule I'm following is: the **value** chisqs are backend-agnostic (jax autodiffs them), and the **gradient** functions stay pure numpy (under jax, autodiff replaces them entirely, so they're never traced). What you spotted was the pol *nfft* values — `chisq_p_nfft` / `chisq_m_nfft` / `chisq_vvis_nfft` were still driving the finufft plan directly, which jax can't trace through. They now go through `nufft2_backend` like the Stokes-I nfft kernels and the non-nfft pol ones. That path actually didn't work under jax at all before this, so I added `test_pol_nfft_*[IP,IV], they pass on the GPU.

**3.** Trimmed the comments, cut the jargon out of the `make_objective_jax` docstring, and added plain one-liners.

ehtim/imaging/imager_utils.py
rohandahale: Done

ehtim/imaging/pol_imager_utils.py
achael: i see, i was misreading what function these lines were in in the github diff. Makes sense; we should just keep in mind this asymmetric jaxification of the _r functions in case they ever did end up being used in an optimization. .

ehtim/imaging/pol_imager_utils.py
achael: ah ok, again my fault on misreading the git diff page.

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+VALUE_RTOL = 1e-9
+GRAD_RTOL = 1e-9
+# nfft is an approximate transform: jax_finufft's VJP matches the analytic adjoint
+# only to the NUFFT accuracy. The residual scales with eps (verified 1e-9 -> ~1e-6,
+# 1e-12 -> ~7e-10 max), so the nfft fixture uses eps=1e-12 and this looser-but-tight
+# rtol; a real VJP bug is eps-independent and fails by orders of magnitude.
+NFFT_EPS = 1e-12
+NFFT_GRAD_RTOL = 1e-7
+GRAD_ATOL = 1e-12
+# Pol grad tests validate jax.grad against central FD (the ground truth), NOT the
```
achael: i think it would be good to add in the tests against the analytic gradients and xfail the failing vvis one until we patch in the recent fix. 

Isn't the gradient singularity for vtv/ptv also present in normal tv when epsilon_tv=0, as it is by default? I don't see why that should prevent us from testing with finite epsilon.

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+    assert np.allclose(g_jax, iu.reggrad_tv(imvec, mask, **kw), rtol=1e-8, atol=1e-10)
+
+
+@pytest.fixture(scope="module")
+def pol_imager(obs_pol_direct, gauss_im_pol, gauss_prior):
+    # IP imaging: Stokes I + linear pol (pvis, m); mcv change-of-variables.
+    # Stokes-I reg only here; pol regs are exercised below.
+    imgr = eh.imager.Imager(
+        obs_pol_direct, gauss_prior, prior_im=gauss_prior, flux=gauss_im_pol.total_flux(),
+        data_term={"amp": 100, "pvis": 100, "m": 50},
```
achael: pull data_term and reg_term out to top as in stokes I imager

achael: and make sure the test exercises all of the pol regularizers and chisqs

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+    assert np.allclose(float(pol_imager_nfft.objfunc(pol_nfft_x0)),
+                       float(pol_imager_nfft.objfunc(jnp.asarray(pol_nfft_x0))),
+                       rtol=VALUE_RTOL, atol=GRAD_ATOL)
+
+
+def test_pol_nfft_grad_finite_difference(pol_imager_nfft, pol_nfft_x0):
+    _assert_pol_grad_matches_fd(pol_imager_nfft, pol_nfft_x0)
+
+
+def _nxcorr(a, b):
```
achael, line 356: set off the end to end test with a section header comment

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+    mfarr = np.array([np.full(8, 2.0), np.full(8, -0.7), np.full(8, 0.1)])  # I0, alpha, beta
+    lf = 0.3
+    out_np = mfu.image_at_freq(mfarr, lf)
+    out_jx = np.asarray(mfu.image_at_freq(jnp.asarray(mfarr), lf))
+    assert np.allclose(out_np, out_jx, rtol=VALUE_RTOL, atol=GRAD_ATOL)
+    g = jax.grad(lambda a: jnp.sum(mfu.image_at_freq(a, lf)))(jnp.asarray(mfarr))
+    assert np.all(np.isfinite(np.asarray(g)))
+
+
+def test_ptv_epsilon_tv_removes_singularity():
```
achael: i dont think this should be in the jax-objective test; probably in test-regularizers instead?

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+# analytic compute_objective_grad: the V-pol analytic gradient is wrong (jax and FD
+# agree, the analytic disagrees -- tracked bug), and ptv/vtv have a sqrt singularity.
+FD_STEP = 1e-6
+FD_MEDIAN_TOL = 1e-3
+FD_MAX_TOL = 1e-2
+N_FD_SAMPLES = 40
+RNG_SEED = 4
+PERTURB = 0.10
+
+DATA_TERM = {"amp": 100, "cphase": 100, "logcamp": 50}
```
achael: this doesn't exercise all of the data terms and regularizer terms -- it would be good to loop over different data_term and reg_term combos that in the union cover all of the individual data_term and reg_term functions.

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+    full = rng.uniform(0.1, 1.0, ny * nx)
+    mask = np.ones(ny * nx, dtype=bool)
+    mask[rng.choice(ny * nx, 8, replace=False)] = False
+    imvec = full[mask]
+    kw = dict(xdim=nx, ydim=ny, psize=1.0, flux=float(full.sum()), beam_size=2.0, norm_reg=True)
+    assert np.allclose(iu.reg_tv(imvec, mask, **kw),
+                       float(iu.reg_tv(jnp.asarray(imvec), mask, **kw)), rtol=1e-12)
+    g_jax = np.asarray(jax.grad(lambda v: iu.reg_tv(v, mask, **kw))(jnp.asarray(imvec)))
+    assert np.allclose(g_jax, iu.reggrad_tv(imvec, mask, **kw), rtol=1e-8, atol=1e-10)
+
```
achael, line 209: set off the polarization section by a block comment
make sure polarization tests have parallel structure to the stokes I tests.

tests/test_objective_jax.py
```diff
@@ -0,0 +1,340 @@
...
+    return base + PERTURB * rng.standard_normal(base.size)
+
+
+def test_pol_v_grad_finite_difference(pol_imager_v, pol_v_x0):
+    _assert_pol_grad_matches_fd(pol_imager_v, pol_v_x0)
+
+
+# nfft pol: routes chisq_{p,m,vvis}_nfft through nufft2_backend (jax_finufft).
+# IP exercises chisq_p_nfft (pvis) + chisq_m_nfft (m); IV exercises chisq_vvis_nfft.
+_POL_NFFT_CASES = [
```
achael: add and IPV test to cover full-stokes imaging and the polcv transform.

tests/test_regularizers.py
```diff
@@ -298,6 +298,46 @@ def test_max_frac_diff(self, polreg_setup, rtype):
             f"{rtype} max fractional gradient diff = {max_frac:.6f} (tol={max_tol})"
         )
 
+    def test_reggrad_ptv_matches_fd_on_boundary(self):
```
achael: adapt this test to run over all tv functions; tv, tv2, poltv, vtv?

achael (commented):
looks great! I misread some of the changes in the git diff before which lead to the wrong comments about function coverage. 

My only remaining comments are on test_objective_jax now. I don't think the current version covers all Stokes I chisq/regularizer functions, and the structure is not parallel between stokes I and polarization. It might be useful just to have Claude rewrite this whole module now that the jax functions are fixed, since probably some of the current non-parallel structure came together as the code was being developed. 

 Here is what I think the full jax test module should contain: 
 
- coverage of  ['direct','nfft'] transforms -- already done
- coverage of different pairs of data_terms and reg_terms that, taken together, exercise all of the regularizers and chisq functions
- coverage of different imaging modes - ['I','IP','P','V','IV','IPV'] that together exercise the different transform functions ['polcv','mcv','vcv']
- uniform comparison to analytic and finite diff gradients in all cases. We should xfail the one fixed by the pending bugfix PR, and use epsilon_tv != uniformly between pol and stokes I gradients to avoid singularities
- similar coverage for standard Stokes I multifrequency chisqs, regularizers -- with a note that we need to add real multifreq synthetic data to the test and that multifreq polarization coverage needs to be added. 

Finally, I think there should also be a simple test on equality between the numpy/jax branches of nufft2_backend (this is already exercised by the chisqs, but I think that its useful to have a dedicated test).

achael:
I merged the IV gradient fix to dev-backend in #299 so in the new tests of the jax vs analytic pol gradients we shouldn't need to xfail anything

rohandahale:
Thanks @achael! I rewrote `test_objective_jax` along the lines you laid out. The broader coverage caught more gradient bugs in the same family as the IV one (#296).

Every case now checks the three things uniformly, value parity (numpy vs jax), analytic-gradient parity (jax.grad vs the hand-written objgrad), and finite differences. The parametrized cases together exercise:
- both transforms (direct + nfft),
- all the Stokes-I chisqs (vis/amp/bs/cphase/camp/logcamp) and regularizers,
- **all six pol modes** : IP/P (mcv), IV/V (vcv), IQUV/IPV (polcv)  and all the pol chisqs (pvis/m/vvis) and pol regs,
- a multifrequency Stokes-I case.

Two things made it actually catch bugs:
1. **An asymmetric image.** A centered, circular source zeros gradient components by symmetry, so the analytic and autodiff gradients can disagree and you'd never see it. Switching to an off-center, elongated source exposed the disagreements.
2. **Comparing analytic vs autodiff vs FD together**: three independent gradients

The bugs it found (all same as #296):
The mcv/vcv transforms have non-diagonal Jacobians (`mprime` couples to both rho and psi, `vprime` likewise) so the gradient has to give both physical-slot gradients that the chain rule consumes. Several places used the term on the wrong `pol_solve` index and dropped it (exactly like `chisqgrad_vvis` did for IV):

- **`chisqgrad_p` / `chisqgrad_m`** dropped grad_psi for `pol='IP'` --> wrong IP/P gradient 
- **V-pol regs** (`vflux`/`l1v`/`l2v`/`vtv`/`vtv2`) dropped grad_rho for IV/V.
- **linear-pol regs** (`msimple`/`hw`/`ptv`) dropped grad_psi for IP/P.

The fix in each is the mirror of #296: compute the needed term whenever the coupled variable is solved (`pol_solve[1] or pol_solve[3]`). These affect the main branch numpy pol imaging (IP especially), symmetric test images had been hiding them. They are in this PR, but we can move them to dev and main if you want.

- **NaN-safe mcv/vcv transforms.** `jax.grad` of the pol transform returned NaN at zero-polarization pixels (`sqrt(0)` / `arcsin(±1)` in the backward pass), so jax pol imaging would NaN on any realistic image. I used `where` to fix it. numpy values are unchanged.
- **`IPV`** was in `POLARIZATION_MODES` but the validator rejected it. It now works (identical to IQUV / polcv).

**test_regularizers**: I generalized the ptv boundary FD test over all the TV funcs (tv/tv2/ptv/vtv/vtv2), and added a parametrized epsilon_tv-removes-singularity test (tv/ptv/vtv). The `image_at_freq` test moved out of the jax module; it's covered there now by a proper multifrequency-imager case instead.

rohandahale:
@achael This PR is now ready to be merged! I resolved all conflicts after #306 fix but please do a quick check before merging.

ehtim/imaging/multifreq_imager_utils.py
```diff
@@ -240,16 +244,17 @@ def reggrad_l2_spec(imvec, mask, **kwargs):
...
     norm = len(imvec) * psize / beam_size if kwargs.get('norm_reg', True) else 1
     im = imvec.reshape(ny, nx)
-    impad = np.pad(im, 1, mode='constant', constant_values=0)
-    im_l1 = np.roll(impad, -1, axis=0)[1:ny+1, 1:nx+1]
-    im_l2 = np.roll(impad, -1, axis=1)[1:ny+1, 1:nx+1]
-    return np.sum(np.sqrt(np.abs(im_l1 - im)**2 + np.abs(im_l2 - im)**2 + EPSILON)) / norm
+    impad = xp.pad(im, 1, mode='constant', constant_values=0)
+    im_l1 = xp.roll(impad, -1, axis=0)[1:ny+1, 1:nx+1]
+    im_l2 = xp.roll(impad, -1, axis=1)[1:ny+1, 1:nx+1]
+    return xp.sum(xp.sqrt(xp.abs(im_l1 - im)**2 + xp.abs(im_l2 - im)**2 + EPSILON)) / norm
```
achael, line 257: just noticed this EPSILON is fixed -- should make it epsilon_tv as well I think, later.

achael (approved):
Thanks @rohandahale, looks good, ready to merge!

3 notes for the next PR: 
1. looks like EPSILON is a global constant in the spectral index TV -- we should change this to epsilon_tv to match the other functions. 
2. I think _rho_psi_safe should be promoted to a helper function like (rho_psi_from_mfrac_vfrac) or something -- it could have utililty outside the specific pol transforms here
3. We should consider reviewing the tests and consolidating the gradient tests in particular soon. They are scattered in different files and I think there is some redundancy & non-uniformity in how they test.

PR #297, opened 2026-06-10
Zero ptv gradient back-neighbor terms at the image boundary

rohandahale:
`reggrad_ptv` (the pol total-variation gradient) was missing the first-row /first-column boundary zeroing that `reggrad_tv` and `reggrad_vtv` both apply. As a result the **entire first row and first column** of the `|P|`-magnitude gradient slots (I, m, psi — `pol_solve` 0/1/3) were wrong; the phase slot (2) was correct.

## Root cause

`reg_ptv` zero-pads the image and sums forward differences. The gradient has a self term plus back-neighbor terms (`m2` from the up-neighbor, `m3` from the left-neighbor). At the top/left edge those back-neighbors are the zero-pad, i.e. not real pixels, so their contribution must be dropped — which is exactly what `reggrad_tv`/`reggrad_vtv` do with `mask1`/`mask2`.

The phase numerators `c2 = 2|P·P_neighbor|·sin(...)` carry a `|P_neighbor|` factor that is identically 0 at the pad, so they self-zero — which is why the phase slot was already correct. The magnitude numerators `m2 = |P| − |P_neighbor|·cos(...)` keep the `|P|` term, which does **not** vanish at the pad, so the edge picked up a spurious neighbor that doesn't exist.

## Fix

Define `mask1` (first row) / `mask2` (first col) and zero `m2`/`m3` there in the three magnitude blocks — the same pattern the sibling regularizers use.

## Validation

Central finite-difference of `reg_ptv` is the ground truth (it knows nothing about the analytic code). On a 6×6 full-mask grid with pol structure on every pixel:

| | corner (0,0), `m`-slot | max |analytic − FD| over all pixels |
|---|---|---|
| before | 0.140 (FD says 0.032) — **4× wrong** | 7.8e-1 (first row + col) |
| after | 0.032 = 0.032 | 1.2e-10 (every slot, incl. boundary) |

New regression test `test_reggrad_ptv_matches_fd_on_boundary` checks every pixel of a full grid (the existing random-pixel FD checks rarely sample the edge, which is why this stayed latent). It **fails on the pre-fix source, passes after**.

## Note

This overlaps `reggrad_ptv` with #295 (which adds the `epsilon_tv` guard to the same function); whichever lands second resolves a small same-function conflict. Surfaced by the jax autodiff-vs-analytic gradient comparison being built in #295.

### DIFF
--- a/ehtim/imaging/pol_imager_utils.py
+++ b/ehtim/imaging/pol_imager_utils.py
@@ -892,12 +892,22 @@ def reggrad_ptv(imarr, mask, **kwargs):
     d3 = np.sqrt(np.abs(im_r2 - im)**2 + np.abs(im_l1r2 - im_r2)**2)
     # Numerators below use cos/sin of the single-angle difference between
     # neighbors, from d|P_l1 - P|^2/d|P| = 2|P| - 2|P_l1|*cos(angle(P_l1) - angle(P)).
+    # The back-neighbor magnitude numerators m2/m3 keep a |P| term that does not
+    # vanish at the zero-pad, so the first row/col would pick up a neighbor that
+    # does not exist; zero those terms there as reggrad_tv/reggrad_vtv do. (The
+    # phase numerators c2/c3 carry a |P_neighbor| factor and self-zero there.)
+    mask1 = np.zeros(im.shape, dtype=bool)
+    mask2 = np.zeros(im.shape, dtype=bool)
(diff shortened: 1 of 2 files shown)


rohandahale:
Folding this into #295 instead (which already touches reggrad_ptv for epsilon_tv) — keeping it to one PR.

PR #304, opened 2026-06-13
Modular JAX optimizers + multi-GPU sharding

rohandahale:
Modular JAX optimizers + multi-GPU sharding for the imaging objective, on top of the JAX objective (#295). Both are opt-in; the default `make_image` path is unchanged (scipy L-BFGS-B, bit-for-bit).

## Optimizer layer

`make_image(optimizer=...)` now accepts:
- `None` (default) — scipy L-BFGS-B, exactly as before
- `'optax-lbfgs'` / `'optax-lbfgs-bt'` (backtracking line search) / `'adam'` / `'adamw'` / `'sgd'` / `'rmsprop'`, or any `optax.GradientTransformation` — an on-device `lax.while_loop` (value_and_grad stays on device, no per-step host sync)
- any callable — escape hatch, given the host value_and_grad

New `ehtim/imaging/optimizers.py`. `optax` is an optional, lazily-imported dependency (`import ehtim` stays optax-free).

## Multi-GPU sharding

`make_image(shard=True, shard_axis='baseline'|'frequency', mesh=None)` splits the data-fidelity (chi²) term across the local GPUs; the image and regularizers stay replicated. Off by default (single device).
- **baseline** — shard the visibility/baseline axis (direct + nfft)
- **frequency** — shard the channel axis (multifrequency)

The sharded objective is **bit-for-bit** the single-device one: the data axis is padded to a mesh multiple with finite-sample, infinite-sigma rows (exact-zero contribution, no closure NaN), and each chi² is corrected by `pad_len/true_len`. Validated `sharded == single-device == numpy` (gradient rel ~1e-16) for Stokes I (vis/amp/cphase/logcamp), polarimetric (IP), and multifrequency, on both transforms.

**nfft caveat:** `jax_finufft`'s `nufft2` forward shards cleanly, but its registered transpose is wrong under `shard_map` (minimal repro filed upstream). We wrap the grid→samples map in a `custom_vjp` whose backward is an explicit **forward** `nufft1` + `psum` — the exact type-1 adjoint, correct on one or many devices.

## GPU parameter surveys

`ehtim/imaging/survey_gpu.py` — `run_survey_gpu(imgr, weight_grid=, regparam_grid=, prior_fwhm=, sys_noise=, ...)` reconstructs a whole hyperparameter grid in one vmapped on-device optimization (the GPU counterpart to the CPU `paramsurvey` tool in `ehtim/survey.py`, which is untouched). Two axis kinds:
- **scalar axes** (data/reg-term weights, RegParams scalars) — trace through the objective, so the whole sub-grid runs as one `jax.vmap`'d optimization;
- **outer axes** (prior FWHM, fractional systematic noise) — rebuild the prior / re-derive the data sigmas on the host, looped and split across GPUs.

Returns the image, objective, per-row grid record, and per-term reduced chi² for each reconstruction (for ranking the Top Set).

**Backtracking line search (`optax-lbfgs-bt`).** Profiling found the survey bottleneck is *not* FP64 (only ~2.5× slower than FP32 on these GPUs) nor the forward model (`value_and_grad` vmap-parallelizes ~44× from batch 1→625). It is the optax **zoom** line-search under `vmap`: its bracket+zoom `while_loop` runs to the *batch worst-case element* every iteration (~20 value_and_grad/iter across a 625-config batch). A backtracking (Armijo) line search — a few value evals per step, keeping L-BFGS curvature — is **~7× faster at matched convergence**, and is the survey default.

**M87 demonstration** (real EHT 2019-D01-01 data, FP32): the Paper IV survey grid (5⁴ reg weights × 3 prior-FWHM × 4 systematic-noise = 7,500 reconstructions) runs in **~17 min on 4 GPUs** (~7.4 reconstructions/s), recovering the asymmetric ring across the Top Set; best fit χ²(cphase / logcamp) = 2.46 / 0.95, flux 0.60 Jy. A denser 8⁴-reg run — **49,152 reconstructions**, exceeding Paper IV's full 37,500-combination eht-imaging survey — completes in **113 min on 4 GPUs**; a single CPU reconstruction (numpy/scipy, measured via SLURM) is ~16 s, so the same survey is ~9 days sequential on CPU — a **~115× speedup**.

## Benchmarks (4× RTX PRO 6000 Blackwell, x64)

**Optimizer layer** — Gaussian reconstruction (direct, 1600 px, 200 iters):

```
  optimizer        wall s   nxcorr   chi2_vis
  scipy-lbfgsb        7.3    0.985    1.332
  optax-lbfgs         2.7    0.982    1.405
  adam                1.1    0.902  226.020
```

optax-lbfgs reaches the same minimum as scipy at **2.7× the speed** (on-device loop, no per-step host sync). adam drops in as an arbitrary optax optimizer (first-order — fast but needs more iterations to converge chi²).

**Sharding** — single value_and_grad:

```
  baseline,  direct, Nvis=2592, Npix=1600:  CPU 9.45 ms  | 1 GPU 0.18 ms (52×  CPU) | 2-4 GPU 0.3-0.4 ms
  baseline,  nfft,   Nvis=2070, Npix=1024:  CPU 37.8 ms  | 1 GPU 4.4  ms (8.6× CPU) | 4 GPU 5.6 ms | matrix-free
  frequency, 8 channels, Npix=2048:         CPU 182  ms  | 1 GPU 1.44 ms (127× CPU) | 4 GPU 1.04 ms (1.40×)
```

At these (small, compile-bound) sizes baseline visibility-axis sharding is communication-bound — multi-GPU adds overhead, so the value is per-device memory and production-scale compute: **nfft is matrix-free** (no dense `A`, so it scales to large images without the O(Nvis·Npix) matrix), and **direct** shards the dense `A` ~1/k across devices. Frequency sharding (independent channels) already scales. Larger sizes via `BENCH_NPIX` / `BENCH_TADV` / `BENCH_NCHAN`; the scripts are compile-bound (the nfft sharded graph especially).

**ALMA-scale (the regime that matters).** At Npix=16384 the dense `A` (2.71 GB) is compute-bound and baseline sharding goes **near-linear**: direct value_and_grad 7.89 ms (1 GPU) → 4.49 (2 GPU, 1.76×) → **2.45 (4 GPU, 3.23×)**; GPU vs CPU on the dense path ≈ 200×. The memory wall is `A` (Nvis×Npix): at **1024² it is ~174 GB** — exceeds a single 96 GB GPU, so baseline sharding across ≥4 GPUs is what makes the dense problem *fit*, while nfft (matrix-free) does 1024² in ~hundreds of MB on one GPU. This confirms **image-axis sharding (a deferred follow-up) is redundant**: the wall is `A`, already split by the visibility axis, and the image vector itself is ~8 MB at 1024². Full tables in `MEMORY_BASELINES.md`.

## Tests

`pytest -m jax` — 143 tests, incl. the new survey tests (objective parity vs the fixed-weight objective, batch-matches-single, prior-FWHM / sys-noise outer axes, per-term chi²). Sharding parity tests (sharded == single-device == numpy, direct/nfft × vis/closures, plus frequency) are gated on ≥2 local GPUs.

## Notes
- Pol imaging's random init is now seeded (`default_rng(0)`) — reproducible run-to-run; existing pol runs get a different (deterministic) initial guess.
- `optax` and `jax-finufft` are added to the `dev` / `gpu` extras (optional); `import ehtim` stays free of both.

---

Large + phased, stacks on #295 (the JAX objective). Based on `feature/jax-objective` for a clean diff (just the optimizer + sharding work); retargets to `dev-backend` once #295 merges.

### DIFF
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -27,7 +27,6 @@ from dataclasses import dataclass
 from typing import Any
 
 import numpy as np
-import scipy.optimize as opt
 
 import ehtim.const_def as ehc
 import ehtim.image
@@ -59,12 +58,14 @@ from ehtim.imaging.imager_backend import (
     compute_reg_dict,
     compute_reggrad_dict,
     make_objective_jax,
+    make_value_and_grad_jax,
     transform_imarr,
     unpack_imarr,
     validate_limits,
     validate_params,
 )
 from ehtim.imaging.imager_utils import embed_imarr
+from ehtim.imaging.optimizers import classify_optimizer, run_optimizer
 
 MAXIT = 200  # number of iterations
 NHIST = 50   # number of steps to store for hessian approx
@@ -203,6 +204,10 @@ class Imager:
         self._fft_pad_factor = kwargs.get('fft_pad_factor', FFT_PAD_DEFAULT)
         self._fft_interp_order = kwargs.get('fft_interp_order', FFT_INTERP_DEFAULT)
         self._nfft_eps = kwargs.get('nfft_eps', NFFT_EPS_DEFAULT)
+        self._optimizer = kwargs.get('optimizer', None)
+        self._shard = kwargs.get('shard', False)
+        self._mesh = kwargs.get('mesh', None)
+        self._shard_axis = kwargs.get('shard_axis', 'baseline')
 
         # multifrequency
         mf = kwargs.get('mf', False)
@@ -513,24 +518,58 @@ class Imager:
 
 
         tstart = time.time()
-        if kwargs.get('use_jax', False):
-            # jax autodiff objective (GPU-capable). make_objective_jax returns a
-            # scipy fun(x) -> (value, grad); jax is imported lazily inside it.
-            fun = make_objective_jax(
+        optimizer = kwargs.get('optimizer', self._optimizer)
+        use_jax = kwargs.get('use_jax', False)
+        device = kwargs.get('jax_device', None)
+        shard = kwargs.get('shard', self._shard)
+        mesh = kwargs.get('mesh', self._mesh)
+        shard_axis = kwargs.get('shard_axis', self._shard_axis)
+
+        # multi-GPU sharding runs through the on-device optax lane, so it needs an
+        # optax optimizer; default to optax-lbfgs when none was given.
+        if shard and (optimizer is None or classify_optimizer(optimizer) != 'optax'):
+            optimizer = 'optax-lbfgs'
+            print("sharding: using optax-lbfgs (the sharded objective runs on-device)")
+
+        # an optax / device optimizer needs the on-device jax objective
+        if optimizer is not None and not use_jax and classify_optimizer(optimizer) == 'optax':
+            print("using the jax objective (required by the optax optimizer)")
+            use_jax = True
+
+        def build_scipy():
+            # host objective/gradient handles for the scipy and callable lanes
+            if use_jax:
+                fun = make_objective_jax(
+                    self._init_arr, self._config, self._which_solve, self._data_tuples,
+                    self._logfreqratio_list, len(self.obslist_next),
+                    self.dat_term_next, self.reg_term_next,
+                    self._prior_arr, self.norm_reg, self._regparams(),
+                    self._embed_mask, device=device,
+                )
+                return fun, True
+            if grads:
+                return self.objfunc, self.objgrad
+            return self.objfunc, None
+
+        def build_device_vg(dev):
+            # un-jitted on-device value_and_grad for the optax lane; sharded across
+            # a GPU mesh when shard=True
+            backend_args = (
                 self._init_arr, self._config, self._which_solve, self._data_tuples,
                 self._logfreqratio_list, len(self.obslist_next),
                 self.dat_term_next, self.reg_term_next,
-                self._prior_arr, self.norm_reg, self._regparams(),
-                self._embed_mask, device=kwargs.get('jax_device', None),
+                self._prior_arr, self.norm_reg, self._regparams(), self._embed_mask,
             )
-            res = opt.minimize(fun, self._init_vec, method='L-BFGS-B', jac=True,
-                               options=optdict, callback=callback_func)
-        elif grads:
-            res = opt.minimize(self.objfunc, self._init_vec, method='L-BFGS-B', jac=self.objgrad,
-                               options=optdict, callback=callback_func)
-        else:
-            res = opt.minimize(self.objfunc, self._init_vec, method='L-BFGS-B',
-                               options=optdict, callback=callback_func)
+            if shard:
+                from ehtim.imaging.sharding import build_mesh, make_sharded_value_and_grad
+                m = mesh if mesh is not None else build_mesh()
+                return make_sharded_value_and_grad(*backend_args, mesh=m, shard_axis=shard_axis)
+            vg, loss_fn, to_device = make_value_and_grad_jax(*backend_args, device=dev)
+            return vg, loss_fn, to_device, None
+
+        res = run_optimizer(optimizer, x0=self._init_vec, optdict=optdict,
+                            callback=callback_func, build_scipy=build_scipy,
+                            build_device_vg=build_device_vg, device=device)
         tstop = time.time()
 
         # Format output
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -525,20 +525,19 @@ class Imager:
         mesh = kwargs.get('mesh', self._mesh)
         shard_axis = kwargs.get('shard_axis', self._shard_axis)
 
-        # multi-GPU sharding runs through the on-device optax lane, so it needs an
-        # optax optimizer; default to optax-lbfgs when none was given.
+        # Multi-GPU sharding runs the objective on-device through the optax lane, so it
+        # needs an optax optimizer. Require the user to opt in explicitly rather than
+        # silently overriding the optimizer they chose.
         if shard and (optimizer is None or classify_optimizer(optimizer) != 'optax'):
-            optimizer = 'optax-lbfgs'
-            print("sharding: using optax-lbfgs (the sharded objective runs on-device)")
-
-        # an optax / device optimizer needs the on-device jax objective
-        if optimizer is not None and not use_jax and classify_optimizer(optimizer) == 'optax':
-            print("using the jax objective (required by the optax optimizer)")
-            use_jax = True
+            raise ValueError(
+                "shard=True runs the sharded objective on-device, which needs an optax "
+                "optimizer; pass optimizer='optax-lbfgs' (or another optax optimizer).")
+        # (the optax lane builds the jax objective itself in build_device_vg below, so the
+        #  user's use_jax flag is irrelevant there and is left untouched.)
 
         def build_scipy():
-            # host objective/gradient handles for the scipy and callable lanes
-            if use_jax:
+            # Host (value, grad) handles for the scipy and callable lanes.
+            if use_jax:    # jitted jax objective + autodiff gradient, as a host fun(x) -> (value, grad)
                 fun = make_objective_jax(
                     self._init_arr, self._config, self._which_solve, self._data_tuples,
                     self._logfreqratio_list, len(self.obslist_next),
@@ -547,29 +546,32 @@ class Imager:
                     self._embed_mask, device=device,
                 )
                 return fun, True
-            if grads:
+            elif grads:    # default scipy with the analytic numpy gradient
                 return self.objfunc, self.objgrad
-            return self.objfunc, None
+            else:          # default scipy, no gradient (scipy finite-differences the objective)
+                return self.objfunc, None
 
         def build_device_vg(dev):
-            # un-jitted on-device value_and_grad for the optax lane; sharded across
-            # a GPU mesh when shard=True
+            # Un-jitted on-device value_and_grad (+ loss, to_device, aux) for the optax lane.
             backend_args = (
                 self._init_arr, self._config, self._which_solve, self._data_tuples,
                 self._logfreqratio_list, len(self.obslist_next),
                 self.dat_term_next, self.reg_term_next,
                 self._prior_arr, self.norm_reg, self._regparams(), self._embed_mask,
             )
-            if shard:
+            if shard:      # sharded across the GPU mesh
                 from ehtim.imaging.sharding import build_mesh, make_sharded_value_and_grad
                 m = mesh if mesh is not None else build_mesh()
                 return make_sharded_value_and_grad(*backend_args, mesh=m, shard_axis=shard_axis)
-            vg, loss_fn, to_device = make_value_and_grad_jax(*backend_args, device=dev)
-            return vg, loss_fn, to_device, None
+            else:          # single device
+                vg, loss_fn, to_device = make_value_and_grad_jax(*backend_args, device=dev)
+                return vg, loss_fn, to_device, None
 
+        # The optax lane uses the on-device builder; the scipy / callable lanes use the host builder.
+        build_loss = (build_device_vg if classify_optimizer(optimizer) == 'optax'
+                      else build_scipy)
         res = run_optimizer(optimizer, x0=self._init_vec, optdict=optdict,
-                            callback=callback_func, build_scipy=build_scipy,
-                            build_device_vg=build_device_vg, device=device)
+                            callback=callback_func, build_loss=build_loss, device=device)
         tstop = time.time()
 
         # Format output
--- a/ehtim/imager.py
+++ b/ehtim/imager.py
@@ -525,18 +525,18 @@ class Imager:
         mesh = kwargs.get('mesh', self._mesh)
         shard_axis = kwargs.get('shard_axis', self._shard_axis)
 
-        # Multi-GPU sharding runs the objective on-device through the optax lane, so it
+        # Multi-GPU sharding runs the objective on-device through the optax path, so it
         # needs an optax optimizer. Require the user to opt in explicitly rather than
         # silently overriding the optimizer they chose.
         if shard and (optimizer is None or classify_optimizer(optimizer) != 'optax'):
             raise ValueError(
                 "shard=True runs the sharded objective on-device, which needs an optax "
                 "optimizer; pass optimizer='optax-lbfgs' (or another optax optimizer).")
-        # (the optax lane builds the jax objective itself in build_device_vg below, so the
+        # (the optax path builds the jax objective itself in build_vg_ondevice below, so the
         #  user's use_jax flag is irrelevant there and is left untouched.)
 
-        def build_scipy():
-            # Host (value, grad) handles for the scipy and callable lanes.
+        def build_vg_onhost():
+            # (value, grad) as host callables, for scipy and for user-supplied optimizers.
             if use_jax:    # jitted jax objective + autodiff gradient, as a host fun(x) -> (value, grad)
                 fun = make_objective_jax(
                     self._init_arr, self._config, self._which_solve, self._data_tuples,
@@ -551,8 +551,8 @@ class Imager:
             else:          # default scipy, no gradient (scipy finite-differences the objective)
                 return self.objfunc, None
 
-        def build_device_vg(dev):
-            # Un-jitted on-device value_and_grad (+ loss, to_device, aux) for the optax lane.
+        def build_vg_ondevice(dev):
+            # Un-jitted on-device value_and_grad (+ loss, to_device, aux) for the optax path.
             backend_args = (
                 self._init_arr, self._config, self._which_solve, self._data_tuples,
                 self._logfreqratio_list, len(self.obslist_next),
@@ -567,11 +567,15 @@ class Imager:
                 vg, loss_fn, to_device = make_value_and_grad_jax(*backend_args, device=dev)
                 return vg, loss_fn, to_device, None
 
-        # The optax lane uses the on-device builder; the scipy / callable lanes use the host builder.
-        build_loss = (build_device_vg if classify_optimizer(optimizer) == 'optax'
-                      else build_scipy)
-        res = run_optimizer(optimizer, x0=self._init_vec, optdict=optdict,
-                            callback=callback_func, build_loss=build_loss, device=device)
+        # Pick the builder that matches the optimizer: optax runs the objective on the
+        # GPU, while scipy and user callables want host arrays.
+        if classify_optimizer(optimizer) == 'optax':
+            build_loss = build_vg_ondevice
+        else:
+            build_loss = build_vg_onhost
+
+        res = run_optimizer(optimizer, build_loss, x0=self._init_vec, optdict=optdict,
+                            callback=callback_func, device=device)
         tstop = time.time()
 
         # Format output
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -219,7 +219,7 @@ class ImagerInitState(NamedTuple):
     data_tuples: dict            # keyed by dname or f"{dname}_{i}"
     embed_mask: np.ndarray       # boolean
     coord_matrix: np.ndarray     # pixel-coord companion to embed_mask
-    logfreqratio_list: list      # log(nu_i / reffreq)
+    logfreqratio_list: np.ndarray  # log(nu_i / reffreq)
     nimage: int                  # number of active pixels = sum(embed_mask)
     which_solve: np.ndarray      # int 0/1 flags
     reffreq: float               # may be re-bound to init.rf when mf=True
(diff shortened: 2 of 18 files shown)


rohandahale:
@achael This PR is ready for review!

ehtim/imaging/imager_backend.py
```diff
@@ -633,15 +633,16 @@ def make_initarr(image, mask, norm_init=False, flux=1,
         # Caller (compute_init_state) decides when random-pol init applies.
         # Here we just honor the flag: True means "use random pol initialization
         # regardless of init image content"; False means "use init image's pol".
+        pol_rng = np.random.default_rng(0)  # seeded so the random pol init is reproducible
```
achael: add a seed argument to the function?

ehtim/imaging/imager_backend.py
```diff
@@ -1974,45 +1975,67 @@ def compute_objective_grad(imvec, initvec, config,
...
     data_d = {key: tuple(put(v) if hasattr(v, "shape") else v for v in val)
               for key, val in data_tuples.items()}
-    prior_d, init_d = put(priorvec), put(initvec)
+    return data_d, put(priorvec), put(initvec), put
+
+
+def _prepare_jax_loss(initvec, config, which_solve, data_tuples, logfreqratio_list,
+                      n_obs, dat_term, reg_term, priorvec, norm_reg, reg_params,
+                      embed_mask, device=None):
+    """Build the shared jax loss(x) closure for the imaging objective.
```
achael: useful to add use case in docstring -- this is in both make_objective_jax and make_value_and_grad_jax

ehtim/imaging/imager_backend.py
```diff
@@ -1974,45 +1975,67 @@ def compute_objective_grad(imvec, initvec, config,
...
                                  logfreqratio_list, n_obs, dat_term, reg_term,
                                  prior_d, norm_reg, reg_params, embed_mask)
 
+    return loss, put
+
+
+def make_objective_jax(initvec, config, which_solve, data_tuples, logfreqratio_list,
+                       n_obs, dat_term, reg_term, priorvec, norm_reg, reg_params,
+                       embed_mask, device=None):
+    """Return a scipy fun(x) -> (value, grad) for the imaging objective, via jax.
```
achael: as a jax novice it would be useful in the docstring here and in make_value_and_grad_jax to to make the use cases for the different objective functions a little clearer

ehtim/imager.py
```diff
@@ -513,24 +518,58 @@ def callback_func(xcur):
...
+
+        # multi-GPU sharding runs through the on-device optax lane, so it needs an
+        # optax optimizer; default to optax-lbfgs when none was given.
+        if shard and (optimizer is None or classify_optimizer(optimizer) != 'optax'):
+            optimizer = 'optax-lbfgs'
+            print("sharding: using optax-lbfgs (the sharded objective runs on-device)")
+
+        # an optax / device optimizer needs the on-device jax objective
+        if optimizer is not None and not use_jax and classify_optimizer(optimizer) == 'optax':
+            print("using the jax objective (required by the optax optimizer)")
```
achael: I think it's better to raise here than automatically changing the user flags

ehtim/imager.py
```diff
@@ -513,24 +518,58 @@ def callback_func(xcur):
...
+        device = kwargs.get('jax_device', None)
+        shard = kwargs.get('shard', self._shard)
+        mesh = kwargs.get('mesh', self._mesh)
+        shard_axis = kwargs.get('shard_axis', self._shard_axis)
+
+        # multi-GPU sharding runs through the on-device optax lane, so it needs an
+        # optax optimizer; default to optax-lbfgs when none was given.
+        if shard and (optimizer is None or classify_optimizer(optimizer) != 'optax'):
+            optimizer = 'optax-lbfgs'
+            print("sharding: using optax-lbfgs (the sharded objective runs on-device)")
```
achael: I think it's better to raise here than automatically changing the optimizer

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,208 @@
...
+the caller unpacks `.x`/`.fun` the same way for every backend.
+"""
+from functools import partial
+
+import numpy as np
+import scipy.optimize
+
+# Built-in optax optimizers, resolved in the optax lane. Listed here so
+# classify_optimizer can route the names without importing optax.
+_OPTAX_BUILTINS = frozenset({"optax-lbfgs", "adam", "adamw", "sgd", "rmsprop"})
```
achael: make parallel names -- "_OPTAX_NAMES" and "_SCIPY_NAMES" ?

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,208 @@
...
+        Carries at least `.x` and `.fun`.
+    """
+    kind = classify_optimizer(optimizer)
+
+    if kind == "scipy":
+        fun, jac = build_scipy()
+        return scipy.optimize.minimize(fun, x0, method="L-BFGS-B", jac=jac,
+                                       options=optdict, callback=callback)
+
+    if kind == "callable":
```
achael: I prefer elif here even if not necessary

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,208 @@
...
+            value_and_grad = fun           # fun already returns (value, grad)
+        elif jac is None:
+            def value_and_grad(x):
+                return fun(x), None
+        else:
+            def value_and_grad(x):
+                return fun(x), jac(x)
+        return optimizer(value_and_grad, x0, maxiter=optdict["maxiter"],
+                         tol=optdict["gtol"], callback=callback)
+
```
achael, line 104: and would prefer this to be in an "else" block

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,208 @@
...
+    # an optax GradientTransformation is a NamedTuple exposing init/update
+    if hasattr(optimizer, "init") and hasattr(optimizer, "update"):
+        return "optax"
+    if callable(optimizer):
+        return "callable"
+    raise TypeError(f"unsupported optimizer {optimizer!r}")
+
+
+def run_optimizer(optimizer, *, x0, optdict, callback=None,
+                  build_scipy=None, build_device_vg=None, device=None, mesh=None):
```
achael: since build_scipy and build_device_vg are never used together, can this be one argument? That is "build_loss = build_scipy" or "build_loss = build_device_vg", and then have a check in each block that the passed in functions are appropriate for the specified lane?

ehtim/imaging/imager_backend.py
```diff
@@ -2021,3 +2044,53 @@ def fun(x):
         return float(value), np.asarray(grad, dtype=np.float64)
 
     return fun
+
```
achael, line 2085: do we need both? it sounds like from the docstring that make_value_and_grad_jax is the better choice for gpus -- is make_objective_jax better for cpus? If not, can we get rid of it and only use make_value_and_grad_jax?

ehtim/imager.py
```diff
@@ -513,24 +518,58 @@ def callback_func(xcur):
...
+
+        # an optax / device optimizer needs the on-device jax objective
+        if optimizer is not None and not use_jax and classify_optimizer(optimizer) == 'optax':
+            print("using the jax objective (required by the optax optimizer)")
+            use_jax = True
+
+        def build_scipy():
+            # host objective/gradient handles for the scipy and callable lanes
```
achael: let's use clear if-else with comments: 
if use_jax: #jitted jax objective and gradient
...
elif grads: # default scipy with analytic gradients
...
else: # default scipy without gradies

ehtim/imager.py
```diff
@@ -513,24 +518,58 @@ def callback_func(xcur):
...
-            res = opt.minimize(self.objfunc, self._init_vec, method='L-BFGS-B', jac=self.objgrad,
-                               options=optdict, callback=callback_func)
-        else:
-            res = opt.minimize(self.objfunc, self._init_vec, method='L-BFGS-B',
-                               options=optdict, callback=callback_func)
+            if shard:
+                from ehtim.imaging.sharding import build_mesh, make_sharded_value_and_grad
+                m = mesh if mesh is not None else build_mesh()
+                return make_sharded_value_and_grad(*backend_args, mesh=m, shard_axis=shard_axis)
+            vg, loss_fn, to_device = make_value_and_grad_jax(*backend_args, device=dev)
```
achael: again add "else" (just easier for me to read this way -- if there's a legit reason not to or its not correct python style i'm happy to be corrected....)

ehtim/imaging/imager_backend.py
```diff
@@ -1974,45 +1975,67 @@ def compute_objective_grad(imvec, initvec, config,
...
-    # and reuse that compilation across every L-BFGS-B step.
     def loss(x):
         return compute_objective(x, init_d, config, which_solve, data_d,
                                  logfreqratio_list, n_obs, dat_term, reg_term,
                                  prior_d, norm_reg, reg_params, embed_mask)
 
+    return loss, put
+
+
+def make_objective_jax(initvec, config, which_solve, data_tuples, logfreqratio_list,
```
achael, line 2042: i think it would help clarify things if make_objective_jax, make_value_and_grad_jax, and make_survey_value_and_grad had more parallel names to better illustrate their use case. One suggestion from my naive read would be "make_objective_jax_jit" "make_objective_jax_device" "make_objective_jax_survey" --> but you have a better picture of the code and so don't adopt these unless they make sense

achael (commented):
looks great overall, but I think my general feedback so far is just to do a human pass on the docstring and comments to make the flow here clearer for non-experts. Full docstrings with explicit argument/output specs for the new functions would help too. 
Some other minor comments as well.

ehtim/imaging/imager_backend.py
rohandahale: Done, added a seed arg to `make_initarr` (default value as 0) and passed it through `compute_init_state`

ehtim/imaging/imager_backend.py
rohandahale: Done, rewrote the docstring

ehtim/imaging/imager_backend.py
rohandahale: Done, the docstring now leads with the use case and points to `make_value_and_grad_jax` for the on-device optax lane. Same treatment on `make_value_and_grad_jax`.

ehtim/imager.py
rohandahale: Removed the auto-switch rather than raise. The optax lane builds the jax objective itself, so `use_jax` has no effect there.

ehtim/imager.py
rohandahale: Done! `shard=True` without an optax optimizer now raises a clear `ValueError` telling the user to pass `optimizer='optax-lbfgs'`, instead of silently switching it.

ehtim/imager.py
rohandahale: Done
```
if use_jax:   # jitted jax objective + autodiff gradient
elif grads:   # scipy with the analytic numpy gradient
else:            # scipy, no gradient
```

ehtim/imager.py
rohandahale: Done, `build_device_vg` now branches if shard: ... else: ... with comments (sharded across the mesh vs single device).

ehtim/imaging/optimizers.py
rohandahale: Done!

ehtim/imaging/optimizers.py
rohandahale: Done, if/elif/else across the three cases.

ehtim/imaging/optimizers.py
rohandahale: Done!

ehtim/imaging/imager_backend.py
rohandahale: I left the names as is for now. The use case docstrings seemed to have cleared up the confusion?

ehtim/imaging/imager_backend.py
rohandahale: Kept both but now `make_objective_jax` into a thin wrapper over `make_value_and_grad_jax`. So there's one place building the device `value_and_grad`, and `make_objective_jax` is visibly just its jitted host scipy adapter. The docstrings now make that relationship explicit.

ehtim/imaging/optimizers.py
rohandahale: Done

rohandahale:
Thanks @achael ! 

Done, I did a pass over the docstrings and comments, and added explicit argument/output descriptions to  the new jax objective builders (make_objective_jax / make_value_and_grad_jax / make_survey_value_and_grad, plus the  _prepare_jax_loss helper). They point to compute_objective for the shared backend arguments rather than repeating all 12 in each. 

Ready for re-review / merge!

ehtim/imager.py
```diff
@@ -513,24 +518,60 @@ def callback_func(xcur):
...
+        # needs an optax optimizer. Require the user to opt in explicitly rather than
+        # silently overriding the optimizer they chose.
+        if shard and (optimizer is None or classify_optimizer(optimizer) != 'optax'):
+            raise ValueError(
+                "shard=True runs the sharded objective on-device, which needs an optax "
+                "optimizer; pass optimizer='optax-lbfgs' (or another optax optimizer).")
+        # (the optax lane builds the jax objective itself in build_device_vg below, so the
+        #  user's use_jax flag is irrelevant there and is left untouched.)
+
+        def build_scipy():
```
achael: what about calling this 'build_vg_scipy'?

ehtim/imager.py
```diff
@@ -513,24 +518,60 @@ def callback_func(xcur):
...
+                    self._prior_arr, self.norm_reg, self._regparams(),
+                    self._embed_mask, device=device,
+                )
+                return fun, True
+            elif grads:    # default scipy with the analytic numpy gradient
+                return self.objfunc, self.objgrad
+            else:          # default scipy, no gradient (scipy finite-differences the objective)
+                return self.objfunc, None
+
+        def build_device_vg(dev):
```
achael: what about calling this 'build_vg_optax'?

ehtim/imager.py
```diff
@@ -513,24 +518,60 @@ def callback_func(xcur):
...
+            if shard:      # sharded across the GPU mesh
+                from ehtim.imaging.sharding import build_mesh, make_sharded_value_and_grad
+                m = mesh if mesh is not None else build_mesh()
+                return make_sharded_value_and_grad(*backend_args, mesh=m, shard_axis=shard_axis)
+            else:          # single device
+                vg, loss_fn, to_device = make_value_and_grad_jax(*backend_args, device=dev)
+                return vg, loss_fn, to_device, None
+
+        # The optax lane uses the on-device builder; the scipy / callable lanes use the host builder.
+        build_loss = (build_device_vg if classify_optimizer(optimizer) == 'optax'
```
achael: this this is a central bit of logic to determine which of the two build functions above are used i'd prefer it set apart in an explicit block: 
if classify_optimizer(optimizer)=='optax':
   build_loss = build_device_vg
else:
  build_loss = build_scipy

ehtim/imager.py
achael: or 'build_vg_ondevice'

ehtim/imager.py
achael: or 'build_vg_onhost'

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,213 @@
...
+        raise ValueError(f"unknown optimizer name {optimizer!r}")
+    # an optax GradientTransformation is a NamedTuple exposing init/update
+    if hasattr(optimizer, "init") and hasattr(optimizer, "update"):
+        return "optax"
+    if callable(optimizer):
+        return "callable"
+    raise TypeError(f"unsupported optimizer {optimizer!r}")
+
+
+def run_optimizer(optimizer, *, x0, optdict, callback=None, build_loss=None,
```
achael: should build_loss be a kwarg? Since we always need it to me it makes more sense as  a normal arg early in the sequence but i'm not sure.

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,213 @@
...
+    Returns
+    -------
+    scipy.optimize.OptimizeResult
+        Carries at least `.x` and `.fun`.
+    """
+    kind = classify_optimizer(optimizer)
+
+    if kind == "scipy":
+        fun, jac = build_loss()
+        return scipy.optimize.minimize(fun, x0, method="L-BFGS-B", jac=jac,
```
achael, line 86: TODO for future -- other scipy optimizers beyond L-BFGS-B using this lane? Or unnecessary?

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,213 @@
...
+                                       options=optdict, callback=callback)
+
+    elif kind == "callable":
+        # The escape hatch: hand the user a host value_and_grad(x) -> (value, grad).
+        fun, jac = build_loss()
+        if jac is True:
+            value_and_grad = fun           # fun already returns (value, grad)
+        elif jac is None:
```
achael, line 94: add line comment this is the finite diff case

ehtim/imaging/optimizers.py
```diff
@@ -0,0 +1,213 @@
...
+        return x, val, i
+
+    x, val, nit = run(x0, aux)
+    return scipy.optimize.OptimizeResult(
+        x=np.asarray(x, dtype=np.float64), fun=float(val), nit=int(nit),
+        njev=int(nit), success=True, status=0,
+        message="optax on-device convergence")
+
+
+def optimize_fixed(value_and_grad, loss, x0, gt, needs_ls, maxiter):
```
achael, line 221: why is this only invoked in survey_gpu? i.e.  can if use use optax in the normal Imager, can we cut it off at maxiter as we can for the scipy optimizers?

achael (changes requested):
Very major and dense upgrade. i have a better understanding of the structure now and think it makes sense overall. just a few minor comments. 

I do think we should revisit these changes (particularly in imager_backend, sharding, optimizers, and imager) and try to touch up the docstrings to make them a little more human readable / less AI in style. Ideally the logic will be transparent even to someone with a just a basic understanding of GPU programming (like me). I don't want to hold things up for now so I think this can happen later -- I may take a pass at this myself later.

ehtim/imager.py
rohandahale: I renamed it as `build_vg_onhost`

ehtim/imager.py
rohandahale: I renamed it as `build_vg_ondevice`

ehtim/imager.py
rohandahale: Done, it's an explicit block now (with the renamed functions):

```
# Pick the builder that matches the optimizer: optax runs the objective on the
# GPU, while scipy and user callables want host arrays.
if classify_optimizer(optimizer) == 'optax':
    build_loss = build_vg_ondevice
else:
   build_loss = build_vg_onhost
```

ehtim/imaging/optimizers.py
rohandahale: Agreed, it's a normal arg now:

`def run_optimizer(optimizer, build_loss, *, x0, optdict, callback=None, device=None):
`

ehtim/imaging/optimizers.py
rohandahale: Added a TODO there. It would be easy enough to add a method argument. But optdict currently carries L-BFGS-B only keys (maxcor, maxls), so those would need to become optional first.

ehtim/imaging/optimizers.py
rohandahale: Added:

```
elif jac is None:
   # no analytic gradient (grads=False); the scipy path would finite-difference
   # here, so hand the user None and let their optimizer decide
```

ehtim/imaging/optimizers.py
rohandahale: It turns out the normal Imager already does cut off at maxiter. That's the `i < maxiter` term in the `while_loop` condition, and it means the same thing as scipy's `maxiter` option.

The only real difference between the two is when they stop.` _run_optax` stops on whichever comes first: gradient norm, relative change in the value, or maxiter. `optimize_fixed` has no early exit at all and always runs the full count.

Surveys need that second one because they run the whole hyperparameter grid at once under vmap, and "stop when this one has converged" doesn't mean anything for a batch. So you would end up running to the slowest one anyway.

Since the stopping rule was the only thing that actually differed, I pulled the shared part out into a small `_optax_step`.

rohandahale:
Thanks @achael! I went through the whole PR again and rewrote many docstrings by manually. I realized that this PR was big and should have been split into two. But that is a lesson for future PRs.

`sharding.py` module's docstrings are fully rewritten. It now walks through it in steps, what actually gets split across the GPUs and why, then the padding (why sigma is infinity but the operator is ones), then why nfft needs a hand written gradient. Same for the functions in there. 

`optimizers.py` and the block in `imager.py` are rewritten as well.

achael (approved):
looks good -- i will plan to go through and review docstrings / minor style structure changes myself in a future PR.

TODO for future will be to make sure survey.py and survey_gpu.py have the same functionality  / similar API. Or just merge them into one file that can handle both cpu and gpu runs.

PR #306, opened 2026-06-16
pol gradient fixes to dev

achael:
## What & why

Stops overloading `pol_solve`, which conflated two distinct masks: (1) which
solver DOFs are optimized, and (2) which physical gradient slots (I, rho, phi,
psi) the chisq/reg kernels must fill. These differ wherever a value-transform's
Jacobian is non-diagonal (mcv, vcv): one solved pol DOF drives BOTH rho and psi.

Introduces `physical_grad_slots(pol_solve, transforms)` as the single source of
truth, wired into both the chisq and regularizer gradient dicts, and reverts the
kernels to clean diagonal gating.

### Correctness fixes
- **mcv/vcv cross-coupling**, centralized in `physical_grad_slots`. This fixes
  IP imaging when V!=0: `mcv_grad`'s m' gradient needs the psi
  physical gradient, which the old mask zeroed.
- **`reggrad_ptv` first-row/col boundary masking** + `epsilon_tv` in
  `reg_ptv`/`reggrad_ptv` and `reg_vtv`/`reggrad_vtv` (default 0, byte-identical).

### Test coverage
- `physical_grad_slots` unit tests; pol chisq FD all-slots (`pvis`/`m`/`vvis` x
  direct/nfft); pol-reg FD driven with all four slots; boundary FD regressions
  for `ptv`/`vtv`/`tv`; one parametrized objective-FD over {IP/mcv, IV/vcv,
  IQUV/polcv} x {direct,nfft} that samples the **polarization DOF block** (the
  existing global-sampling tests missed the m'-coupling bug).
- FD fixtures switched to an **asymmetric** image (offset double-Gaussian) with
  spatially-varying pol (`add_random_pol`, ccorr>0) so EVPA/vfrac/rho/psi all
  vary -- surfaces boundary/axis bugs a centered constant-fraction Gaussian hides.

### Cosmetic
- Comment/docstring cleanup pass on `imager_utils.py` and `pol_imager_utils.py`
  (no behavior change), ruff-clean.

## Supersedes part of #295
The `reggrad_ptv` boundary + `epsilon_tv` fixes and the reg-grad cross-coupling
land here; #295 should drop those and stay jax-agnostic plumbing (Option B).

## Testing
Full suite: 1776 passed, 2 skipped, 1 xfailed.

PR #307, opened 2026-06-16
pol gradient fixes + tests to main (v1.3.2)

achael:
## What & why
Patches a latent bug in the polarimetric imager gradients on `main`,
mirroring the dev fix in #306, and fixes several further latent
pol-gradient bugs uncovered while adding test coverage. Ships an interim
gradient-validation suite and wires it into CI.

### Core fix (mirrors #306)
- `physical_grad_slots` separates the physical-gradient mask from the
  solve mask, centralizing the mcv/vcv cross-coupling (one solved pol DOF
  drives BOTH rho and psi). Fixes IP imaging when V!=0 (and the IV analog).
- `mcv_grad`/`vcv_grad`: add the off-diagonal Jacobian terms (dpsi/dm',
  drho/dv'); return 3-wide and assign to outarr[1:4] so the Stokes-I log
  Jacobian survives. Same preservation for the polcv path.
- `stv_pol_grad` (ptv): first-row/col back-neighbor masking (corner was
  ~4x off vs finite differences).

### Further latent bugs fixed (found via the new tests; all pre-existing, all in never-exercised Stokes-I / V-flux slots)
These paths are already correct on dev -- its gradient test coverage caught
them; this PR backports the equivalent fixes to main.
- `chisqgrad_m`: restore dropped `mimage*exp(-2i*chi)` factor (was a
  TypeError from `np.real(  * ...)`).
- `smgrad` / `svfluxgrad`: `outgrad` / `gradoutout` -> `gradout` (NameError).
- `make_chisq_dict`: pass the Stokes-I row to chisq for standard data
  terms in pol mode (the gradient path was already correct).

### #233 port
- NumPy >= 1.24 fixes (`dtype=object` on ragged diag arrays, `np.float` ->
  `float`) so the diagonalized closures build.

### Tests + CI
- `tests/test_gradients_main_tmp.py`: interim gradient suite (71 tests) --
  nfft-vs-direct, analytic-vs-FD for all intensity/pol chi^2 terms and
  regularizers, and the mcv/vcv/polcv transforms vs FD (isolated +
  end-to-end through the Imager objective). Temporary; delete when dev
  lands on main.
- CI `test` job on py3.8 + py3.11 running the suite (pynfft absent in CI
  -> nfft cases skip, direct-FT cases run).
- Drop the `*test*` gitignore rule (it hid top-level tests/).

### Version
- Bump to 1.3.2.

## How to test
`python3 -m pytest tests/test_gradients_main_tmp.py`
System python (has pynfft): 71 pass. Without pynfft: 44 pass + 12 nfft-skip.

## Design decisions
- Interim suite is intentionally gradient-only; broader coverage comes
  from dev.
- Pol fixtures build Q/U/V manually instead of add_random_pol, avoiding a
  scipy-version-fragile scattering path (and the rect-image scattering bug).
- nfft gated on pynfft (what main's nfft ttype actually uses).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,24 @@ on:
     branches: [main]
 
 jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.11"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+      - run: pip install -e ".[dev]"
+      # Interim gradient suite only. pynfft (NFFT C library) is not installed
+      # here, so the nfft cases skip and the direct-FT cases run.
+      - run: pytest tests/test_gradients_main_tmp.py
+
   package:
     runs-on: ubuntu-latest
     timeout-minutes: 10
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ vex_files/*
 docs/_build/*
 docs/build/
 local/*
-*test*
+.pytest_cache/
 imagingChallenge*
 tmp/*
 *.uvfits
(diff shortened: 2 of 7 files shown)


rohandahale (approved):
I reviewed and ran the gradient tests locally and all of them pass. Looks good to merge!

PR #311, opened 2026-06-23
Fix NaN in spectral TV regularizer on flat maps (epsilon_tv=0)

rohandahale:
## What

`reg_tv_spec` / `reggrad_tv_spec` produced `NaN` on a **flat spectral map** when `epsilon_tv == 0` (the default since #310). A flat coefficient map is exactly the multifrequency imaging start — `add_const_mf` initialises a constant spectral index, so every neighbour difference is zero and the naive `sqrt(0)` leaves a `0/0`:

- in the **value** (the autodiff path under JAX), and
- in the **analytic gradient** (the NumPy path).

Downstream this killed the optimizer: an HL Tau `mf_order=1` run gave `J: nan` and reconstructed a flat blob instead of the disk. With the fix it converges to `chi2_vis ~ 1` and nxcorr 0.999 vs the exact v1.2.5 result.

## Fix

Double-`where` guard in both functions (autodiff-safe; a flat pixel gets gradient 0, not `0/0`). Minimal and self-contained — only `multifreq_imager_utils.py` is touched, no new shared helper.

## Test

New `TestSpectralTVFlatMapFinite` in `test_gradients.py`: every `tv_*` spectral regularizer stays finite on a flat map at `epsilon_tv=0`, checked on **both** the NumPy analytic gradient and `jax.grad` of the value. The existing S5 FD-parity suite uses a structured map with `epsilon_tv != 0`, which is why it never caught this.

## Origin

Regression from #310, which flipped the `epsilon_tv` default from `1e-12` to `0`. Backend-touching. May warrant a release cherry-pick like the #240 `stv_pol_grad` fix.

### DIFF
--- a/ehtim/imaging/multifreq_imager_utils.py
+++ b/ehtim/imaging/multifreq_imager_utils.py
@@ -254,7 +254,11 @@ def reg_tv_spec(imvec, mask, **kwargs):
     impad = xp.pad(im, 1, mode='constant', constant_values=0)
     im_l1 = xp.roll(impad, -1, axis=0)[1:ny+1, 1:nx+1]
     im_l2 = xp.roll(impad, -1, axis=1)[1:ny+1, 1:nx+1]
-    return xp.sum(xp.sqrt(xp.abs(im_l1 - im)**2 + xp.abs(im_l2 - im)**2 + epsilon)) / norm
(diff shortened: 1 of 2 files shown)


achael (approved):
looks good -- main still has default epsilon=1e-12, so I think we just need to cherry pick this into dev as well.

PR #314, opened 2026-07-02
Fast non-averaging uvfits loading (multi-frequency obslist) + speedups default

rohandahale:
Multi-frequency ALMA/VLA imaging needs each channel/IF kept as its own frequency point, loaded without the per-record loops that don't finish on large unaveraged files. The averaging default is unchanged, so existing scripts are unaffected.

1. `load_uvfits` now has `average_if` and `average_channel` (both default True). Left at the default it behaves exactly as before, one averaged `Obsdata`. Set either to False and it keeps that spectral axis resolved and returns a `list[Obsdata]`, one per surviving frequency group, each with its own reference frequency and its own u,v. That list is exactly what the multi-frequency imager takes as its obslist, so you can go straight from one native multi-channel / multi-IF uvfits to an mf obslist instead of pre-splitting per-SPW by hand in CASA.

2. **`trial_speedups` → `speedups`, default True.** The vectorized read paths added in 2022 are now verified identical to the legacy, so `load_uvfits` uses them by default and the flag drops the "trial". This makes the default load fast.

## How it works

New internal `load_obs_uvfits_spectral` reads the DATA one (IF, channel) slice at a time and groups the (IF, channel) grid — an averaged axis collapses to a single inverse-variance group, a kept axis yields one group per index. `load_obs_uvfits` (the averaged path) is unchanged apart from the rename.

## Backward-compatibility

- **`trial_speedups` renamed.** It was an experimental opt-in (default False), so external use should be near nil, but a script passing `trial_speedups=...` now gets a `TypeError`
- `Obsdata` when both flags are True (the default), `list[Obsdata]` otherwise.

### DIFF
--- a/ehtim/io/load.py
+++ b/ehtim/io/load.py
@@ -1051,7 +1051,7 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
                     allow_singlepol=True, force_singlepol=None,
                     channel=all, IF=all, remove_nan=False,
                     ignore_pzero_date=True,
-                    trial_speedups=True,
+                    speedups=True,
                     invvar_channel_avg=True):
     """Load observation data from a uvfits file.
 
@@ -1067,9 +1067,9 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
            remove_nan (bool): whether or not to remove entries with nan data
            ignore_pzero_date (bool): if True, ignore the offset parameters in DATE field
                                      TODO: what is the correct behavior per AIPS memo 117?
-           trial_speedups (bool): if True (default), use faster vectorized
-                                  array/telescope handling paths; produces data
-                                  and tarr identical to the legacy paths
+           speedups (bool): if True (default), use faster vectorized
+                            array/telescope handling paths; produces data
+                            and tarr identical to the legacy paths
            invvar_channel_avg (bool): if True, average IFs/channels with inverse-variance
                                       weighting. If False, use the original simple averaging.
        Returns:
@@ -1379,7 +1379,7 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
     t2c = t2c - 1
 
     # TODO make site identificantion faster
-    if trial_speedups and (not np.any(tnums!=np.arange(len(tnums)))):
+    if speedups and (not np.any(tnums!=np.arange(len(tnums)))):
         sites = tarr['site']
         t1 = sites[t1c]
         t2 = sites[t2c]
@@ -1519,7 +1519,7 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
         poldict_out = ehc.POLDICT_STOKES
 
     #TODO new, faster,
-    if trial_speedups:
+    if speedups:
         datatable = np.empty((len(times)),dtype=dtpol_out)
         datatable['time'] = times
         datatable['tint'] = tints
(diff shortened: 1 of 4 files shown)


ehtim/io/load.py
```diff
@@ -1570,6 +1570,390 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
...
+            Each carries its own rf and per-group u,v, so ``[o.rf for o in obslist]`` is
+            the frequency list the multi-frequency imager expects.
+    """
+
+    if polrep not in ('stokes', 'circ'):
+        raise Exception("polrep should be 'stokes' or 'circ' in load_uvfits_spectral")
+    if not (force_singlepol is None or force_singlepol is False) and polrep != 'stokes':
+        raise Exception("force_singlepol is incompatible with polrep!='stokes'")
+
+    if isinstance(filename, fits.HDUList):
```
achael, line 1649: is there a way to avoid so much code duplication with load_obs_uvfits? i.e. can we merge the two functions so they share a lot of the preliminaries? or do you prefer to keep them separate for now to avoid too many modifications with the original loader? if so i suggest a TODO

ehtim/obsdata.py
```diff
@@ -124,7 +124,7 @@ class Obsdata:
     def __init__(self, ra, dec, rf, bw, datatable, tarr, scantable=None,
                  polrep='stokes', source=ehc.SOURCE_DEFAULT, mjd=ehc.MJD_DEFAULT, timetype='UTC',
                  ampcal=True, phasecal=True, opacitycal=True, dcal=True, frcal=True,
-                 trial_speedups=False):
+                 speedups=False):
```
achael: see below, i would remove if the non sped up version of reorder_baselines no longer exists.

ehtim/obsdata.py
```diff
@@ -511,7 +511,7 @@ def _warn_singlepol(n, msg):
 
         return newobs
 
-    def reorder_baselines(self, trial_speedups=False):
+    def reorder_baselines(self, speedups=False):
```
achael: if the non-vectorized version is removed from reorder_baselines, we should remove the argument. Old scripts that use it would break anyway because we have changed the kwarg.

tests/test_obsdata.py
```diff
@@ -235,8 +235,8 @@ def test_reorder_baselines_time_sorted(obs_direct):
     assert np.all(np.diff(obs_direct.data["time"]) >= 0)
 
 
-def test_trial_speedups_matches_default(obs_direct):
-    # The trial_speedups flag in reorder_baselines is now a no-op; both
+def test_speedups_matches_default(obs_direct):
```
achael: remove or change to an exception test if we remove the speedups option from reorder_baselines

achael (changes requested):
looks good, my main concern is just that the new load_obs_uvfits_spectral has so much duplicated code with the original load_obs_uvfits. If there is an elegant way to reduce the overlap (create a helper function that can be called by both for some of the header material, etc that would be great. If you think it's better to keep load_obs_uvfits relatively untouched that also makes sense, but then i think we should mark better integration of the two functions as a TODO since we want to avoid future changes that touch one and not the other.

ehtim/io/load.py
rohandahale: `_read_antenna_table`, `_read_source_position` and `_read_scan_table` (new helpers) were identical that both loaders call. So I removed that duplication. I left the rest separate.

ehtim/obsdata.py
rohandahale: Done

ehtim/obsdata.py
rohandahale: Removed

tests/test_obsdata.py
rohandahale: Changed to an exception test

rohandahale:
Thanks for checking @achael! I added helpers rather than a TODO. 

I did leave the data reading separate on purpose: one averages the spectral axis away and the other keeps it.

Ready for re-review / merge!

achael (approved):
looks good -- we can probably remove the speedups option for load_uvfits in a future PR

PR #315, opened 2026-07-12
obs_helpers: seed noise from crc32 instead of builtin hash

Lftw:
## What
The hash-seeded random helpers in `obs_helpers.py` (`cerror_hash`, `hashrandn`, `hashrand`, `hashmultivariaterandn`) seeded numpy from the builtin `hash()` of a repr string built from each quantity's identity (sites/time/pol/seed). They now route through a single new `_stable_seed()` helper that uses `zlib.crc32`.

## Why
CPython salts `str` hashing per process (`PYTHONHASHSEED`), so the seed and all simulated thermal noise and gain/D-term errors changed on every run, even with the same `seed` argument. `zlib.crc32` of the same text is identical across processes, machines, and Python versions, so simulated noise is now reproducible while preserving the per-quantity independence the helpers were designed for (draws keyed to a visibility's identity, independent of data ordering).

Surfaced while regression-testing simulated observations: clean (noise-free) obs were bit-identical across processes but noisy obs were not; pinning `PYTHONHASHSEED=0` made them identical, which pointed at the salted `hash()`.

## How to test
- `pytest tests/test_obs_helpers.py`  pins the crc32 implementation and asserts drawn noise/gain values are byte-identical across `PYTHONHASHSEED` ∈ {0, 1, 999999} via subprocesses (fails on the old `hash()`).
- Full calibration/noise sweep (`test_obs_simulate`, `test_self_cal`, `test_network_cal`, `test_pol_cal`, `test_polgains_cal`, `test_caltable`, `test_obsdata`): 340 passed.

## Notes
- Seed *values* change vs. the old code (crc32 ≠ hash), so noise realizations differ from previous runs but the old ones were never reproducible anyway.
- Kept separate from the mixed-pol work.

### DIFF
--- a/tests/test_obs_helpers.py
+++ b/tests/test_obs_helpers.py
@@ -1,16 +1,20 @@
-"""Tests for ehtim.observing.obs_helpers.
-
-Focus: the hash-seeded random helpers must be reproducible across Python
-processes. CPython salts str hashing per process (PYTHONHASHSEED), so seeding
-numpy from the builtin hash() produced different simulated noise every run;
-_stable_seed (zlib.crc32) makes the draws identical across processes.
-"""
+"""Tests for ehtim.observing.obs_helpers."""
 import os
 import subprocess
 import sys
 import zlib
 
 from ehtim.observing import obs_helpers as obsh
+
+
+# ---------------------------------------------------------------------------
+# Hash-seeded random helpers: reproducibility across processes
+#
+# The hash-seeded random helpers must be reproducible across Python processes.
+# CPython salts str hashing per process (PYTHONHASHSEED), so seeding numpy from
+# the builtin hash() produced different simulated noise every run; _stable_seed
+# (zlib.crc32) makes the draws identical across processes.
+# ---------------------------------------------------------------------------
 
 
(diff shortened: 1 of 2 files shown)


rohandahale:
@Lftw Looks good!

For the next PR, we can do local generator seeded through `np.random.default_rng(np.random.SeedSequence(key)).standard_normal()`, no global state.

rohandahale (dismissed):

tests/test_obs_helpers.py
```diff
@@ -0,0 +1,53 @@
+"""Tests for ehtim.observing.obs_helpers.
+
+Focus: the hash-seeded random helpers must be reproducible across Python
```
achael: i'd remove this part from the top docstring; we will want to add more tests of other obs_helpers functions. Could add a header above the current tests to indicate they are focused on the seeds/random draws

achael (dismissed):
looks good! only a minor suggestion on organization of the test file.

tests/test_obs_helpers.py
Lftw: Sounds good! fixed below

rohandahale (approved):

PR #317, opened 2026-07-22
Mixed/linear-feed polarimetry: simulator + Jones + uvfits round-trip (Phase 6 & 7)

Lftw:
## What changed

Adds end-to-end support for **linear** and **mixed** feed polarization, previously gated behind `NotImplementedError`. 
Two coupled pieces:

**Phase 6 — simulation + forward Jones**
- `polrep` widened to `stokes`/`circ`/`lin`/`mixed` through `Array.obsdata` → `make_uvpoints` → `sample_vis`.
- `make_uvpoints`: per-station feed dispatch (`feed_dtype_for_polrep`); the four correlation sigmas are built by pairing each station's two feeds via `Array.sefd_for_feed`, replacing hardcoded R/L.
- `sample_vis`: `lin` branch mirrors `circ`; `mixed` samples Stokes and the caller converts per-baseline.
- New forward Jones layer in `pol_conventions.py`: `assemble_jones`, `field_rotation_matrix`, `apply_jones_to_coherency`, factored as **J = G·(I+D)·Φ** (Paper VII Eq. 4). `make_jones`/`make_jones_inverse`/`add_jones_and_noise`/`add_noise` are now polrep-agnostic (circ/lin/mixed).

**Phase 7 — uvfits load/save round-trip**
- `load_obs_uvfits`: per-station **POLTYA/POLTYB → `feed_type`**; native basis inferred from the feed-type set, not CRVAL3.
- `save_obs_uvfits`: `polrep_out` extended to lin/mixed; POLTYA/POLTYB emitted from `feed_type`.

**Docs:** `polarization_conventions.md` §9–12 rewritten (J = G(I+D)Φ, linear/mixed conventions).

## Why

Phase 6/7 of the mixed-pol track — the forward model and I/O prerequisites for joint polarimetric imaging on heterogeneous arrays (ALMA/VLA). Removes the lin/mixed simulation and loader gates.

## Design decisions worth discussing

- **Conversion in the caller (Option A):** Stokes→coherency for mixed baselines lives in a shared helper `pack_sampled_visibilities` (used by `Image.observe_same_nonoise` + `Movie`), keeping `sample_vis` a pure uv→vis sampler.
- **Φ is basis-dependent:** diagonal-phase `diag(e^{-iφ}, e^{+iφ})` for circular feeds, real rotation `[[cos,sin],[−sin,cos]]` for linear; `Φ_lin = F⁻¹Φ_circ F`. Absolute sign is convention-dependent (TMS §4.7) and matches our existing circular convention — confirmed by Rohan.
- **Native basis from the feed-type set, not CRVAL3** (for mixed, CRVAL3 is only a nominal 4-slot grid). CRVAL3/feed **disagreement** now raises loudly; non-unit-step CDELT3 is rejected.
- **`sefd_p1`/`sefd_p2`** replace whole-column `tarr['sefdr']` access — fixes a latent `TarrView` guard trip that blocked constructing *any* mixed `Obsdata`.

## Correctness

- Homogeneous-circular and stokes output stays **byte-identical** to base, including fixed-seed noise (`PYTHONHASHSEED=0`, ~1e-15) — pinned by regression tests.
- Corrupt→invert Jones round-trips to ~1e-9 for circ/lin/mixed; mixed pack→unpack at machine precision.
- Save→load→save round-trips lin+mixed (idempotent to the float32 uvfits floor ~3e-8, same as circ).

## How to test

\`\`\`bash
PYTHONHASHSEED=0 micromamba run -n jax-ehtim pytest \
  tests/test_mixedpol.py tests/test_obs_simulate.py tests/test_pol_conventions.py tests/test_image.py
\`\`\`
542 passed / 1 skipped locally against the current \`dev-backend-mixpol\` base.

## Intentionally gated (raise up front)

Hybrid `'rx'` feeds · non-circular `frcal=True, dcal=False` · `stokes → switch_polrep('mixed')` · `force_singlepol` on non-circ. Each raises a clear error rather than silently producing physically wrong results.

## Known open / out of scope (not blockers)

- CDELT3 positional-plane assumption not yet verified against a **real** mixed uvfits — need Iniyan's sample to confirm real STOKES/POLTY encoding.
- Pre-existing `obsdata.py:653` sorts `tarr` via whole-column `tarr['sefdr']` → breaks on lin/mixed `TarrView` (flagging, not fixing here).
- AIPS convention variance / hybrid `POLTYA=R, POLTYB=X` stations — open question.

## Not in this PR

- crc32 noise-seed fix → separate PR to `dev-backend` (`fix/stable-noise-seed`).
- Slot-variable rename (`rr_2d`/`llweight` → `p1p1`-style, ~211 edits) → deferred follow-up; behavior is test-pinned so it's independent.
- `caltable.py` for now (Phase 5 boundary — Jones layer owned here, calibration adapts it later).

Part of #212.

### DIFF
--- a/ehtim/const_def.py
+++ b/ehtim/const_def.py
@@ -86,6 +86,32 @@ VALID_FEED_TYPES = frozenset({
     'xr', 'xl', 'yr', 'yl',
 })
 
+# Canonical ordering of the two feeds of a station: R before L before X before
+# Y. In memory a station's feed_type is ordered ('lr' means p1 = L, p2 = R and
+# so p1p1vis is the LL product), but on disk a uvfits STOKES axis labels its
+# planes absolutely (CRVAL3 = -1 -> RR, LL, RL, LR). The two agree only when
+# every station's feeds are in canonical order, so uvfits I/O canonicalizes.
+FEED_CHAR_ORDER = 'rlxy'
+
+
+def canonical_feed_type(feed_type):
+    """Return (canonical_feed_type, was_reversed) for a two-character feed code.
+
+       Args:
+           feed_type (str): a member of VALID_FEED_TYPES, e.g. 'rl', 'lr', 'xr'
+
+       Returns:
+           (str, bool): the pair reordered to FEED_CHAR_ORDER, and whether that
+           reordering swapped the two feeds ('lr' -> ('rl', True)).
+    """
+    ft = str(feed_type).lower()
+    if ft not in VALID_FEED_TYPES:
+        raise ValueError(f"canonical_feed_type: {feed_type!r} is not one of "
+                         f"{sorted(VALID_FEED_TYPES)}")
+    if FEED_CHAR_ORDER.index(ft[0]) <= FEED_CHAR_ORDER.index(ft[1]):
+        return ft, False
+    return ft[1] + ft[0], True
+
 # Observation recarray datatypes
 # DTARR uses generic names primary because it is a single shared dtype across
 # all stations; legacy names are title aliases. Per-Obsdata/Caltable dtypes
@@ -169,8 +195,8 @@ DTSCANS = [('time', 'f8'), ('interval', 'f8'), ('startvis', 'f8'), ('endvis', 'f
 
 
 # TODO: the feed_dtype_for_polrep / feed_poldict / upgrade_* helpers below
-# should migrate to ehtim/observing/pol_conventions.py (created in
-# MixPol Phase 3) when Obsdata.switch_polrep is wired up to it.
+# should migrate to ehtim/observing/pol_conventions.py when
+# Obsdata.switch_polrep is wired up to it.
 
 def feed_dtype_for_polrep(polrep):
     """Return the DTPOL_* field-spec list for a given polrep."""
--- a/ehtim/const_def.py
+++ b/ehtim/const_def.py
@@ -194,10 +194,6 @@ DTCAL = DTCAL_CIRC  # legacy alias
 DTSCANS = [('time', 'f8'), ('interval', 'f8'), ('startvis', 'f8'), ('endvis', 'f8')]
 
 
-# TODO: the feed_dtype_for_polrep / feed_poldict / upgrade_* helpers below
-# should migrate to ehtim/observing/pol_conventions.py when
-# Obsdata.switch_polrep is wired up to it.
-
 def feed_dtype_for_polrep(polrep):
     """Return the DTPOL_* field-spec list for a given polrep."""
     try:
--- a/ehtim/io/load.py
+++ b/ehtim/io/load.py
@@ -33,6 +33,7 @@ import ehtim.movie
 import ehtim.obsdata
 import ehtim.observing
 import ehtim.vex
+import ehtim.warnings as ehw
 
 warnings.filterwarnings("ignore", message="Mean of empty slice")
 warnings.filterwarnings("ignore", message="invalid value encountered in true_divide")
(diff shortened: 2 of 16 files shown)


rohandahale (changes requested):
@Lftw Nice work on this! 

1) 'lr' and 'yx' are in `VALID_FEED_TYPES` and the loader accepts them, but the four planes are still assigned to p1p1/p2p2/p1p2/p2p1 by position, with nothing checking that the feed order agrees with CRVAL3. I took sample.uvfits, tagged the stations alternately L/R and R/L, and it loaded as polrep='mixed' with polbasis like ['rllr', 'rlrl', ...], a purely circular file, reinterpreted, and gives no warning. Simplest fix is probably to cross-check the parsed feed order against CRVAL3 and either canonicalize 'lr' -->'rl' (swapping the slots to match) or just reject reversed pairs with a clear message.
  
2) A blank POLTYB now fails where it used to load. If POLTYA='R' and POLTYB is empty, you get NotImplementedError: unsupported feed pair. The fallback at the top only fires when both columns are missing, so a blank string slips past it. Falling back to the default feed per station (with a warning) would fix it.

3) polrep='lin' doesn't work from a Stokes or circular source. singlepol_hand defaults to 'R' and the stokes-->lin branch requires 'X'/'Y', so load_uvfits(..., polrep='lin') and save_uvfits(stokes_obs, polrep_out='lin') both raise. That means the new lin write path only ever runs when the obs is already 'lin', which is what the round-trip test does. So CI stays green and the gap doesn't show. 

@achael: a mixed file is currently indistinguishable from a circular one, polrep_out='mixed' writes CRVAL3=-1, so the POLTY tags are the only thing saying it's mixed.

4) The mixed round-trip test only proves save and load agree with each other, which is unavoidable until there's an external reference, using _tarr(['xy','rl','rl']) so the first baseline is xyrl would at least exercise the polbasis swap through the IO path (it does work but untested). Both round-trips check *vis but not *sigma. And the two prints in load.py would be better as warnings.warn with the categories in ehtim/warnings.py, especially the one saying you got a different polrep than you asked for.

 5) Unrelated to your changes, but I noticed load.py:1345 has if force_singlepol in ['L' or 'LL'], which Python evaluates as ['L']. So 'LL' and 'RR' are ignored.

ehtim/io/load.py
```diff
@@ -1109,32 +1113,58 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
...
+    # maps to a 2-char feed_type code ('rl' circular, 'xy' linear, 'rx' hybrid,
+    # etc.).
+    def _feedchar(col, i):
+        if col is None:
+            return ''
+        p = col[i]
+        return (p.decode() if isinstance(p, bytes) else str(p)).strip().upper()
+
+    # When POLTY tags are absent, infer the default feed basis from the STOKES
+    # axis (CRVAL3) rather than blindly assuming circular: -5 (linear XX/YY/...)
```
achael: clarify the docstring: AIPS Memo 117 defines the STOKES axis CRVAL=1 (slots 1-4 ) for IQUV, CRVAL=-1 (slots -1 -- -4) as circular products, and CRVAL=-5 (slots -5--8) as linear products. 

we should note somewhere in our standards document that the latest revision to the AIPS Memo has replaced X&Y with V&H; i think we want to keep X and Y ourselves but we should write this down

ehtim/io/load.py
```diff
@@ -1109,32 +1113,58 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
...
+    poltya = poltyb = None
+    try:
+        poltya = hdulist['AIPS AN'].data['POLTYA']
+    except KeyError:
+        pass
+    try:
+        poltyb = hdulist['AIPS AN'].data['POLTYB']
+    except KeyError:
+        pass
+    if poltya is None and poltyb is None:
```
achael: raising Rohan's comment here: there should be a fallback if only one of POLTYA or POLTYB is undefined.

rohandahale (dismissed):
@Lftw  This looks great! The homogeneous vs  mixed split in the canonicalization is a better solution than what I suggested. Tests look good too.

One small thing: The FeedTagWarning docstring says a single blank tag gets completed from the known feed (R -> rl), but load.py raises instead, and the comment right there says not to complete it. You just need to update the docstring to match.

ehtim/io/load.py
```diff
@@ -1203,10 +1312,44 @@ def load_obs_uvfits(filename, polrep='stokes', flipbl=False,
...
     print("Number of uvfits Correlation Products:", num_corr)
+
+    # Consistency between the STOKES axis and the parsed feed types. (a) The
+    # axis naming (CRVAL3) must not contradict the feeds -- e.g. linear planes
+    # (-5) tagged with circular feeds. (b) The four correlation planes are read
+    # positionally (slot0..3, below); this is only valid for the standard
+    # unit-step axis, so reject a non-(-1)/(+1) CDELT3 rather than silently
+    # mis-slotting a reordered/rescaled STOKES axis.
+    crval3 = header['CRVAL3']
+    if crval3 == -5 and polrep_uvfits == 'circ':
```
achael, line 1380: Good to have these checks on CRVAL3 -- but if we infer 'mixed' from the station table, we should proceed also print a warning. We should warn that we have inferred mixed from the station table, but the STOKES axis would otherwise imply a circular (for CRVAL3==-1) or linear (for CRVAL5==-1) file.

ehtim/observing/pol_conventions.py
```diff
@@ -488,3 +488,171 @@ def apply_inverse_jones_to_coherency(V_obs, J1, J2):
...
+    ``fr_angle_D`` term in ``obs_simulate.make_jones``); it rotates the D-terms
+    by ``Phi_D^{-1}``. With ``fr_angle_D = 0`` this is the plain
+    ``J = G (I + D) Phi``.
+
+    Reproduces ``obs_simulate.make_jones``'s circular assembly exactly for
+    ``feed_type='rl'``, and extends to linear feeds (``'xy'``) via the linear
+    ``Phi`` in :func:`field_rotation_matrix`. Pure and broadcasting (no in-place
+    mutation), so it is safe for the JAX backend.
+
+    LIMITATION (ASSUMPTION, to verify; see docs/polarization_conventions.md sec 10):
```
achael, line 606: I think this is right but I want to return to it....

ehtim/const_def.py
```diff
@@ -169,8 +195,8 @@
 
 
 # TODO: the feed_dtype_for_polrep / feed_poldict / upgrade_* helpers below
-# should migrate to ehtim/observing/pol_conventions.py (created in
-# MixPol Phase 3) when Obsdata.switch_polrep is wired up to it.
+# should migrate to ehtim/observing/pol_conventions.py when
```
achael: i think switch_polrep is already wired to pol_conventions, no? i think it's fine to keep these here or move them, your call -- let's just delete this TODO if you agree

ehtim/observing/obs_simulate.py
```diff
@@ -345,11 +353,121 @@ def sample_vis(im_org, uv, sgrscat=False, polrep_obs='stokes',
...
 
+def pack_sampled_visibilities(obsdata, data, polrep):
+    """Write the four sampled visibility vectors from sample_vis into the
+    correct fields of an obsdata recarray.
+
+       Args:
+           obsdata (np.recarray): the data table to fill, in representation
+               `polrep`. For 'mixed' it must carry a populated 'polbasis' column.
+           data (list): the four visibility vectors returned by sample_vis. For
+               'stokes'/'circ'/'lin' these are the four basis correlations; for
```
achael: nitpicky, but i'd say "for circ/lin these are the basis correlations, for stokes/mixed these are I,Q,U,V visiblities"

achael (changes requested):
I'm so sorry for the long delay on this, but it looks fantastic. Really clean refactor and very readable. I just have a few very minor comments.

ehtim/io/load.py
achael: looks good

achael (approved):
looks good!

PR #318, opened 2026-07-30
Fix multifrequency polarimetric gradients (rho chain rule + alpha slot)

rohandahale:
Two bugs in the multifrequency polarimetric gradient, both pre-existing on `dev`, both silent: the run
converges, it just converges somewhere other than the objective it reports. Found while auditing the backend
ahead of the release.

## The spectral index was never fitted

`physical_grad_slots` builds its mask from the 4-wide Stokes block and knows nothing about multifrequency.
Under mf, `validate_params` only allows `pol='P'`, so Stokes I at the reference frequency is not a solved
DOF and the mask zeroes slot 0. But `mf_all_grads_chain` builds the spectral-index gradient out of exactly
that slot (`dfunc_dalpha = dfunc_dI * imvec * L`), so it came back identically zero and L-BFGS-B could never
move alpha.

Nothing else can fill that slot either: under mf, `validate_params` restricts the non-Stokes-I regularizers
to `REGULARIZERS_POL + REGULARIZERS_POLSPECTRAL`, so `l2_alpha` and friends are rejected outright and the
chi^2 term is the only possible source of gradient for alpha. With it zeroed, alpha had no gradient from
anywhere.

End to end on a two-frequency polarized observation, 200 iterations, true alpha = -0.7 starting from 0,
identical input with only the slot widening toggled:

| | specvec mean | specvec std | flux-weighted alpha |
|---|---|---|---|
| before | +0.00000 | 0.00000 | +0.0000 |
| after | -0.21414 | 0.42385 | **-0.6512** |
| truth | | | -0.7000 |

Before, alpha sat at its initial value to the last digit. After, it recovers the input spectral index to
about 7 per cent, at chi^2 0.97 and 0.96 on the two bands. (Flux-weighted because the faint pixels are
unconstrained and noisy; the unweighted mean is dominated by them.)

Under jax the same run always moved alpha, since autodiff never consulted the mask, so numpy and jax
disagreed on identical input.

## The rho chain rule used rho where it needs rho'

`multifreq_imager_utils.py:129-131`. The spectral expansion is on rho', not rho:

    log(rho') = log(rho0) + alpha_pol*L + beta_pol*L^2
    rho       = (rho'^-D + 1)^(-1/D)

so `d(rho')/d(rho0) = rho'/rho0` and the two spectral slots pick up `rho' * L^n`. The code used
`rhovec_cur`, the physical rho after the squashing transform, in all three. `rhovec_prime` was already
computed on line 111 and used only in `drho_drhoprime`. The old code's error factor is
`(1 - rho^D)^(1/D)`, which at the shipped `DD_RHOPOL=1` is `1 - rho`, so 20 to 45 per cent at ordinary
polarization fractions on `rho0`, `alpha_pol` and `beta_pol`. The fix is generic in `D`.

The Stokes-I block just above is correct, because I has no second transform layer between the coefficients
and the image. That is likely how the pol block came to be written this way.

## What changed

- `mf_all_grads_chain` uses `rhovec_prime` in the three rho-family slots.
- `physical_grad_slots` takes the multifrequency mask and fills the physical slot each solved spectral row
  chains through (rows 4,5 through I; 6,7 through rho; 8 through phi). `compute_chisqgrad_dict` is the only
  caller that passes it. Only slot 0 can currently be starved, since `compute_which_solve` hardcodes
  `do_rho = do_phi = 1` for every mf-pol run, but all three are handled symmetrically so that the first mode
  holding `rho0` or `phi0` fixed while solving `alpha_pol` or `rm` does not reintroduce this one row over.
- A comment on the pol regularizer branch explaining why its spectral rows stay at zero, which is correct
  and should not be "fixed" later: that regularizer is evaluated on the reference-frequency image and never
  runs through the mf chain, and the value side matches.

The widened slot does not leak anywhere it should not: `pack_imarr` drops unsolved slots, and `mcv_grad`,
`vcv_grad` and `polcv_grad` all read only `gradarr[1:4]`.

## Why it survived

No finite-difference test anywhere combined `mf=True` with a polarization mode. `test_gradients.py` had
`TTYPES = ["direct", "nfft"]` and no mf cases at all, the one mf gradient test asserted only `isfinite`, and
`test_objective_jax.py:283` has a comment saying as much.

## Tests

New section S7 in the canonical FD suite, committed first so it fails before either fix:

- `TestMfChainRule` finite-differences `mf_all_grads_chain` against `image_at_freq` using an arbitrary dense
  upstream gradient, so every coefficient slot is exercised. Same reason S6 uses a dense `grad_phys`: a real
  objective is blind to whichever slot it happens to zero. Stokes-I and full-pol cases, the Stokes-I one
  passing throughout as a control.
- `TestMfPolChisqGradient` drives the production composition through `compute_chisq_dict` and
  `compute_chisqgrad_dict` on a two-frequency polarized observation, comparing every slot
  `compute_which_solve` marks as solved.

Before the fixes: 0.47 max fractional error on the chain rule, 0.42 on the composed test at slot 1. After
the rho fix alone the composed test moved on to slot 4 at exactly 1.00, an analytic zero against a nonzero
finite difference, which is the second bug surfacing.

Also added, and the check most likely to catch the next one of these:

- **mf-pol cases in `test_objective_jax.py`.** `jax.grad` differentiates `image_at_freq` directly and never
  executes `mf_all_grads_chain`, so it is an independent opinion on the hand-written chain rule rather than
  a second reading of it. Reverting either fix fails it. It also means `use_jax=True` mf-pol runs were
  getting the correct gradient all along while numpy was not.
- Four unit tests of the slot widening in `test_imager_backend.py`, including the rows that are only
  latently at risk. `physical_grad_slots` had no test of this behaviour at all.

## How to test

    pytest tests/test_gradients.py -k "MfChainRule or MfPolChisq"

Full suite on a compute node, `-m "not gpu"`: **1911 passed, 1 skipped, 1 xfailed**, no failures.
(`tests/test_interactive.py` is deselected: it blocks on a plotting renderer on a headless node,
unrelated to this change. The 18 gpu-marked tests are run separately.)

## Noticed while in here, not fixed

The `REGULARIZERS_ALLFREQS_I` path is broken under mf-pol: `compute_reggrad_dict` passes a `(4, npix)` array
to `reggrad_flux`, which returns `np.ones(len(imvec))`, i.e. shape `(4,)`, and that feeds
`mf_all_grads_chain` as if it were four per-pixel Stokes rows. Silently wrong, reachable with `flux_mf` plus
mf-pol. Pre-existing and unrelated to these two fixes, so it wants its own PR.

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -610,13 +610,14 @@ def physical_grad_slots(pol_solve, transforms, mf_spectral_I=False):
     return mask
 
 
-# Rows 4 and 5 of a multifrequency imarr are the Stokes-I spectral index and curvature
-# (see multifreq_imager_utils.image_at_freq); the pol expansion terms follow at 6-9.
-MF_SPECTRAL_I_SLOTS = slice(4, 6)
+def _mf_solves_spectral_I(which_solve, mf):
+    """True when a multifrequency run solves for the Stokes-I spectral index or curvature.
 
-
-def mf_solves_spectral_I(which_solve, mf):
-    """True when a multifrequency run is solving for the Stokes-I spectral index or curvature.
+    Only the 10-row mf-pol layout needs this: there rows 4 and 5 are the Stokes-I spectral
+    index and curvature (see multifreq_imager_utils.image_at_freq), while the block gating
+    the chi^2 kernel is built from rows 0-3 alone, so the kernel never learns that a row it
+    cannot see depends on Stokes I. The 3-row Stokes-I mf layout runs an ungated kernel and
+    needs no widening.
 
     Parameters
     ----------
@@ -628,12 +629,10 @@ def mf_solves_spectral_I(which_solve, mf):
     Returns
     -------
     bool
-        Whether physical_grad_slots needs to request the Stokes-I gradient slot.
+        Whether physical_grad_slots should request the Stokes-I gradient slot.
     """
-    if not mf:
-        return False
     which_solve = np.asarray(which_solve)
-    return bool(np.any(which_solve[MF_SPECTRAL_I_SLOTS]))
+    return bool(mf and len(which_solve) == 10 and np.any(which_solve[4:6]))
 
 
 def make_initarr(image, mask, norm_init=False, flux=1,
@@ -1618,7 +1617,7 @@ def compute_chisqgrad_dict(imcur, dat_term_keys, config,
     pol_solve = _pol_solve_block(which_solve, pol)
     pol_grad_slots = physical_grad_slots(
         pol_solve, config.transforms,
-        mf_spectral_I=mf_solves_spectral_I(which_solve, mf))
+        mf_spectral_I=_mf_solves_spectral_I(which_solve, mf))
     # np.array((...)) below copies, so sharing zero_row across iterations is safe.
     zero_row = np.zeros(nimage)
     is_pol_mode = pol in POLARIZATION_MODES
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -559,7 +559,7 @@ def transform_gradients(gradarr, imarr, transforms, which_solve):
     return outarr
 
 
-def physical_grad_slots(pol_solve, transforms):
+def physical_grad_slots(pol_solve, transforms, mf_spectral_I=False):
     """Physical gradout slots the gradient kernels must fill, given the DOF mask.
 
     pol_solve is the 4-wide Stokes DOF mask (I, rho/m', chi, psi/v'); the chisq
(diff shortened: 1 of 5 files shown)


ehtim/imaging/multifreq_imager_utils.py
```diff
@@ -126,9 +126,11 @@ def mf_all_grads_chain(funcgrad, image_cur, mfarr, log_freqratio):
             # TODO: what to do about rho=0?
             drho_drhoprime = (rhovec_prime**(-1-DD_RHOPOL))*((1 + rhovec_prime**(-DD_RHOPOL))**(-1-1/DD_RHOPOL))
 
-            dfunc_drho0     = dfunc_drho * drho_drhoprime * rhovec_cur / rho0
-            dfunc_dalphapol = dfunc_drho * drho_drhoprime * rhovec_cur * log_freqratio
-            dfunc_dbetapol  = dfunc_drho * drho_drhoprime * rhovec_cur * log_freqratio * log_freqratio
+            # The expansion is on rho', not rho: log(rho') = log(rho0) + alpha_pol*L +
```
achael: dumb mistake, probably never caught because the test problem i played with when initially putting this together 2 years ago only did RM and didn't vary rho. 

Fix is correct, but we don't need the comment.

ehtim/imaging/imager_backend.py
```diff
@@ -572,6 +580,22 @@ def physical_grad_slots(pol_solve, transforms):
...
+    transforms : sequence of str
+        Active change-of-variables names; only 'mcv' and 'vcv' widen the mask.
+    mf_solve : sequence of int, optional
+        Full solved-DOF mask when this is a multifrequency run. Spectral rows reach the
+        objective only through a physical slot, so a solved spectral row needs that slot
+        filled even when the slot is not a DOF itself.
+
+    Returns
+    -------
+    mask : np.ndarray of int
```
achael: mask is also len 4, right? (same dimension as pol_solve). I'd note that.

ehtim/imaging/imager_backend.py
```diff
@@ -559,7 +559,15 @@ def transform_gradients(gradarr, imarr, transforms, which_solve):
...
-def physical_grad_slots(pol_solve, transforms):
+# Multifrequency pol imarr layout: 4 physical rows then 6 spectral ones. Each spectral row
+# reaches the objective only through the physical slot listed here.
+MF_POL_ROWS = 10
+MF_SPECTRAL_ROWS = {0: slice(4, 6),   # alpha, beta   -> I
+                    1: slice(6, 8),   # alpha_pol, beta_pol -> rho
+                    2: slice(8, 9)}   # rm            -> phi
+
+
+def physical_grad_slots(pol_solve, transforms, mf_solve=None):
```
achael: It looks like physical_grad_slots is only called in this module, and the input "pol_solve" is always just the first 4 entries of the "which_solve" array (through the helper function "_pol_solve_block". Then for the mf pol case below, we pass in "mf_solve=which_solve", which seems redundant. 

I would just rewrite this function to take the full "which_solve" array every time instead of "pol_solve" by default and "which_solve" only in the mf pol case, incorporate the helper function "_pol_solve_block" into the function body (since it is used nowhere else).

achael (changes requested):
Good catch. I suggested re-writing the "physical_grad_slots" function to be clearer -- since it is only called a few places i think it is better to change the arguments to better accommodate the mf pol case rather than adding a more confusing optional argument. 

Has the follow-up issue on the flux_mf gradient been looked into in another PR? If it's a relatively quick fix i'd suggest folding it in here.

PR #321, opened 2026-08-04
Run the jax tests in CI

rohandahale:
CI runs `pytest -m "not jax"`, which deselects 153 of 1935 tests: the entire JAX backend.
`ehtim/imaging/survey_gpu.py` has no CI coverage at all as a result, and the modules that back the JAX
path are covered only incidentally, through the numpy tests that happen to import them. Statement
coverage before and after:

    survey_gpu.py     0%  ->  100%
    optimizers.py    19%  ->   88%
    backends.py      50%  ->   97%
    sharding.py       0%  ->    0%   (still gpu-marked, see below)

`origin/dev` runs plain `pytest --cov`, so the deselection is new on this branch and would otherwise ship
with the release.

It reads as a GPU exclusion, but it is not one. JAX runs perfectly well on CPU, and the job already sets
`JAX_PLATFORMS: cpu`. Of the 153, 18 are `gpu`-marked, and `-m "not gpu"` keeps exactly those out and runs
the other 135.

Nothing new to install: `pip install -e ".[dev]"` already pulls `jax`, `jax-finufft` and `optax`.

## What this costs

Measured on 4 cores, single CPU device, matching the runner's configuration:

    -m "not jax"       (what CI runs today)   1773 tests    498s
    -m "jax and not gpu" (what this adds)      135 tests     66s

CI has since run this branch, so the real numbers replace the projection:

    test (3.11)   22m20s before  ->  22m08s after
    test (3.12)   23m59s before  ->  25m53s after

Both green. The added cost is inside the runner's own noise band (two baseline runs of the *unchanged*
command differ by more than three minutes), but 3.12 now lands at 25m53s.

That is why `timeout-minutes` goes 30 to 45 in the same PR: 25m53s leaves barely four minutes under the
old cap, which one slow runner would eat. The job was already close to the cap before this change, so the
bump is arguably overdue independently. The timeout is a cap rather than a wait, so raising it costs
nothing when the job behaves.

All 135 pass on CPU: **135 passed, 1800 deselected, in 66.43s**, with a single device and `JAX_PLATFORMS=cpu`
set exactly as the workflow sets it. Verified with one device specifically, since a stray
`--xla_force_host_platform_device_count` would not be present on the runner.

## What is still not covered

`tests/test_sharding_jax.py` is marked `[jax, gpu]` at module level, so `sharding.py` stays at zero CI
coverage even after this change, as the table above shows. That is worth fixing separately: `build_mesh` already takes a `devices=`
kwarg, and with `--xla_force_host_platform_device_count` JAX will present several addressable CPU devices,
so the sharding math is testable without hardware. A quick check of the baseline case on 2 fake CPU devices
matched single-device to 8.6e-16 and the NumPy analytic gradient to 1.3e-15, closures included, in about
3 seconds.

Not done here because #320 already edits that file and I would rather not create a conflict between two
open PRs over a marker line. It follows once #320 lands.

## How to test

    pytest -m "not gpu"

which is the new CI invocation. Run on a compute node: **1908 passed, 1 skipped, 18 deselected, 1 xfailed**,
no failures. 1908 is the 1773 that run today plus the 135 this adds, and the 18 deselected are the
gpu-marked ones.

On a compute node add `--ignore=tests/test_interactive.py`: it blocks on a plotting renderer there, while
the runner is headless and falls back to Agg, which is why CI runs it today.

Two adjacent fixes ride along, both consequences of CI now depending on jax. `pyproject.toml` gains a
version floor on `jax` (its neighbours already had floors, and CI resolves with no lockfile), and the
`jax` marker's help text no longer claims CI deselects it. The `L-BFGS-B` recovery floor in
`tests/test_chisq_jax.py` also widens: it scores ~0.97 and asserted 0.95, and #232 already had to drop an
e2e threshold from 0.90 to 0.80 when the runner's scipy and BLAS gave 0.867 against 0.97 locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
       - run: pip install -e ".[dev]"
-      - run: pytest -m "not jax" --cov --cov-branch --cov-report=xml
+      # jax runs on CPU here (JAX_PLATFORMS above); only the tests needing real GPUs are skipped
+      - run: pytest -m "not gpu" --cov --cov-branch --cov-report=xml
       - uses: codecov/codecov-action@v4
         if: matrix.python-version == '3.12'
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
     strategy:
       matrix:
         python-version: ["3.11", "3.12"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,12 +46,12 @@ dev = [
     "pandas",
     "paramsurvey",
     "plotly",
-    "jax",
+    "jax>=0.7",
     "jax-finufft>=1.3",
     "optax>=0.2",
 ]
 gpu = [
-    "jax[cuda12]",
+    "jax[cuda12]>=0.7",
     "jax-finufft>=1.3",
     "optax>=0.2",
 ]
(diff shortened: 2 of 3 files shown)


achael (approved):
looks good

PR #322, opened 2026-08-04
Refuse jax and sharded runs on unsupported transforms and data terms

rohandahale:
Not every transform and data term has a jax kernel. Where one is missing the run does not produce wrong
numbers, it dies somewhere inside jax with an error that describes an internal invariant rather than
anything the caller did. This is a diagnostics PR: same set of working configurations, but the
unsupported ones now say what is unsupported and what to use instead.

What a user sees today, all reproduced on a compute node before writing anything:

    ttype='fast' + use_jax=True     TracerArrayConversionError: The numpy.ndarray conversion method
                                    __array__() was called on traced array with shape float32[32,32]
    cphase_diag  + use_jax=True     TypeError: Error interpreting argument to stage as a JAX value
    logcamp_diag + use_jax=True     TypeError: (same)
    sharded baseline, ttype='fast'  IndexError: tuple index out of range
    build_mesh() with no GPU        RuntimeError: Unknown backend: 'gpu' requested
    run_survey_gpu, 2 observations  KeyError: 'vis_0_0'

Controls run in the same session: `direct` + `use_jax` images fine, and `cphase_diag` on numpy converges
normally. So nothing here is a numerical fix.

## Where the guard goes, and what it keys on

`check_jax_supported(ttype, dat_terms)` in `imager_backend.py`, called from `make_image` next to the
existing `shard` refusal and from the three backend builders (`_prepare_jax_loss`, the survey, and the
sharded one), so the factories are guarded too and not only the `make_image` route.

The two failure modes are not the same shape. `ttype='fast'` has no jaxified kernels: no `chisq_*_fft`
dispatches through `array_namespace` and `fft_imvec` is hard numpy, so it fails once a value is traced.
The `_diag` terms fail earlier and never reach a kernel: their `data` and `sigma` come back as ragged
`dtype=object` arrays, which `device_put` rejects outright with `Dtype object is not a valid JAX array
type`. `logamp` is numpy-only too and is refused for the same reason, though the Imager never routes to
it.

The part worth reviewing is what the check is keyed on:

```python
if use_jax or shard or classify_optimizer(optimizer) == 'optax':
```

not `use_jax` alone, and it runs before the first iteration is printed rather than after, so an
unsupported run no longer prints a bogus `Imager run 1` header and a `J:` line before refusing.
`use_jax` is not what decides whether jax runs. `make_image` picks its builder by
optimizer class, and `build_vg_ondevice` builds a jax objective whether or not the flag is set, so
`optimizer='optax-lbfgs'` with `ttype='fast'` reaches the tracer error with `use_jax=False`. Sharding
always goes through the same path. There is a test for exactly this, and it is the one that fails if the
guard is written the obvious way.

## The other three

- **Baseline sharding, `ttype='fast'`.** The `NotImplementedError` at the bottom of that branch is
  unreachable: 'fast' hands over a tuple, so the `isinstance(A, (tuple, list))` branch is taken first and
  the run dies in `_pad_rows`. Test `config.ttype` at the top of the branch instead.
- **`build_mesh` with no GPU.** It defaults to `jax.devices("gpu")`, so a CPU box gets jax's own backend
  error. It now points at the `mesh=` kwarg, which is the actual way to shard over chosen devices.
- **`run_survey_gpu` with more than one observation.** The survey keys data terms by bare name; with
  `n_obs > 1` they arrive already suffixed and get suffixed again. This one is a keying bug rather than a
  design limit: stripping the suffix looks sufficient, and an external review got a two-observation survey
  running end to end that way. Multi-observation surveys have no test coverage, so this refuses clearly
  and carries a TODO rather than shipping an unverified path.

## What is still not guarded

Two survey `regparam_grid` sweeps turn `RegParams` fields into tracers and fail inside the regularizer
rather than at a boundary the guard can see: sweeping `flux` with `tvlog` / `tv2log`, and sweeping
`beam_size` with any tv-family term. Both fail loudly, so they are the same class of diagnostics gap as
the rest of this PR rather than a correctness one, but they need the regularizers themselves to change and
are not in scope here.

## Tests

`tests/test_jax_guards.py`, 24 tests, committed before the fixes. `TestCheckJaxSupported` is pure python
and carries no jax mark, so those 8 keep running even when the jax tests are deselected; the classes that
build imagers are marked individually. Every refusal is paired with a control
that must still be accepted, because a guard that is too broad is its own bug: `fast` and `cphase_diag`
both still image on numpy, `direct` + `use_jax` still runs, and the survey still runs on one observation.

Each of the five fixes was reverted in turn and the corresponding test confirmed to fail:

    remove the check_jax_supported call            -> imager guard tests fail
    key the guard on use_jax alone                 -> the optax-path test fails
    drop the ttype check in baseline sharding      -> sharding test fails
    restore build_mesh's bare jax.devices("gpu")   -> build_mesh test fails
    drop the survey n_obs guard                    -> survey test fails

The `build_mesh` test monkeypatches `jax.devices` rather than skipping when a GPU is present, so it runs
on GPU boxes too instead of only in CI. The sharding test builds a one-device mesh, so it needs no GPU
either: the transform check fires before any padding.

## How to test

    pytest tests/test_jax_guards.py

Full suite on a compute node, `-m "not gpu"`: **1932 passed, 1 skipped, 1 xfailed**, no failures. That
is the 1908-test baseline plus the 24 added here. (`tests/test_interactive.py` deselected: it blocks on a
plotting renderer on a headless node, unrelated to this change.)

### DIFF
--- a/ehtim/imaging/imager_backend.py
+++ b/ehtim/imaging/imager_backend.py
@@ -56,6 +56,12 @@ REGULARIZERS_ISPECTRAL = REGULARIZERS_SPECIND + REGULARIZERS_CURV
 REGULARIZERS_POLSPECTRAL = REGULARIZERS_SPECIND_P + REGULARIZERS_CURV_P + REGULARIZERS_RM + REGULARIZERS_CM
 REGULARIZERS_SPECTRAL = REGULARIZERS_ISPECTRAL + REGULARIZERS_POLSPECTRAL
 
+# What the jax backend covers. 'fast' has no jaxified kernels (no chisq_*_fft dispatches
+# through array_namespace and fft_imvec is hard numpy), and the _diag closures carry a
+# 2-tuple of per-scan matrices that jax cannot bind as an argument.
+JAX_TTYPES = ['direct', 'nfft']
+JAX_UNSUPPORTED_DATATERMS = ['cphase_diag', 'logcamp_diag']
+
 # Default initial-polarization parameters used when init image has no Q/U/V.
 MEANPOL_INIT = 0.2     # mean polarization fraction
 SIGMAPOL_INIT = 1.e-2  # perturbation scale
@@ -1978,6 +1984,37 @@ def compute_objective_grad(imvec, initvec, config,
     grad = datterm + regterm
     grad = transform_gradients(grad, imcur_prime, transforms, which_solve)
     return pack_imarr(grad, which_solve)
+
+
+def check_jax_supported(ttype, dat_terms):
+    """Check that the jax backend covers this transform and these data terms.
+
+    Call before building any jax objective. Without it the run fails deep inside jax with a
+    tracer or argument-binding error that says nothing about the cause, since the missing
+    kernels fall back to numpy on traced values.
+
+    Parameters
+    ----------
+    ttype : str
+        Transform type: 'direct', 'nfft' or 'fast'.
+    dat_terms : iterable of str
+        Names of the active data terms.
+
+    Raises
+    ------
+    ValueError
+        If the transform or any data term has no jax kernel.
+    """
+    if ttype not in JAX_TTYPES:
+        raise ValueError(
+            f"ttype={ttype!r} has no jax kernels; use ttype='nfft' (or 'direct') to run on "
+            f"jax, or drop use_jax and the optax optimizers to image it on numpy.")
+    unsupported = sorted(set(dat_terms) & set(JAX_UNSUPPORTED_DATATERMS))
+    if unsupported:
+        raise ValueError(
+            f"data terms {unsupported} have no jax kernels; use their undiagonalized "
+            f"counterparts ('cphase', 'logcamp') to run on jax, or drop use_jax and the "
+            f"optax optimizers to image them on numpy.")
 
 
 def _place_jax_arrays(data_tuples, priorvec, initvec, device=None):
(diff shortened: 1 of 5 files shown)


ehtim/imaging/imager_backend.py
```diff
@@ -56,6 +56,12 @@
 REGULARIZERS_POLSPECTRAL = REGULARIZERS_SPECIND_P + REGULARIZERS_CURV_P + REGULARIZERS_RM + REGULARIZERS_CM
 REGULARIZERS_SPECTRAL = REGULARIZERS_ISPECTRAL + REGULARIZERS_POLSPECTRAL
 
+# What the jax backend covers. 'fast' has no jaxified kernels, so it dies once a value is
+# traced; the _diag closures never reach a kernel at all, since their ragged dtype=object
+# data is rejected by device_put. 'logamp' is numpy-only but is a valid dispatch key.
+JAX_TTYPES = ['direct', 'nfft']
```
achael: side note - i think we should just go ahead and delete 'fast' throughout before the next release. it's dead weight at this point.

ehtim/imaging/imager_backend.py
```diff
@@ -56,6 +56,12 @@
...
+JAX_UNSUPPORTED_DATATERMS = ['cphase_diag', 'logamp', 'logcamp_diag']
```
achael: what was the reason we don't have jax versions for these?

ehtim/imaging/sharding.py
```diff
@@ -212,6 +239,8 @@ def loss(x, aux):
                         datterm = datterm + dat_term[dname] * (chi2[key] * correction[key] - 1.0)
                 return datterm + regterm_of(imcur, aux["prior"])
         else:
+            # only 'direct' reaches here: check_jax_supported above rejects 'fast', whose
```
achael: why not nfft?

achael (approved):
looks good, just a few questions.

PR #323, opened 2026-08-04
Keep the log regularizers finite on a partial embed mask

rohandahale:
`reg_tvlog` and `reg_tv2log` take the log of the embedded image, so pixels outside the embed mask need a
strictly positive fill. They fill with `epsilon_tv`, which defaults to 0, so on any partial mask the value
is `inf` or `nan` and the gradient carries dozens of non-finite components. End to end that aborts the
optimizer on its first evaluation: `J: nan`, `ABNORMAL`.

**This fires at default settings.** `compute_embed` is `imvec > clipfloor`, a strict inequality, so the
mask is partial whenever the prior has exact-zero pixels, with `clipfloor` left at its default of 0. A
tophat prior, a `threshold(fill_val=0)` prior, or a wide-FOV Gaussian whose wings underflow all qualify.
I originally described this as reachable through `Imager(clipfloor > 0)`, which understated it.

Long-standing, not new: `epsilon_tv` has defaulted to `0.` in `imager_utils.py` since 2020 (`ca746f1`,
with the comment `#TODO: epsilon is 0 by default for backwards compatibility`).

The comment above these functions claimed the opposite, that they "use clipfloor=epsilon_tv (not the
default 0) so the log transform stays defined". `git blame` lands on me: I wrote it in #238, and it was
wrong the day it was written.

## Why not simply make the fill a small positive number

Because the value comes back finite and wrong, which is harder to notice than the crash. `epsilon_tv`
smooths the TV square root; 0 is a legitimate setting for that, which is what #311 established when it
made the spectral TV NaN-safe *at* `epsilon_tv=0` rather than changing the default. A fill far below the
image scale puts a large fake edge along the whole mask boundary, and the regularizer measures that.

On the committed fixture (`make_asym_image(32, 48)`), half the pixels masked out, against the value the
same image gives with no mask at all:

    tvlog    unmasked reference  0.37590   ratio 1.000
             fill = 1e-12        1.62333   ratio 4.318
             fill = 1e-8         1.00749   ratio 2.680
             fill = flux/npix    0.27495   ratio 0.731

    tv2log   unmasked reference  0.02503   ratio 1.000
             fill = 1e-12        0.37977   ratio 15.17
             fill = 1e-8         0.14517   ratio 5.800
             fill = flux/npix    0.01789   ratio 0.715

## The fix

One helper, `_embed_for_log`, used by all four sites (two values, two gradients), filling at `flux / npix`:
the mean pixel value, which is the scale these functions already normalize the log by
(`logflux = npix * |log(flux / npix)|` two lines down). It comes from kwargs rather than from `imvec`, so
it adds no gradient path, and it is scale-free, so it needs no new constant.

`epsilon_tv` is unchanged and still reaches `reg_tv` / `reg_tv2` through `log_kwargs`.

Ratios against the unmasked value climb back towards 1 as the mask grows, which is the property worth
having: it means the fill sits at the image's own scale rather than dominating it.

    tvlog    keep 30% 0.733   40% 0.729   50% 0.731   60% 0.780   70% 0.841   90% 0.992
    tv2log   keep 30% 0.722   40% 0.716   50% 0.715   60% 0.719   70% 0.732   90% 0.849

## What this does not fix

Worth being explicit, because the fill improves the boundary term without removing it. Measured against a
reference that sums only over edges with both ends inside the mask, at a 50% mask the remaining boundary
contribution is about 10x the real interior structure for `tvlog` and about 700x for `tv2log`, down from
roughly 60x and 15000x with a 1e-12 fill. So this is a large improvement and not a clean solution.

Two consequences a reviewer should weigh:

- On a compact core with a faint halo, `flux / npix` sits above most real pixels (98 to 99 per cent of
  them on such images), so the fill is a bright rim rather than a floor and the partial-mask value can
  slightly exceed the unmasked one, up to about 1.09. The gradient along the rim is one-signed. At strong
  `tvlog` weights that shows up as a brighter boundary ring and a worse data fit.
- Filling at `clipfloor` instead avoids the one-signed rim, but `clipfloor` is 0 by default, which is
  exactly the case that produced the bug, so it would need a `clipfloor if clipfloor > 0 else flux/npix`
  rule plus a new `RegParams` field.

The principled fix is to drop the fill and exclude cross-boundary edges from the TV sum, mirroring the
masking pattern used for `scm` / `reg_cm` in task 2.27. That is a bigger change than making the thing
finite, so it is a follow-up rather than this PR.

Also changed, and larger than it sounds: with `epsilon_tv > 0` **and** a partial mask, the old fill was
`epsilon_tv * |N(0, 1)|`, which was finite but arbitrary. Those runs change value here by a factor of
roughly 0.007 to 0.6.

## Tests

`tests/test_regularizers.py` (four per regularizer plus three on the helper) and a numpy/jax parity test in
`tests/test_regularizers_jax.py`, committed before the fix. The existing `reg_setup` fixture uses
`mask = imvec > 0`, a full mask, which is why this path had no coverage.

Reverting the fix three ways:

    revert the fill to epsilon_tv (the original bug)   9 failed
    the naive fix, a tiny positive floor (1e-12)       4 failed
    fill 1e-6 below the image scale                    4 failed

The middle row is the point: a naive fix is finite, so a finiteness-only suite would accept it.

The regularizer-level tests alone were not enough, and an external review found the hole: because the
fixture has `flux` near 1 and `npix` about twice `nmask`, a fill of `1/npix` (ignoring `flux` entirely) or
of `flux/nmask` reproduced every one of them. `test_fill_is_the_mean_pixel_value` now pins the fill
directly at two different `flux` values, and both of those wrong implementations fail.

Two test claims were also weakened to what actually holds. `few < many <= reference` compares widely
separated mask sizes because the ratio is not monotone in between (it dips at 40 per cent above). And the
`<= reference` bound is a regression guard against the tiny-fill mode on this fixture, not a universal
invariant: it is violated by this PR's own fill on compact-core images, which the test now says.

The partial-mask gradient sample was widened from 15 pixels to those plus a random spread, since this is
the only partial-mask gradient coverage in the suite (`test_gradients.py` uses full masks throughout).
The old comment calling them "best-conditioned" was backwards: the gradient carries a `1/I`, so the
largest components are the faintest pixels.

## Not in this PR

- `randomfloor` is inert everywhere: the fill is `clipfloor * |N(0, 1)|` and the remaining callers leave
  `clipfloor` at 0. That comment fix sits against the lines #322 adds, so it goes in the docstring pass.
- `reg_tv` pads with `constant_values=0`, and `reg_tvlog` feeds it the *log* image, so the pad asserts
  1 Jy just outside the grid. On this fixture that pad accounts for most of the full-mask value. Same
  class of bug, larger, and it deserves its own issue.
- `reg_msimple` is `-inf` with NaN gradient components at zero-polarization pixels under polcv.

## How to test

    pytest tests/test_regularizers.py tests/test_regularizers_jax.py

Full suite on a compute node, `-m "not gpu"`: **1920 passed, 1 skipped, 1 xfailed**, no failures.


CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
```
achael: remove CHANGELOG from the commit

ehtim/imaging/imager_utils.py
```diff
@@ -1832,40 +1832,76 @@ def reggrad_tv(imvec, mask, **kwargs):
...
-    epsilon = kwargs.get('epsilon_tv', 0.)
-    if np.any(np.invert(mask)):
-        imvec = embed(imvec, mask, clipfloor=epsilon, randomfloor=True)
-    # parameters and normalization -- update kwargs
-    nx, ny = kwargs['xdim'], kwargs['ydim']
+    nx, ny, psize = kwargs['xdim'], kwargs['ydim'], kwargs['psize']
     flux = kwargs['flux']
     npix = nx * ny
+    if np.any(np.invert(mask)):
+        imvec = embed(imvec, mask, clipfloor=flux/npix)
```
achael: add a check that this clipfloor value is >=0 (it should be, but just in case)

ehtim/imaging/imager_utils.py
```diff
@@ -1832,40 +1832,76 @@ def reggrad_tv(imvec, mask, **kwargs):
...
+    # parameters and normalization
+    beam_size = kwargs.get('beam_size') or psize
+    epsilon = kwargs.get('epsilon_tv', 0.)
     logflux = npix * np.abs(np.log(flux / npix))
-    log_kwargs = dict(kwargs)
-    log_kwargs['flux'] = logflux
-    return reg_tv(xp.log(imvec), np.ones(imvec.shape, dtype=bool), **log_kwargs)
+    norm = logflux * psize / beam_size if kwargs.get('norm_reg', True) else 1
+    # compute TV of the log image
```
achael: good catch. Would porting this exact behavior over to standard tv,tv2 meaningfully change behavior? In a case where the edge pixels are zero they should be near-identical, and I think this version with edge padding is strictly better for all versions. 

So I would consider re-writing standard tv/tv2 and revert to the log versions calling them. It's ok if they are not byte-identical to the existing versions as long as they pass the finite diff tests.

ehtim/imaging/imager_utils.py
```diff
@@ -1832,40 +1832,76 @@ def reggrad_tv(imvec, mask, **kwargs):
...
+    im = xp.log(imvec).reshape(ny, nx)
+    mask2d = mask.reshape(ny, nx)
+    impad = xp.pad(im, 1, mode='edge')
+    maskpad = np.pad(mask2d, 1, mode='edge')
+    m_l1 = np.roll(maskpad, -1, axis=0)[1:ny+1, 1:nx+1]
```
achael: add short comment explaining the roll of the shifted masks

ehtim/imaging/imager_utils.py
```diff
@@ -1924,36 +1960,63 @@ def reggrad_tv2(imvec, mask, **kwargs):
...
+    if np.any(np.invert(mask)):
+        imvec = embed(imvec, mask, clipfloor=flux/npix)
+    # parameters and normalization
+    beam_size = kwargs.get('beam_size') or psize
     logflux = npix * np.abs(np.log(flux / npix))
-    log_kwargs = dict(kwargs)
-    log_kwargs['flux'] = logflux
-    return reg_tv2(xp.log(imvec), np.ones(imvec.shape, dtype=bool), **log_kwargs)
+    norm = psize**4 * logflux**2 / beam_size**4 if kwargs.get('norm_reg', True) else 1
+    # compute TV2 of the log image. Each difference belongs to the pixel it is
```
achael: same comment as above for tv

ehtim/imaging/multifreq_imager_utils.py
```diff
@@ -251,13 +255,17 @@ def reg_tv_spec(imvec, mask, **kwargs):
     epsilon = kwargs.get('epsilon_tv', 0.)
     norm = len(imvec) * psize / beam_size if kwargs.get('norm_reg', True) else 1
     im = imvec.reshape(ny, nx)
-    impad = xp.pad(im, 1, mode='constant', constant_values=0)
-    im_l1 = xp.roll(impad, -1, axis=0)[1:ny+1, 1:nx+1]
-    im_l2 = xp.roll(impad, -1, axis=1)[1:ny+1, 1:nx+1]
+    mask2d = mask.reshape(ny, nx)
```
achael: same comment as above for log tv

achael (changes requested):
looks good. I think edge padding for all tv regularizers is probably strictly better than zero padding. So i'd suggest exploring back-porting that behavior to the standard tv, tv2, ptv regularizers as well, then reverting to having the log, spec reg terms calling them. It's ok if they are not byte-identical to the existing versions as long as they pass the finite diff tests. But let me know if you disagree.

PR #254, opened 2026-05-18
Phase 1: mixed-pol schema additions and warnings module

achael:
## Summary

Implements Phase 1 of [`obsdata_mixedpol_plan_v2.md`](../obsdata_mixedpol_plan_v2.md): schema layer only. Behavior on circular-feed inputs is **numerically bit-identical** to pre-Phase-1.

- **`ehtim/warnings.py`** (new): `MixedPolConventionWarning`, `MixedPolClosureSkipWarning`. No emission sites yet — wired up in later phases.
- **`ehtim/const_def.py`**:
  - **DTARR**: 12 fields, with generic names primary (`sefd_p1`, `sefd_p2`, `d_p1`, `d_p2`) and legacy names as title aliases. New `feed_type` field (`'U2'`). Generic-primary inversion is deliberate — DTARR is a single dtype shared across all stations, so the title-alias mitigation alone isn't enough on heterogeneous arrays.
  - **DTPOL_CIRC**: gains title aliases (`p1p1vis ↔ rrvis`, etc.). Physical names remain primary so existing code (e.g. `obs.data.dtype.names[-4:]`) is unchanged.
  - **DTPOL_LIN, DTPOL_MIXED**: new. `DTPOL_MIXED` carries a per-row `polbasis` field (populated in Phase 3a).
  - **DTCAL_CIRC, DTCAL_LIN**: new, with D-term fields (Phase 4 reads `data[site]['d_p1']`). Legacy `DTCAL` kept as an alias of `DTCAL_CIRC`.
  - **Helpers**: `feed_poldict`, `feed_dtype_for_polrep`, plus `upgrade_tarr` / `upgrade_dtpol_circ` / `upgrade_dtcal_circ` for silent schema migration. TODO note marks these for migration to `pol_conventions.py` when that module lands in Phase 3b.
- **Schema-upgrade plumbing** in `Array.__init__/__setstate__`, `Obsdata.__init__/__setstate__`, `Caltable.__init__/__setstate__`: legacy recarrays are upgraded silently. DTPOL_CIRC upgrade is a zero-copy view-cast; DTARR and DTCAL are allocate-and-copy (new fields).
[...]

PR #331, opened 2026-08-12
Caltable: split gains and D-terms into two tables

Lftw:
## What and why

D-terms used to be welded into the gain table, sharing one time column, so a static leakage value had to be duplicated across every gain row. `Caltable` now holds two tables per site on independent time grids:

```python
ct.gains[site]    # DTCAL:   (time, rscale, lscale)
ct.dterms[site]   # DTDTERM: (time, dr, dl)   -- {} if there is no leakage
ct.data           # alias for gains, kept for backward compatibility
```

A one-row table is how a time-constant D-term is stored.

## Since the last review

- `split_dtcal` -> `upgrade_caltable`: detects feed basis from field names, upgrades legacy/solver shapes, raises on anything it can't interpret. The short-lived combined gain+D-term format is no longer read at all, per the 08-20 call.
- `dterms` kwarg renamed `dtermdict`, moved next to `tarr`.
- `save_caltable`/`save_txt` take `overwrite=False`; refuse to clobber existing files unless asked, clean up stale gain and D-term files on save.
- `DTDTERM_LIN` gained `dx`/`dy` aliases to match `DTCAL_LIN`.
- Non-dict `datadict` raises instead of being stored silently.
- Comments trimmed; the 1-2 line rule is now in CLAUDE.md.
- NOTE + TODO at the top of caltable.py: no D-term table means fall back to tarr D-terms; resolving both-defined is a follow-up PR.

## Bugs found along the way

- Single-row gain files failed to load (`ndmin=2`)
- List/0-d construction crashes in the solver caldict path, since #254
- `merge` aliased one side's gain arrays, so mutating the result wrote back into the input
- Byte-swapped legacy tables were reinterpreted instead of converted
- A leakage-only site lost its D-terms on save (`array.txt` came from the wrong tarr)
- `pad_scans` recast linear-feed gains as circular
- `scripts/imaging.py`, broken by #254, works again

## Deferred (unreachable via the current API)

- `applycal` doesn't apply D-terms yet (Part A)
- `DTDTERM_LIN` doesn't survive save/load (needs the basis dispatch Part A adds anyway)
- `plot_dterms` still plots tarr; warns when `self.dterms` disagrees with it
- Two-sided D-term merge raises: leakage composes as a Jones product, not a gain-style multiply, and needs an order convention first

399 tests pass. Sections 1-10 of `test_caltable.py` are unmodified except `_unity_caltable`, which is the no-regression contract for applycal, save/load, and the transforms.

### DIFF
--- a/ehtim/caltable.py
+++ b/ehtim/caltable.py
@@ -187,6 +187,11 @@ class Caltable:
                     show=True, grid=True, export_pdf=""):
         """Make a plot of the D-terms.
 
+           Plots the leakage recorded in the array table (tarr), one point per
+           site. The cal table's own D-term table (self.dterms) is not read
+           here: it can hold several rows per site, which a single point per
+           site cannot show.
+
            Args:
                sites (list) : list of sites to plot
                label (str) : title for plot
@@ -710,6 +715,10 @@ class Caltable:
 
     def save_txt(self, obs, datadir='.', sqrt_gains=False):
         """Saves a Caltable object to text files in the given directory
+
+           Gains and D-terms go to separate per-site files; see save_caltable
+           for the layout.
+
            Args:
                obs (Obsdata): The observation object associated with the Caltable
                datadir (str): directory to save caltable in
@@ -799,6 +808,12 @@ class Caltable:
 
 def load_caltable(obs, datadir, sqrt_gains=False):
     """Load apriori Caltable object from text files in the given directory
+
+       Reads the per-site gain files, and the per-site D-term files beside them
+       when present -- a directory written before D-terms had an on-disk slot
+       simply loads with no leakage. Gain files are what makes a directory a
+       cal table: with none, this returns False even if D-term files are there.
+
        Args:
            obs (Obsdata): The observation object associated with the Caltable
            datadir (str): directory to save caltable in
@@ -889,6 +904,15 @@ def load_caltable(obs, datadir, sqrt_gains=False):
 
 def save_caltable(caltable, obs, datadir='.', sqrt_gains=False):
     """Saves a Caltable object to text files in the given directory
+
+       Three kinds of file: array.txt for the telescope array, one
+       '<source>_<site>.txt' per site of gains (headerless, five columns:
+       time_mjd rre rim lre lim), and one '<source>_<site>_dterms.txt' per site
+       that carries leakage. The gain format is unchanged and stays headerless;
+       the D-term files, being new, open with a version line. A site with no
+       D-terms gets no D-term file, and sqrt_gains is a gain convention that
+       does not touch them.
+
        Args:
            obs (Obsdata): The observation object associated with the Caltable
            datadir (str): directory to save caltable in
--- a/ehtim/caltable.py
+++ b/ehtim/caltable.py
@@ -149,7 +149,7 @@ class Caltable:
         if isinstance(datadict, dict):
             self.gains = {}
             for site, d in datadict.items():
-                site_gains, site_dterms = ehc.split_dtcal(d)
+                site_gains, site_dterms = ehc.upgrade_caltable(d)
                 self.gains[site] = site_gains
                 if site_dterms is not None:
                     self.dterms[site] = site_dterms
@@ -186,7 +186,7 @@ class Caltable:
             normalized = {}
             reshaped = False
             for site, table in datadict.items():
-                site_gains, site_dterms = ehc.split_dtcal(table)
+                site_gains, site_dterms = ehc.upgrade_caltable(table)
                 if site_dterms is not None:
                     raise TypeError(
                         f"data is the gain table, but the table for {site} "
@@ -220,7 +220,7 @@ class Caltable:
                     gains = {}
                     dterms = dict(state.get('dterms', {}))
                     for site, d in legacy.items():
-                        g, dt = ehc.split_dtcal(d)
+                        g, dt = ehc.upgrade_caltable(d)
                         gains[site] = g
                         if dt is not None:
                             dterms[site] = dt
--- a/ehtim/caltable.py
+++ b/ehtim/caltable.py
@@ -90,20 +90,33 @@ class Caltable:
         self.tarr = ehc.upgrade_tarr(tarr)
         self.tkey = {self.tarr[i]['site']: i for i in range(len(self.tarr))}
 
-        # Save the data (upgrade per-site DTCAL recarrays as needed)
+        # Save the data, splitting out D-terms from any older welded tables
+        self.dterms = {}
         if isinstance(datadict, dict):
-            self.data = {site: ehc.upgrade_dtcal_circ(d)
-                         for site, d in datadict.items()}
+            self.data = {}
+            for site, d in datadict.items():
+                gains, dterms = ehc.split_dtcal(d)
+                self.data[site] = gains
+                if dterms is not None:
+                    self.dterms[site] = dterms
         else:
             self.data = datadict
 
     def __setstate__(self, state):
-        # Silently upgrade legacy pickles to the current mixedpol schema.
+        # Silently upgrade legacy pickles to the current schema.
         if 'tarr' in state:
             state['tarr'] = ehc.upgrade_tarr(state['tarr'])
         if 'data' in state and isinstance(state['data'], dict):
-            state['data'] = {site: ehc.upgrade_dtcal_circ(d)
-                             for site, d in state['data'].items()}
+            gains = {}
+            dterms = dict(state.get('dterms', {}))
+            for site, d in state['data'].items():
+                g, dt = ehc.split_dtcal(d)
+                gains[site] = g
+                if dt is not None:
+                    dterms[site] = dt
+            state['data'] = gains
+            state['dterms'] = dterms
+        state.setdefault('dterms', {})
         self.__dict__.update(state)
 
     def copy(self):
@@ -410,9 +423,8 @@ class Caltable:
                     preL = 1.
                     postL = 1.
 
-                # TODO: time-dependent D-term field added but unpopulated
-                valspre = np.array([(timepre, preR, preL, 0j, 0j)], dtype=ehc.DTCAL)
-                valspost = np.array([(timepost, postR, postL, 0j, 0j)], dtype=ehc.DTCAL)
+                valspre = np.array([(timepre, preR, preL)], dtype=ehc.DTCAL)
+                valspost = np.array([(timepost, postR, postL)], dtype=ehc.DTCAL)
 
                 gg = np.insert(gg, 0, valspre)
                 gg = np.append(gg, valspost)
@@ -595,9 +607,8 @@ class Caltable:
                     # TODO can we do this faster?
                     datatable = []
                     for i in range(len(times_merge)):
-                        # TODO: time-dependent D-term field added but unpopulated
                         datatable.append(
-                            np.array((times_merge[i], rscale_merge[i], lscale_merge[i], 0j, 0j),
+                            np.array((times_merge[i], rscale_merge[i], lscale_merge[i]),
                                      dtype=ehc.DTCAL))
                     data1[site] = np.array(datatable)
 
@@ -672,8 +683,7 @@ class Caltable:
                 gains_r_avg = np.mean(gains_r[np.array(times_stable == scan[0])])
 
                 # add them to a new datatable
-                # TODO: time-dependent D-term field added but unpopulated
-                datatable.append(np.array((scan[0], gains_r_avg, gains_l_avg, 0j, 0j), dtype=ehc.DTCAL))
+                datatable.append(np.array((scan[0], gains_r_avg, gains_l_avg), dtype=ehc.DTCAL))
 
             datatables[site] = np.array(datatable)
 
@@ -745,8 +755,7 @@ def load_caltable(obs, datadir, sqrt_gains=False):
             if sqrt_gains:
                 rscale = rscale**.5
                 lscale = lscale**.5
-            # TODO: time-dependent D-term field added but unpopulated
-            datatable.append(np.array((time, rscale, lscale, 0j, 0j), dtype=ehc.DTCAL))
+            datatable.append(np.array((time, rscale, lscale), dtype=ehc.DTCAL))
 
         datatables[site] = np.array(datatable)
     if len(datatables) > 0:
@@ -826,8 +835,7 @@ def make_caltable(obs, gains, sites, times):
         datatable = []
         for t in range(0, ntimes):
             gain = gains[s * ntimes + t]
-            # TODO: time-dependent D-term field added but unpopulated
-            datatable.append(np.array((times[t], gain, gain, 0j, 0j), dtype=ehc.DTCAL))
+            datatable.append(np.array((times[t], gain, gain), dtype=ehc.DTCAL))
         datatables[sites[s]] = np.array(datatable)
     if len(datatables) > 0:
         caltable = Caltable(obs.ra, obs.dec, obs.rf,
--- a/ehtim/caltable.py
+++ b/ehtim/caltable.py
@@ -35,21 +35,27 @@ from ehtim.warnings import MixedPolConventionWarning
 # Caltable object
 ##################################################################################################
 
-# D-terms are saved beside the gain files, one per site. The gain files stay in
-# their long-standing headerless five-column format; these carry a version line
-# so the format can move later. np.loadtxt ignores '#' lines.
+# NOTE: caltables loaded from old data carry no D-term tables; any D-term
+# information stays in the tarr, assumed fixed over the observation.
+# TODO: when the new calibration module lands, decide what happens when both
+# the tarr and the caltable define D-terms (zero the tarr? warn and prefer
+# the table?).
+
+# Any optional time-dependent D-term tables are saved beside the gain files,
+# opened with a version line. Gain files keep their headerless 5-column format.
 DTERM_FILE_SUFFIX = '_dterms.txt'
 DTERM_FILE_HEADER = '# ehtim caltable dterms format v1: time_mjd d1re d1im d2re d2im'
 
 
 class Caltable:
-    """A calibration table holding per-station gains and leakage (D-terms).
+    """A calibration table of time-dependent station gains and leakage (D-terms).
 
-       Gains and D-terms are stored in two separate per-site tables, each with
-       its own time column, because the two quantities have different natural
-       cadences: gains vary with the atmosphere (solved per scan or finer) while
-       D-terms are instrumental and static or slowly varying. A single-row table
-       represents a time-constant quantity.
+       Gains and D-terms live in separate per-site tables, each with its own
+       time column; a single-row table represents a time-constant quantity.
+       A caltable with no D-term table falls back to the D-terms in the tarr,
+       assumed fixed over the observation.
+       TODO: flag or resolve cases where both the tarr and the caltable
+       define D-terms.
 
        Attributes:
            source (str): The source name
@@ -149,9 +155,8 @@ class Caltable:
 
     @data.setter
     def data(self, datadict):
-        # Same normalization the constructor does, so a solver's caldict works
-        # either way. Leakage is refused, not split: splitting would move the
-        # D-terms out of the caller's own dict.
+        # normalize like the constructor does, but refuse welded tables:
+        # splitting here would move D-terms out of the caller's own dict
         if isinstance(datadict, dict):
             normalized = {}
             reshaped = False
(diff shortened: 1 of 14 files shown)


rohandahale (approved):

rohandahale (dismissed):

ehtim/observing/obs_simulate.py
```diff
@@ -738,11 +738,13 @@ def make_jones(obs, opacitycal=True, ampcal=True, phasecal=True, dcal=True,
...
 
         if caltable_path:
+            # the D-terms live on the tarr, which is what gets saved alongside
+            # the gains; the cal table itself carries gains only
             obs_tmp.tarr[i]['dr'] = dR
             obs_tmp.tarr[i]['dl'] = dL
             datatable = []
             for j in range(len(times)):
-                datatable.append(np.array((times[j], gainR[j], gainL[j], dR, dL),
+                datatable.append(np.array((times[j], gainR[j], gainL[j]),
```
achael, line 750: should we put in a single time entry dterm table into the saved caltable, or should we keep the default caltable format as having an empty dterm table if the dterms are fixed (dterm information only in the tarr)? 

Either way, let's put a TODO at the top of obs_simulate to consider adding time-variable dterms at some point.

ehtim/const_def.py
```diff
@@ -173,15 +174,24 @@
...
 DTCAL = DTCAL_CIRC  # legacy alias
 
+# D-term (leakage) tables, kept separate from the gain tables (DTCAL_*) so gains
+# and D-terms can be sampled on independent time grids. The CIRC field specs
+# mirror the D-term columns of DTCAL_CIRC; DTDTERM_LIN mirrors DTCAL_LIN's
+# asymmetry (no established physical names for linear leakage).
+DTDTERM_CIRC = [('time', 'f8'),
+                (('d_p1', 'dr'), 'c16'), (('d_p2', 'dl'), 'c16')]
+
+DTDTERM_LIN = [('time', 'f8'), ('d_p1', 'c16'), ('d_p2', 'c16')]
```
achael: i don't understand the comment above about mirroring DTCAL_LIN's asymmetry -- DTCAL_LIN does have aliases 'xscale','yscale' for 'p1scale','p2scale'. Seems like we should here have 'dx','dy' aliases for 'd_p1','d_p2' as well? or am I missing something?

ehtim/const_def.py
```diff
@@ -243,23 +253,74 @@ def upgrade_dtpol_circ(data):
     return data
 
 
-def upgrade_dtcal_circ(data):
-    """Upgrade a legacy DTCAL recarray (time, rscale, lscale) to DTCAL_CIRC.
-
-    Legacy DTCAL has no D-term fields; the upgrade allocates a new recarray
-    and zero-fills dr/dl. Idempotent.
+def split_dtcal(data):
```
achael: we didn't have the dummy dterm columns in the dtcal table for that long, so i'd prefer making less of a big deal about it in this function & docstring. It's probably still worth keeping the catch for the exceptionally rare circumstance if someone has a time variable dterm table saved but i expect that will never actually happen. I'd prefer we call this function 'upgrade_caltable', and make the main docstring simply 'upgrade legacy caltable types to current mixed-pol complient format. Pass dterm tables separately if appended to caltable'

ehtim/const_def.py
```diff
@@ -243,23 +253,74 @@ def upgrade_dtpol_circ(data):
...
+    if data is None:
+        return data, None
+    if not isinstance(data, _np.ndarray):
+        data = _np.asarray(data)
+    if data.dtype.names is None:
+        return data, None
+    data = _np.atleast_1d(data)
+
+    fields = data.dtype.fields
+    lin = 'xscale' in fields
```
achael: is this a robust enough check for if the table is linear? i'm slightly worried about if a datatable with only generic p1scale, p2scale keywords is passed in. Can we check if either the pair 'xscale'/'yscale' or 'rscale'/'lscale' exist in 'fields'? If neither exist, default to circular but print a warning.

ehtim/const_def.py
```diff
@@ -243,23 +253,74 @@ def upgrade_dtpol_circ(data):
...
+        # Gains only. Copy per field rather than re-viewing the buffer: a view
+        # would hand the caller's own array back to Caltable, so invert_gains
+        # and enforce_positive would write through to it, and it would also
+        # reinterpret a byte-swapped (big-endian) table instead of converting.
+        if data.dtype.names == gain_t.names:
+            gains = _np.zeros(len(data), dtype=gain_t)
+            for name in gain_t.names:
+                gains[name] = data[name]
+            return gains, None
+        return data, None
```
achael: i'm confused on this return here -- if the datatable.dtype.names != gain_t.names, why do we pass the raw data back as the gain table? should this not be an exception if we cannot interpret the gain fields?

ehtim/const_def.py
```diff
@@ -243,23 +253,74 @@ def upgrade_dtpol_circ(data):
...
+
+    if 'p1scale' not in fields:
```
achael: also confused by this if statement. if 'p1scale' is not in fields, that means that the datatable is probably in the old format, (just 'rscale','lscale'), right? don't we want to upgrade it to the new format then, not just pass through?

ehtim/const_def.py
```diff
@@ -243,23 +253,74 @@ def upgrade_dtpol_circ(data):
...
+        # would hand the caller's own array back to Caltable, so invert_gains
+        # and enforce_positive would write through to it, and it would also
+        # reinterpret a byte-swapped (big-endian) table instead of converting.
+        if data.dtype.names == gain_t.names:
+            gains = _np.zeros(len(data), dtype=gain_t)
+            for name in gain_t.names:
+                gains[name] = data[name]
+            return gains, None
+        return data, None
+
```
achael: add comment here that everything below is for rare case where dterms are appended to gain table

ehtim/caltable.py
```diff
@@ -28,14 +29,56 @@
 import ehtim.io.load
 import ehtim.io.save
 import ehtim.observing.obs_helpers as obsh
+from ehtim.warnings import MixedPolConventionWarning
 
 ##################################################################################################
 # Caltable object
 ##################################################################################################
 
+# D-terms are saved beside the gain files, one per site. The gain files stay in
```
achael: say "Any optional time-dependent d-term tables are saved beside the gain files'

ehtim/caltable.py
```diff
@@ -28,14 +29,56 @@
...
+# their long-standing headerless five-column format; these carry a version line
+# so the format can move later. np.loadtxt ignores '#' lines.
+DTERM_FILE_SUFFIX = '_dterms.txt'
+DTERM_FILE_HEADER = '# ehtim caltable dterms format v1: time_mjd d1re d1im d2re d2im'
 
-class Caltable:
+
+def _as_dterm_table(table):
```
achael: the content of this function is generic and not specific to dterm tables -- it fixes the one-row edge case for recarrays and returns none if  there is truly no data in the array or it has no field names. Let's call this something else like '_normalize_recarray'. 

Since this is likely to be a useful helper function in many contexts I would move this to const_def.py

ehtim/caltable.py
```diff
@@ -28,14 +29,56 @@
...
+    if not isinstance(table, np.ndarray):
+        table = np.asarray(table)
+    if table.dtype.names is None:
+        return None
+    table = np.atleast_1d(table)
+    return table if len(table) else None
+
+
+class Caltable:
+    """A calibration table holding per-station gains and leakage (D-terms).
```
achael: say 'time-dependent station gains and leakage'

also note that a caltable with no dterm table will use the dterms in the tarr, assumed to be fixed over an entire observation. Add a TODO about future work to avoid/flag cases where both might be defined

ehtim/caltable.py
```diff
@@ -48,12 +91,17 @@ class Caltable:
...
+                          DTDTERM. Empty when the table carries no leakage solution.
+           data (dict): backwards-compatible alias for ``gains``; it is the live
+                        dict, so in-place mutation through it is visible on the table
 
     """
 
     def __init__(self, ra, dec, rf, bw, datadict, tarr,
-                 source=ehc.SOURCE_DEFAULT, mjd=ehc.MJD_DEFAULT, timetype='UTC'):
+                 source=ehc.SOURCE_DEFAULT, mjd=ehc.MJD_DEFAULT, timetype='UTC',
+                 *, dterms=None):
```
achael: can we renamethe 'dterms' kwarg to 'dtermdict' and put it right after tarr in the list?

ehtim/caltable.py
```diff
@@ -90,20 +144,92 @@ def __init__(self, ra, dec, rf, bw, datadict, tarr,
...
-                         for site, d in datadict.items()}
+            self.gains = {}
+            for site, d in datadict.items():
+                site_gains, site_dterms = ehc.split_dtcal(d)
+                self.gains[site] = site_gains
+                if site_dterms is not None:
+                    self.dterms[site] = site_dterms
         else:
-            self.data = datadict
+            self.gains = datadict
```
achael: i see this was in the old code, but shouldn't there be an exception if isinstance(datadict,dict) is False? when does this else statement ever get used?

ehtim/caltable.py
```diff
@@ -90,20 +144,92 @@ def __init__(self, ra, dec, rf, bw, datadict, tarr,
...
+                reshaped |= site_gains is not table
+                normalized[site] = site_gains
+            # keep the caller's own dict when nothing needed reshaping
+            if reshaped:
+                datadict = normalized
+        self.gains = datadict
 
     def __setstate__(self, state):
-        # Silently upgrade legacy pickles to the current mixedpol schema.
+        # Rebind the migration onto a copy. Unpickling hands over a throwaway
```
achael: rewrite this comment to be more human readable. I think something just like "upgrade old version pickled caltables to the current schema" still captures the essence here

ehtim/caltable.py
```diff
@@ -90,20 +144,92 @@ def __init__(self, ra, dec, rf, bw, datadict, tarr,
...
+        state = dict(state)
+
+        # Silently upgrade legacy pickles to the current schema.
         if 'tarr' in state:
             state['tarr'] = ehc.upgrade_tarr(state['tarr'])
-        if 'data' in state and isinstance(state['data'], dict):
-            state['data'] = {site: ehc.upgrade_dtcal_circ(d)
-                             for site, d in state['data'].items()}
+
+        # Pre-split pickles carry 'data'; pop it, since the property shadows any
```
achael: this will not be understandable once we forget what 'pre-split' means. just say "deal with pickles before old 'data' attribute was renamed to 'gains'"

ehtim/caltable.py
```diff
@@ -121,6 +247,11 @@ def plot_dterms(self, sites='all', label=None, legend=True, clist=ehc.SCOLORS,
                     show=True, grid=True, export_pdf=""):
         """Make a plot of the D-terms.
 
+           Plots the leakage recorded in the array table (tarr), one point per
+           site. The cal table's own D-term table (self.dterms) is not read
```
achael: say "The caltable's own time-dependent D-term table"

ehtim/caltable.py
```diff
@@ -450,6 +585,13 @@ def applycal(self, obs, interp='linear', extrapolate=None,
         if not (self.tarr == obs.tarr).all():
             raise Exception("The telescope array in the Caltable is not the same as in the Obsdata")
 
+        if self.dterms:
+            warnings.warn(
+                "This cal table carries D-terms for "
```
achael: "cal table carries time-dependent D-terms"

ehtim/caltable.py
```diff
@@ -555,15 +697,43 @@ def merge(self, caltablelist, interp='linear', extrapolate=1):
...
         for caltable in caltablelist:
 
             # TODO check metadata!
 
             # TODO CHECK ARE THEY ALL REFERENCED TO SAME MJD???
             tarr2 = caltable.tarr.copy()
             tkey2 = caltable.tkey.copy()
-            data2 = caltable.data.copy()
+
+            # A site solved on one side only comes through as it is. Two
```
achael: sorry i really don't like the style of a lot of the comment from this version of claude.... way too long and technical 

let's just say 
"# merging time-dependent D-term tables."
"# If both caltables carry time-dependent D-terms for a site, raise for now. Composing D-terms requires full Jones treatment."

ehtim/caltable.py
```diff
@@ -555,15 +697,43 @@ def merge(self, caltablelist, interp='linear', extrapolate=1):
...
+            # resampled onto the D-term grid. That waits for the solvers that
+            # write leakage into a cal table. Two tables loaded from disk can
+            # reach this.
+            for site, dterm_table in caltable.dterms.items():
+                if site in dterms1:
+                    raise NotImplementedError(
+                        f"merge: both caltables carry D-terms for site {site}. "
+                        "Composing two leakage solutions is deferred; it is a "
+                        "Jones product, not a gain-style multiply.")
+                # A site can carry leakage with no gain rows, so it would never
```
achael: reduce comments to 

"# if D-terms are only in the merged-in caltable, we can add them"
"# be sure to add the tarr row; it may be missed in the gain merge below"

ehtim/caltable.py
```diff
@@ -672,33 +849,46 @@ def scan_avg(self, obs, incoherent=True):
...
 
                 # add them to a new datatable
-                # TODO: time-dependent D-term field added but unpopulated
-                datatable.append(np.array((scan[0], gains_r_avg, gains_l_avg, 0j, 0j), dtype=ehc.DTCAL))
+                datatable.append(np.array((scan[0], gains_r_avg, gains_l_avg), dtype=ehc.DTCAL))
 
             datatables[site] = np.array(datatable)
 
         if len(datatables) > 0:
+            # averaging is a gain-table operation; the leakage solution, which
```
achael: put TODO: we may want to eventually have a function to average D-terms.

ehtim/caltable.py
```diff
@@ -760,6 +979,15 @@ def load_caltable(obs, datadir, sqrt_gains=False):
 
 def save_caltable(caltable, obs, datadir='.', sqrt_gains=False):
     """Saves a Caltable object to text files in the given directory
+
+       Three kinds of file: array.txt for the telescope array, one
+       '<source>_<site>.txt' per site of gains (headerless, five columns:
+       time_mjd rre rim lre lim), and one '<source>_<site>_dterms.txt' per site
+       that carries leakage. The gain format is unchanged and stays headerless;
```
achael: "time-dependent leakage"

ehtim/caltable.py
```diff
@@ -804,6 +1034,31 @@ def save_caltable(caltable, obs, datadir='.', sqrt_gains=False):
...
+    # without gains. sqrt_gains is a gain convention and does not touch these.
+    for site_info in caltable.tarr:
+        site = site_info['site']
+        filename = datadir + '/' + src + '_' + site + DTERM_FILE_SUFFIX
+
+        if len(caltable.dterms.get(site, [])) == 0:
+            # Left over from an earlier save into this directory; it would be
+            # read back as this table's leakage.
+            if os.path.exists(filename):
+                os.remove(filename)
```
achael, line 1007: we don't autoremove gain tables if they already exist in the directory, so this is asymmetric. We should treat these similarly, and add a a catch for both. 

Also, add an initial loop to the top of the save function to see if any table text files would be overwritten or deleted.  Include an overwrite=False keyword argument; raise an exception if any files would be overwritten or deleted unless the user sets this to True.

achael (changes requested):
Looks good overall, and very sorry for the delays.

My comments are mostly stylistic, except I am confused by the logic of the split_dtcal function. Can you look at my confused comments and go over this again, making the logic more transparent or correcting issues if they do exist? 

We should be very clear to note at the top of the caltable.py file that the default for now is for the caltable to carry no dterms tables when loading old data, but there will still be dterm data in the tarr. When we build the new calibration module we will need to explicitly address any ambiguities between having dterm data in both places -- we might want to set tarr dterms to zero if the dterm field in the caltable is nonzero, or we should keep the tarr dterms as is and just warn that they will be ignored in favor of any existing dterm data in the tables. Let's not decide now--just put in a TODO. But we should talk about this and make this behavior a quick follow up with its own small PR. 

the comments in this version of Claude (Opus 5?) are really way too long and confusing for my taste -- they tend to over-explain every design decision and fixed bug in a way that will not be useful for future readers (at least future human readers, but i want the code to stay human-readable). Maybe add something to CLAUDE.md or the project plan to favor short, 1-2 line comments?

ehtim/const_def.py
rohandahale: You are right, I had this backwards. DTDTERM_LIN now has `dx`/`dy` aliases for `d_p1`/`d_p2`, matching DTCAL_LIN. The comment about asymmetry was wrong and is gone.

ehtim/const_def.py
rohandahale: Renamed to `upgrade_caltable` with your docstring, and  the appended dterm case is dropped.

ehtim/const_def.py
rohandahale: Now checks for `xscale`/`yscale` first, then `rscale`/`lscale`, and only falls back to circular when neither is present, with a warning in exactly the generic names case you described.

ehtim/observing/obs_simulate.py
rohandahale: Left the simulated caltable with no dterm table for now, since the dterms are fixed per site and already round trip through the tarr, which keeps the default consistent with what you describe for loading old data. Added the TODO at the top of the file about time-variable dterms.

ehtim/const_def.py
rohandahale: Both were wrong. The old code passed unrecognised tables through as gains, which is how a d-term table handed to `datadict` ended up stored as gains and dropped on save. It now raises on any field set it cannot interpret. `p1scale` absent means the old format, which is upgraded, not passed through.

ehtim/const_def.py
rohandahale: Both were wrong. The old code passed unrecognised tables through as gains, which is how a d-term table handed to `datadict` ended up stored as gains and dropped on save. It now raises on any field set it cannot interpret. `p1scale` absent means the old format, which is upgraded, not passed through.

ehtim/const_def.py
rohandahale: The is moot now. That is deleted

ehtim/caltable.py
rohandahale: Reworded

ehtim/caltable.py
rohandahale: Agreed. Renamed `_normalize_recarray` and moved to const_def.py.

ehtim/caltable.py
rohandahale: Reworded to "time-dependent station gains and leakage". Added the tarr fallback note and the TODO.

ehtim/caltable.py
rohandahale: Renamed `dtermdict` and moved to right after `tarr`. Note this makes it positional, so I fixed the one in-repo caller that passed `source`/`mjd` positionally past `tarr` (scripts/imaging.py).

ehtim/caltable.py
rohandahale: Checked it and it is from the original code. It now raises `TypeError`.

ehtim/caltable.py
rohandahale: Reworded

ehtim/caltable.py
rohandahale: Reworded

ehtim/caltable.py
rohandahale: reworded to "time-dependent"

ehtim/caltable.py
rohandahale: reworded to "time-dependent"

ehtim/caltable.py
rohandahale: reworded to "time-dependent"

ehtim/caltable.py
rohandahale: Reworded

ehtim/caltable.py
rohandahale: Added the TODO about averaging D-terms.

ehtim/caltable.py
rohandahale: Reworded

ehtim/caltable.py
rohandahale: Good catch! `save_caltable` and `save_txt` now take `overwrite=False`. Stale gain files are now removed the same way stale dterm files were.

rohandahale:
@achael Thanks! I have addressed all comments. 

1) Added a NOTE + TODO at the top of caltable.py and in the class docstring: a caltable loaded from old data carries no D-term table, and the tarr D-terms stay put, assumed fixed over the observation. Agreed we should not decide the tarr vs table now. 

2) You are right about over explaining comments. I went through each one and cut them back. 

3) split_dtcal. The splitting job is gone entirely. What remains is a plain `upgrade_caltable`, similar to `upgrade_tarr`: it normalizes the shapes the solvers actually produce, upgrades legacy gain tables to the current dtype and raises on anything it cannot interpret, including the combined format.

ehtim/observing/obs_simulate.py
achael: looks good

ehtim/const_def.py
```diff
@@ -243,23 +253,70 @@ def upgrade_dtpol_circ(data):
...
+_CAL_GAIN_FIELDS = frozenset({'rscale', 'lscale', 'xscale', 'yscale', 'p1scale', 'p2scale'})
+
+
+def upgrade_caltable(data):
+    """Upgrade legacy per-site caltable gains to the current mixed-pol format.
+
+    Idempotent; raises on field names it cannot interpret.
+    """
+    import numpy as _np
+    if data is None:
```
achael, line 283: isn't this the same logic as normalize_recarray? couldn't we just call that?

ehtim/caltable.py
achael: looks good

achael (dismissed):
looks good now, just one super small point where i think you could use the normalize_recarr helper

ehtim/const_def.py
rohandahale: You're right, it was the same thing duplicated. `upgrade_caltable` now calls `_normalize_recarray`.

rohandahale:
@achael Ready to merge!

ehtim/const_def.py
achael: looks good!

achael (approved):
looks good!

